PR #7, opened 2026-01-23
aws onboarding page improvements

beng360:
added check for aws keys before configuring aws, and added docs

### DIFF
--- a/client/src/app/aws/onboarding/page.tsx
+++ b/client/src/app/aws/onboarding/page.tsx
@@ -449,6 +449,66 @@ make dev`}</pre>
     );
   }
 
+  // Show error page if credentials are configured but account ID cannot be retrieved (invalid credentials)
+  if (onboardingData && !onboardingData.auroraAccountId && !isConfigured) {
+    return (
+      <div className="min-h-screen bg-black flex items-center justify-center p-4 sm:p-6">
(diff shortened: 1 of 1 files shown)


server/routes/aws/onboarding.py
```diff
@@ -25,6 +25,51 @@ def get_authenticated_user_id():
     return get_user_id_from_request()
 
 
+@onboarding_bp.route('/aws/credentials/check', methods=['GET', 'OPTIONS'])
+def check_aws_credentials():
+    """
```
OlivierTrudeau, line 30: I find the name of this misleading. Maybe something more related to env, setup, verification rather than credentials

client/src/app/aws/onboarding/page.tsx
OlivierTrudeau, line 1: Possible to make it not all on one line?
Screenshot 2026-01-23 at 10 17 11 AM

OlivierTrudeau: Same file but Idk where the code is but the Credentials thing I find is confusing. Should be more setup/env oriented
image

server/routes/aws/onboarding.py
beng360: fixed

client/src/app/aws/onboarding/page.tsx
beng360: fixed as well

PR #10, opened 2026-01-23
github fixes for setup guide

beng360:
github check for env variable

### DIFF
--- a/client/src/components/github-provider-integration.tsx
+++ b/client/src/components/github-provider-integration.tsx
@@ -112,14 +112,15 @@ export class GitHubIntegrationService {
       }
 
       return data.oauth_url;
-    } catch (error: any) {
-      // Re-throw with handled flag to prevent React from treating it as uncaught
-      if (error.isHandled) {
+    } catch (error) {
+      // Re-throw handled errors
+      if ((error as any)?.isHandled) {
         throw error;
       }
-      // For any unexpected errors, mark as handled
-      (error as any).isHandled = true;
-      throw error;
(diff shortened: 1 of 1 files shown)


client/src/components/github-provider-integration.tsx
```diff
@@ -72,23 +73,54 @@ export class GitHubIntegrationService {
...
+      return data.oauth_url;
+    } catch (error: any) {
+      // Re-throw with handled flag to prevent React from treating it as uncaught
+      if (error.isHandled) {
+        throw error;
+      }
+      // For any unexpected errors, mark as handled
+      (error as any).isHandled = true;
+      throw error;
+    }
```
damianloch, line 124: This just catches al unexpected errrors and marks as handled when they should surface naturally

client/src/components/github-provider-integration.tsx
```diff
@@ -503,13 +535,50 @@ export default function GitHubProviderIntegration() {
...
+      // Prevent error from being logged as uncaught
+      if (error.isHandled) {
+        // Error is already marked as handled, just log for debugging
+        console.log("OAuth error (handled):", error.message);
+      } else {
+        console.error("OAuth error:", error);
+      }
+      
+      // Check if this is a configuration error
+      if (error.errorCode === 'GITHUB_NOT_CONFIGURED' || error.message?.includes('not configured')) {
```
damianloch: error.message?.includes('not configured') seems fragile, trusting the first half of this line should be enough

PR #12, opened 2026-01-24
Fix/ovh connector status

isiddharthsingh:
## Summary                                                                                                                                                                        
                                                                                                                                                                                    
  Fix OVH connector to show correct connection status on connectors page and add project management dialog.                                                                         
                                                                                                                                                                                    
  ## Changes                                                                                                                                                                        
                                                                                                                                                                                    
  - Add `useCustomConnection: true` to OVH config so it checks API for connection status                                                                                            
  - Add OVH provider integration component with project toggle, root project selection, and disconnect                                                                              
  - Add OVH dialog to connector dialogs (consistent with GCP)                                                                                                                       
  - Fix OAuth2 redirect_uri to include `/ovh_api` prefix                                                                                                                            
                                                                                                                                                                                    
  ## Testing                                                                                                                                                                        
                                                                                                                                                                                    
  - Connect OVH account via OAuth2                                                                                                                                                  
  - Verify OVH shows as "Connected" on connectors page                                                                                                                              
  - Click "Manage" to view/toggle projects                                                                                                                                          
  - Set root project and verify badge appears                                                                                                                                       
  - Disconnect and verify status updates

### DIFF
--- a/client/src/components/ovh-provider-integration.tsx
+++ b/client/src/components/ovh-provider-integration.tsx
@@ -17,6 +17,7 @@ export default function OvhProviderIntegration({ onDisconnect }: OvhProviderInte
   const [userId, setUserId] = useState<string | null>(null);
   const [projects, setProjects] = useState<Project[]>([]);
   const [isLoading, setIsLoading] = useState(true);
+  const [isSaving, setIsSaving] = useState(false);
   const [isDisconnecting, setIsDisconnecting] = useState(false);
   const [togglingProjectId, setTogglingProjectId] = useState<string | null>(null);
   const { toast } = useToast();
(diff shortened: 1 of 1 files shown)


client/src/components/ovh-provider-integration.tsx
```diff
@@ -0,0 +1,238 @@
...
+        title: "Error",
+        description: error.message || "Failed to update project",
+        variant: "destructive",
+      });
+    } finally {
+      setTogglingProjectId(null);
+    }
+  }
+
+  async function handleSetAsRoot(_providerId: string, projectId: string): Promise<void> {
```
damianloch: _providerId parameter is never used

client/src/components/ovh-provider-integration.tsx
damianloch, line 1: there is no loading state for setting a root project unlike how GCP has...

client/src/components/ovh-provider-integration.tsx
isiddharthsingh: removed the underscore from _providerId to match how gcp does it (the param is still needed for the callback signature but yeah it wasn't being used)

client/src/components/ovh-provider-integration.tsx
isiddharthsingh: added the isSaving state for the set root project action, same pattern as gcp now

PR #13, opened 2026-01-24
Fix/scaleway connector status

isiddharthsingh:
## Summary                                                                                                                                                                        
                                                                                                                                                                                    
  Add project management dialog for Scaleway connector on connectors page.                                                                                                          
                                                                                                                                                                                    
  ## Changes                                                                                                                                                                        
                                                                                                                                                                                    
  - Add `useCustomConnection: true` to Scaleway config for API-based status check                                                                                                   
  - Add Scaleway provider integration component with project toggle, root project selection, and disconnect                                                                         
  - Add Scaleway dialog to connector dialogs (consistent with GCP/OVH)                                                                                                              
                                                                                                                                                                                    
  ## Testing                                                                                                                                                                        
                                                                                                                                                                                    
  - Connect Scaleway account                                                                                                                                                        
  - Verify Scaleway shows as "Connected" on connectors page                                                                                                                         
  - Click "Manage" to view/toggle projects                                                                                                                                          
  - Set root project and verify badge appears                                                                                                                                       
  - Disconnect and verify status updates                                                                                                                                            
                                                                                                                                                                                    
  ## Note                                                                                                                                                                           
                                                                                                                                                                                    
  This branch is based on `fix/ovh-connector-status`. After OVH PR is merged, rebase on main and force push.

### DIFF
--- a/client/src/components/scaleway-provider-integration.tsx
+++ b/client/src/components/scaleway-provider-integration.tsx
@@ -1,6 +1,6 @@
 'use client';
 
-import { useState, useEffect } from 'react';
+import { useState, useEffect, useCallback } from 'react';
 import { Button } from "@/components/ui/button";
 import { useToast } from '@/hooks/use-toast';
 import { Loader2, RefreshCw, LogOut } from 'lucide-react';
@@ -36,16 +36,10 @@ export default function ScalewayProviderIntegration({ onDisconnect }: ScalewayPr
     fetchUserId();
   }, []);
 
-  useEffect(() => {
-    if (userId) {
-      loadProjects();
-    }
-  }, [userId]);
-
-  async function loadProjects(forceRefresh = false): Promise<void> {
+  const loadProjects = useCallback(async (forceRefresh = false): Promise<void> => {
     setIsLoading(true);
     try {
-      const fetchedProjects = await fetchProjects('scaleway', forceRefresh, projects);
+      const fetchedProjects = await fetchProjects('scaleway', forceRefresh);
       setProjects(fetchedProjects);
     } catch (error: any) {
       console.error('Error loading Scaleway projects:', error);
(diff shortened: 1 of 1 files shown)


client/src/components/scaleway-provider-integration.tsx
```diff
@@ -0,0 +1,238 @@
...
+        title: "Error",
+        description: error.message || "Failed to update project",
+        variant: "destructive",
+      });
+    } finally {
+      setTogglingProjectId(null);
+    }
+  }
+
+  async function handleSetAsRoot(_providerId: string, projectId: string): Promise<void> {
```
damianloch: _providerId unused

client/src/components/scaleway-provider-integration.tsx
```diff
@@ -0,0 +1,238 @@
...
+      await saveProjects('scaleway', updatedProjects);
+
+      const toggledProject = updatedProjects.find(p => p.projectId === projectId);
+      toast({
+        title: "Success",
+        description: `Project ${toggledProject?.enabled ? 'enabled' : 'disabled'}`,
+      });
+    } catch (error: any) {
+      console.error('Error toggling project:', error);
+      loadProjects(true);
```
damianloch: When toggle fails, it calls loadProjects(true) to refresh, but the optimistic update on line 65-67 has already modified the UI

client/src/components/scaleway-provider-integration.tsx
```diff
@@ -0,0 +1,238 @@
...
+      }
+    }
+    fetchUserId();
+  }, []);
+
+  useEffect(() => {
+    if (userId) {
+      loadProjects();
+    }
+  }, [userId]);
```
damianloch: loadProjects is called inside useEffect but not declared as a dependency. This violates the exhaustive-deps rule and can cause stale closures.

client/src/components/scaleway-provider-integration.tsx
isiddharthsingh: the providerId param is needed because of the callback signature from ProjectListItem - it passes both providerId and projectId when you click "set as root". we don't use it since we already know we're in scaleway, but we need to accept it to match the type. removed the underscore to match how gcp does it.

client/src/components/scaleway-provider-integration.tsx
isiddharthsingh: fixed. now it saves the original projects before the optimistic update and restores directly from that on error, instead of making another api call. cleaner and faster.

client/src/components/scaleway-provider-integration.tsx
isiddharthsingh: fixed - wrapped loadProjects in useCallback and added it to the dependency array so it properly follows exhaustive-deps.

PR #21, opened 2026-01-26
Optimize LLM costs for testing

damianloch:
To optimize costs, changed the RCA cost optimization model to haiku 3, acccesible via openrouter or anthropic. Replaced all other references to gemini models so that the code can run with only 1 API key (openrouter or anthropic). Create a unified model config to centralize model management

### DIFF
--- a/server/chat/backend/agent/llm.py
+++ b/server/chat/backend/agent/llm.py
@@ -24,15 +24,8 @@ class ModelConfig:
     MAIN_MODEL = "anthropic/claude-sonnet-4.5"
     VISION_MODEL = "anthropic/claude-sonnet-4.5"
     
-    # Background RCA model - selected based on RCA_OPTIMIZE_COSTS env var
-    @staticmethod
-    def get_rca_model() -> str:
-        """Get RCA model based on RCA_OPTIMIZE_COSTS env var (defaults to cost-optimized)."""
-        optimize_costs = os.getenv("RCA_OPTIMIZE_COSTS", "true").lower() == "true"
-        if optimize_costs:
-            return "anthropic/claude-3-haiku"  # Cost-optimized
-        else:
-            return "anthropic/claude-opus-4.5"  # High-quality
+    # Background RCA model - selected based on RCA_OPTIMIZE_COSTS env var (defaults to cost-optimized)
+    RCA_MODEL = "anthropic/claude-3-haiku" if os.getenv("RCA_OPTIMIZE_COSTS", "true").lower() == "true" else "anthropic/claude-opus-4.5"
     
     # Summarization models
     INCIDENT_REPORT_SUMMARIZATION_MODEL = "anthropic/claude-sonnet-4.5"  # For incident reports and chat context
--- a/server/chat/backend/agent/utils/chat_context_manager.py
+++ b/server/chat/backend/agent/utils/chat_context_manager.py
@@ -74,9 +74,9 @@ class ChatContextManager:
 
     @classmethod
     def create_conversation_summary(
-        cls, messages: List[Any], model_name: str = None
+        cls, messages: List[Any]
     ) -> str:
-        """Create a summary of the conversation history."""
+        """Create a summary of the conversation history using centralized model config."""
         try:
             # Convert messages to a readable format for summarization
             conversation_text = cls._format_messages_for_summary(messages)
@@ -106,7 +106,7 @@ Provide a detailed summary that preserves essential context:"""
             from ..llm import ModelConfig
             llm_manager = LLMManager()
             summary = llm_manager.summarize(
-                conversation_text, model=model_name or ModelConfig.INCIDENT_REPORT_SUMMARIZATION_MODEL
+                conversation_text, model=ModelConfig.INCIDENT_REPORT_SUMMARIZATION_MODEL
             )
 
             logger.info(f"Generated conversation summary ({len(summary)} chars)")
@@ -248,7 +248,7 @@ Provide a detailed summary that preserves essential context:"""
             logger.info(
                 f"Summarizing entire conversation with {len(messages)} messages"
             )
-            summary_text = cls.create_conversation_summary(messages, model_name)
+            summary_text = cls.create_conversation_summary(messages)
 
             # Create a system message with the summary that replaces everything
             summary_message = SystemMessage(
--- a/server/chat/background/task.py
+++ b/server/chat/background/task.py
@@ -414,7 +414,7 @@ async def _execute_background_chat(
             selected_project_id=None,
             messages=[human_message],
             question=initial_message,
-            model=ModelConfig.get_rca_model(),
+            model=ModelConfig.RCA_MODEL,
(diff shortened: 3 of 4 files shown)


server/chat/backend/agent/utils/chat_context_manager.py
```diff
@@ -105,9 +106,10 @@ def create_conversation_summary(
 Provide a detailed summary that preserves essential context:"""
 
             # Create isolated LLM instance for summarization
+            from ..llm import ModelConfig
             llm_manager = LLMManager()
             summary = llm_manager.summarize(
-                conversation_text, model=summarization_model
+                conversation_text, model=summarization_model or ModelConfig.SUMMARIZATION_MODEL
             )
```
OlivierTrudeau, line 110: good idea to centralize but maybe make it cleaner. Because now you cold centralize but it would be overwritten here. Also, line 82 of this file, it's model_name or gemini.... Remove that and just keep ModelConfig

server/chat/backend/agent/llm.py
```diff
@@ -13,46 +13,103 @@
...
+class ModelConfig:
+    """Centralized model configuration for all Aurora LLM usage.
+    
+    All model selections are defined here in one place for easy maintenance.
+    Change these values to switch providers across the entire application.
+    """
+    
+    # Primary models for chat and operations
+    MAIN_MODEL = "anthropic/claude-sonnet-4.5"
+    SQL_MODEL = "anthropic/claude-sonnet-4.5"
```
OlivierTrudeau: Remove the SQL model and sql logic in code since sql isn't used

server/chat/backend/agent/llm.py
```diff
@@ -13,46 +13,103 @@
...
+    VISION_MODEL = "anthropic/claude-sonnet-4.5"
+    
+    # Background RCA models - with cost optimization
+    # When RCA_OPTIMIZE_COSTS=true:
+    #   - Use Claude 3 Haiku (reliable, fast, and cheap)
+    #   - Works in both direct and OpenRouter modes
+    RCA_MODEL_COST_OPTIMIZED_DIRECT = "anthropic/claude-3-haiku"
+    RCA_MODEL_COST_OPTIMIZED_OPENROUTER = "anthropic/claude-3-haiku"  # Proven reliable
```
OlivierTrudeau: Make this one variable. No use in differentiating between openrouter and direct especially considering you can only have one or the other

server/chat/backend/agent/llm.py
```diff
@@ -13,46 +13,103 @@
...
+    RCA_MODEL_HIGH_QUALITY = "anthropic/claude-opus-4.5"  # High-quality RCA (when RCA_OPTIMIZE_COSTS=false)
+    
+    # Summarization and utility models
+    SUMMARIZATION_MODEL = "anthropic/claude-sonnet-4.5"  # For general summarization
+    FAST_SUMMARIZATION_MODEL = "anthropic/claude-sonnet-4.5"  # For quick summaries (tool context)
```
OlivierTrudeau: Maybe change the name but should be able to infer the difference between these models here

server/chat/backend/agent/llm.py
```diff
@@ -13,46 +13,103 @@
...
+    SQL_MODEL = "anthropic/claude-sonnet-4.5"
+    VISION_MODEL = "anthropic/claude-sonnet-4.5"
+    
+    # Background RCA models - with cost optimization
+    # When RCA_OPTIMIZE_COSTS=true:
+    #   - Use Claude 3 Haiku (reliable, fast, and cheap)
+    #   - Works in both direct and OpenRouter modes
+    RCA_MODEL_COST_OPTIMIZED_DIRECT = "anthropic/claude-3-haiku"
+    RCA_MODEL_COST_OPTIMIZED_OPENROUTER = "anthropic/claude-3-haiku"  # Proven reliable
+    RCA_MODEL_HIGH_QUALITY = "anthropic/claude-opus-4.5"  # High-quality RCA (when RCA_OPTIMIZE_COSTS=false)
```
OlivierTrudeau: Actually, make all these 3 variables one and make an if else blcok to only have the optimized cost or high cost depending on the env var. So then in code it's always just using RCA_MODEL and no logic in code

server/chat/backend/agent/llm.py
```diff
@@ -13,46 +13,103 @@
...
+    
+    # Email report generation
+    EMAIL_REPORT_MODEL = "anthropic/claude-sonnet-4.5"
+    
+    @classmethod
+    def get_rca_model(cls, optimize_costs: bool = True, provider_mode: Optional[str] = None) -> str:
+        """Get the appropriate RCA model based on cost optimization setting and provider mode.
+        
+        Args:
+            optimize_costs: If True, use cost-optimized models
```
OlivierTrudeau: once you fix the logic for the RCA_MODEL variable you won't need this method anymore

server/chat/backend/agent/llm.py
```diff
@@ -13,46 +13,103 @@
...
         self.provider_mode = provider_mode or os.getenv("LLM_PROVIDER_MODE")
 
-        # Default models
-        self.default_main_model = main_model
-        self.default_sql_model = sql_model
-        self.default_vision_model = vision_model
+        # Use centralized model config if not explicitly overridden
+        self.default_main_model = main_model or ModelConfig.MAIN_MODEL
+        self.default_sql_model = sql_model or ModelConfig.SQL_MODEL
+        self.default_vision_model = vision_model or ModelConfig.VISION_MODEL
```
OlivierTrudeau: Just make it the model.config one no or

server/chat/background/summarization.py
```diff
@@ -572,10 +573,11 @@ def generate_incident_summary_from_chat(
             citations=all_citations if all_citations else None,
         )
 
-        # Use Flash to generate summary (fast, no tool calls needed)
-        # Uses provider-aware factory - will use direct Google API if configured, else OpenRouter
+        # Use centralized model config for email report generation
+        from chat.backend.agent.llm import ModelConfig
```
OlivierTrudeau: Two imports in the same file? Could you avoid the lazy import?

server/chat/background/task.py
```diff
@@ -401,8 +401,14 @@ async def _execute_background_chat(
...
         human_message = HumanMessage(content=initial_message)
 
+        # Import centralized model config
+        from chat.backend.agent.llm import ModelConfig
+
         # Create state with is_background=True and rca_context for system prompt
-        # Use Gemini 3 Pro for background chats by default, Claude Sonnet 4.5 for cost optimization
+        # Use centralized model configuration for RCA with provider mode awareness
+        optimize_costs = os.getenv("RCA_OPTIMIZE_COSTS", "").lower() == "true"
+        provider_mode = os.getenv("LLM_PROVIDER_MODE", "direct")
```
OlivierTrudeau: If you can also use the logic for the provider in the config instead of in code would be ideal

PR #24, opened 2026-01-26
Sms10221/dev 742 aurora learn

isiddharthsingh:
## Summary                                                                                                                                                                        
                                                                                                                                                                                    
  Implements **Aurora Learn** - a feature that learns from user feedback on completed RCAs to improve future incident investigations.                                               
                                                                                                                                                                                    
  ## What it does                                                                                                                                                                   
                                                                                                                                                                                    
  - Users can rate completed RCA analyses with thumbs up/down                                                                                                                       
  - Positively-rated RCAs are stored in Weaviate for vector similarity search                                                                                                       
  - When new incidents occur, Aurora searches for similar past incidents with positive feedback                                                                                     
  - If a match is found (similarity ≥ 70%), context from the helpful RCA is injected into the new investigation                                                                     
  - Users can toggle Aurora Learn on/off in Settings > Knowledge Base                                                                                                               
                                                                                                                                                                                    
  ## Changes                                                                                                                                                                        
                                                                                                                                                                                    
  ### Backend                                                                                                                                                                       
  - **Database**: Added `incident_feedback` table with RLS policies                                                                                                                 
  - **API**: New `/api/incidents/<id>/feedback` endpoints for submitting/retrieving feedback                                                                                        
  - **API**: New `/api/user/preferences/aurora-learn` endpoints for toggle setting                                                                                                  
  - **Weaviate**: New `IncidentKnowledge` collection for storing helpful RCAs                                                                                                       
  - **RCA Prompt Builder**: Added `inject_aurora_learn_context()` helper for context injection                                                                                      
  - **Connectors**: Integrated Aurora Learn into Grafana, Datadog, Netdata, and Splunk                                                                                              
                                                                                                                                                                                    
  ### Frontend                                                                                                                                                                      
  - **IncidentFeedback**: New component with thumbs up/down buttons and tooltips                                                                                                    
  - **IncidentCard**: Added feedback section for completed analyses                                                                                                                 
  - **KnowledgeBaseSettings**: Added Aurora Learn toggle with description                                                                                                           
  - **API Routes**: Next.js proxy routes for feedback and preferences                                                                                                               
                                                                                                                                                                                    
  ## How it works                                                                                                                                                                   
                                                                                                                                                                                    
  1. RCA completes → User sees 👍/👎 buttons                                                                                                                                        
  2. **Thumbs up** → Full RCA stored in Weaviate (summary, thoughts, citations)                                                                                                     
  3. **Thumbs down** → Stored in DB only (for audit), optional comment                                                                                                              
  4. New similar incident → Aurora searches Weaviate for matches                                                                                                                    
  5. Match found → Context injected into RCA prompt (invisible to user)                                                                                                             
                                                                                                                                                                                    
  ## Test plan                                                                                                                                                                      
                                                                                                                                                                                    
  - [x] Trigger alert, let RCA complete                                                                                                                                             
  - [x] Click thumbs up → verify stored in Weaviate                                                                                                                                 
  - [x] Click thumbs down with comment → verify stored in DB only                                                                                                                   
  - [x] Trigger similar alert → check logs for `[AURORA LEARN] Injecting context`                                                                                                   
  - [x] Toggle Aurora Learn off → verify feedback blocked                                                                                                                           
  - [x] Toggle Aurora Learn off → verify no context injection on new alerts

### DIFF
--- a/server/chat/background/rca_prompt_builder.py
+++ b/server/chat/background/rca_prompt_builder.py
@@ -141,6 +141,21 @@ def _get_similar_good_rcas_context(
                         context_parts.append(f"- {truncated}")
                     context_parts.append("")
 
+            # Add commands used during investigation (without outputs)
+            citations = match.get("citations", [])
+            if citations:
+                commands = [
+                    c.get("command", "")
+                    for c in citations
+                    if c.get("command")
+                ][:5]
+                if commands:
+                    context_parts.append("**Commands used in investigation:**")
+                    for cmd in commands:
+                        truncated = cmd[:150] + "..." if len(cmd) > 150 else cmd
+                        context_parts.append(f"- `{truncated}`")
+                    context_parts.append("")
+
         context_parts.extend([
             "---",
             "**Note**: Use the above context as guidance. The current incident may have different root causes.",


client/src/app/incidents/components/IncidentFeedback.tsx
```diff
@@ -0,0 +1,257 @@
...
+        title: 'Feedback submitted',
+        description: feedbackType === 'helpful'
+          ? 'Thanks! Aurora will learn from this analysis.'
+          : 'Thanks for your feedback.',
+      });
+    } catch (err) {
+      console.error('Error submitting feedback:', err);
+      const message = err instanceof Error ? err.message : 'Failed to submit feedback';
+
+      if (message.includes('Aurora Learn is disabled')) {
```
damianloch: We should use error codes or something more robust here

client/src/app/incidents/components/IncidentFeedback.tsx
```diff
@@ -0,0 +1,257 @@
...
+        skipComment ? undefined : comment
+      );
+
+      setFeedback({
+        id: '',
+        feedbackType,
+        comment: skipComment ? undefined : comment,
+        createdAt: new Date().toISOString(),
+      });
+      setState('submitted');
```
damianloch, line 136: is this sending fake data?

server/routes/incident_feedback/weaviate_client.py
```diff
@@ -0,0 +1,342 @@
...
+    global _client, _collection
+
+    if _client is not None:
+        try:
+            if _client.is_ready():
+                return _client, _collection
+            logger.warning("[AURORA LEARN] Connection not ready, reconnecting...")
+        except Exception:
+            logger.warning("[AURORA LEARN] Connection lost, reconnecting...")
+        _reset_client()
```
damianloch: I think we need thread locking here, seem like 2 simultaneous connections may overwrite eachother

server/routes/incident_feedback/weaviate_client.py
```diff
@@ -0,0 +1,342 @@
...
+
+
+def _reset_client() -> None:
+    """Reset the client state, attempting to close any existing connection."""
+    global _client, _collection
+    if _client is not None:
+        try:
+            _client.close()
+        except Exception:
+            pass
```
damianloch: we should log the exception

server/routes/incident_feedback/routes.py
damianloch, line 1: Context manager with manual commit/rollback is error-prone. Early returns (lines 97, 113, 128) 
skip the commit, and the exception handler doesn't explicitly rollback. Use explicit 
try/commit/except/rollback/finally pattern instead.

server/routes/incident_feedback/routes.py
```diff
@@ -0,0 +1,312 @@
...
+        UUID(value)
+        return True
+    except (ValueError, TypeError):
+        return False
+
+
+def _is_aurora_learn_enabled(user_id: str) -> bool:
+    """Check if Aurora Learn is enabled for a user. Defaults to True."""
+    setting = get_user_preference(user_id, AURORA_LEARN_PREFERENCE_KEY, default=True)
+    return setting is True or setting == "true"
```
damianloch: which is it? Delete the or which it won't ever be

server/routes/incident_feedback/routes.py
```diff
@@ -0,0 +1,312 @@
...
+                        """,
+                        (incident_id,),
+                    )
+                    citation_rows = cursor.fetchall()
+                    citations = [
+                        {
+                            "key": row[0],
+                            "tool_name": row[1],
+                            "command": row[2],
+                            "output": row[3][:500] if row[3] else "",  # Truncate output
```
damianloch, line 182: why this truncation? Just trying to understnad the logic behind this number choice.

client/src/lib/services/incident-feedback.ts
```diff
@@ -0,0 +1,80 @@
...
+  comment?: string;
+  createdAt: string;
+}
+
+export interface SubmitFeedbackResponse {
+  success: boolean;
+  feedback_id: string;
+  feedback_type: FeedbackType;
+  stored_for_learning: boolean;
+}
```
damianloch, line 32: not a huge deal but there is inconsistencies since both camel case and snake case used here

client/src/lib/services/user-preferences.ts
damianloch, line 1: This file is near-identical to incident-feedback.ts. Error handling, headers, and export 
patterns are duplicated. Extract shared API client logic to avoid maintaining the same 
code in multiple files.

server/chat/background/rca_prompt_builder.py
```diff
@@ -12,6 +16,149 @@
...
+
+
+def _is_aurora_learn_enabled(user_id: str) -> bool:
+    """Check if Aurora Learn is enabled for a user. Defaults to True."""
+    if not user_id:
+        return False
+    try:
+        from utils.auth.stateless_auth import get_user_preference
+        setting = get_user_preference(user_id, "aurora_learn_enabled", default=True)
+        return setting is True or setting == "true"
```
damianloch: We should only be using the bool or string, not both unless we return both types in different instances

server/chat/background/rca_prompt_builder.py
```diff
@@ -12,6 +16,149 @@
...
+        for i, match in enumerate(matches, 1):
+            similarity_pct = int(match["similarity"] * 100)
+            context_parts.extend([
+                f"### Past Incident {i} (Similarity: {similarity_pct}%)",
+                f"- **Alert**: {match.get('alert_title', 'Unknown')}",
+                f"- **Service**: {match.get('alert_service', 'Unknown')}",
+                f"- **Source**: {match.get('source_type', 'Unknown')}",
+                "",
+                "**Summary of what was found:**",
+                match.get("aurora_summary", "No summary available")[:1000],  # Limit length
```
damianloch, line 123: was there logic behind this and other truncation limits within this file (just to know)

server/routes/incident_feedback/routes.py
isiddharthsingh: removed the or setting == "true" since we validate it's a boolean before storing in set_aurora_learn_setting()

client/src/lib/services/user-preferences.ts
isiddharthsingh: removed this file

server/routes/incident_feedback/routes.py
isiddharthsingh: citation outputs can be huge (kubectl dumps, log files, etc). truncating to 500 chars prevents bloating weaviate storage and keeps context injection manageable. the number is somewhat arbitrary but keeps it reasonable without losing key info.

client/src/lib/services/incident-feedback.ts
isiddharthsingh: fixed, updated backend response and frontend interface to use camelCase consistently

server/chat/background/rca_prompt_builder.py
isiddharthsingh: fixed, removed the string check since we always store it as a boolean

server/chat/background/rca_prompt_builder.py
isiddharthsingh: yeah it's to prevent token bloat when injecting context into the llm prompt. summaries get 1000 chars, thoughts get 200 chars - keeps the "similar past incidents" section from overwhelming the main rca prompt

client/src/app/incidents/components/IncidentFeedback.tsx
isiddharthsingh: added error_code: "AURORA_LEARN_DISABLED" to the backend response and updated frontend to check apiError.code instead of string matching

client/src/app/incidents/components/IncidentFeedback.tsx
isiddharthsingh: no we are not sending any fake data

server/routes/incident_feedback/weaviate_client.py
isiddharthsingh: added threading.Lock() with double-checked locking - fast path without lock when client is ready, lock only acquired when reconnecting

server/routes/incident_feedback/weaviate_client.py
isiddharthsingh: now logs the error with logger.warning()

server/routes/incident_feedback/routes.py
isiddharthsingh: the early returns are for validation errors before any db writes (missing user_id, invalid uuid, duplicate feedback check, etc) so no data to rollback there. the transaction starts with the INSERT, and if weaviate fails we explicitly rollback. for not_helpful we commit right after insert. psycopg2 auto-rollbacks uncommitted transactions when the connection closes, but let me know if im missing anything or edge cases

PR #25, opened 2026-01-27
Feat/rca visualization streaming

damianloch:
Add progressive RCA visualization with streaming updates

This adds infrastructure visualization to RCA incidents that updates in real-time as the investigation progresses. The visualization shows entities (services, pods, databases, alerts) and their relationships as a graph, helping SREs understand the incident state visually. Only the backend is implemented so far (so up until dynamic generation of the visualization during an RCA investigation). Nothing is changed in the frontend.

Key changes:
- Visualization updates every 30 seconds during RCA + final comprehensive update at the end
- LLM (Claude Sonnet) extracts infrastructure entities from tool outputs into structured graph data
- SSE endpoint streams updates to frontend with auth and proper Redis cleanup
- Database schema extended with visualization_code and visualization_updated_at columns
- Incremental extraction only processes recent tool calls to minimize cost/latency

Technical details:
- New Celery task update_visualization handles extraction and rendering
- Smart debouncing with per-incident trigger instances prevents duplicate updates
- Tool outputs truncated to 5KB to manage token limits
- All existing nodes/edges passed to LLM for context-aware incremental updates
- Proper error handling and resource cleanup throughout


server/chat/background/visualization_extractor.py
```diff
@@ -0,0 +1,136 @@
...
+    def _merge(self, existing: VisualizationData, new: VisualizationData) -> VisualizationData:
+        """Merge new entities with existing ones."""
+        merged_nodes = {n.id: n for n in existing.nodes}
+        
+        for node in new.nodes:
+            if node.id in merged_nodes:
+                # Update status if new info is more specific
+                if node.status != 'investigating' and merged_nodes[node.id].status == 'investigating':
+                    merged_nodes[node.id].status = node.status
+                merged_nodes[node.id].metadata.update(node.metadata)
```
Zarlanx: Infranodes don't have a metadata field currently

server/chat/background/visualization_extractor.py
```diff
@@ -0,0 +1,136 @@
+"""Extracts infrastructure entities from RCA transcripts for visualization."""
+import logging
+from datetime import datetime
+from typing import Any, Dict, List, Literal, Optional
+from pydantic import BaseModel, Field
+from chat.backend.agent.llm import LLMManager
```
Zarlanx: Is this import being used?

server/chat/backend/agent/tools/cloud_tools.py
```diff
@@ -28,6 +28,10 @@
 from .github_commit_tool import github_commit, GitHubCommitArgs
 from .github_rca_tool import github_rca, GitHubRCAArgs
 from .github_fix_tool import github_fix, GitHubFixArgs
+
+# Visualization trigger caching
+_viz_triggers: Dict[str, Any] = {}
```
Zarlanx: The dict lives for the lifetime of the worker process without ever being cleaned up leading to unbounded growth, either clean it up or use bounded/TTL cache

server/chat/background/task.py
```diff
@@ -24,10 +24,34 @@
...
 
 
 logger = logging.getLogger(__name__)
 
+# Constants
+MAX_TOOL_OUTPUT_CHARS = 5000
+INFRASTRUCTURE_TOOLS = frozenset(['on_prem_kubectl', 'cloud_exec', 'terminal_exec', 'tailscale_ssh'])
+
+
+def _extract_tool_calls_for_viz() -> List[Dict]:
```
Zarlanx: _extract_tool_calls_for_viz() is called from run_background_chat() the sync Celery wrapper after asyncio.run(_execute_background_chat(...)) returns. set_user_context(state=state) was called inside the async function, which ran in its own event loop hence by the time we get here, the contextvar is stale or gone which will return [] every time, meaning the force_full=True final visualization never gets tool calls from memory and falls back to the DB fetch path.

server/routes/visualization_stream.py
```diff
@@ -0,0 +1,97 @@
...
+        pubsub = None
+        try:
+            redis_client = redis.from_url(os.getenv('REDIS_URL', 'redis://redis:6379/0'))
+            pubsub = redis_client.pubsub()
+            channel = f"visualization:{incident_id}"
+            pubsub.subscribe(channel)
+            
+            yield f"data: {json.dumps({'type': 'connected', 'incident_id': incident_id})}\n\n"
+            
+            for message in pubsub.listen():
```
Zarlanx: Blocks forever with no timeout or heartbeat. If the client disconnects silently (no GeneratorExit), this thread sits idle permanently.

server/chat/background/visualization_generator.py
```diff
@@ -0,0 +1,180 @@
...
+    except Exception as e:
+        logger.error(f"[Visualization] Failed to store viz: {e}")
+        raise
+
+
+def _notify_sse_clients(incident_id: str, version: int):
+    """Notify SSE listeners via Redis pub/sub."""
+    redis_client = None
+    try:
+        redis_client = redis.from_url(os.getenv('REDIS_URL', 'redis://redis:6379/0'))
```
Zarlanx: Creates and destroys a Redis connection on every notification. The codebase already has get_redis_client() from utils.cache.redis_client should reuse that.

server/chat/background/visualization_extractor.py
```diff
@@ -0,0 +1,136 @@
...
+
+logger = logging.getLogger(__name__)
+
+# Tool output truncation limit (matches task.py constant)
+MAX_TOOL_OUTPUT_CHARS = 5000
```
Zarlanx: This constant is duplicated in 4 files: cloud_tools.py, task.py, visualization_extractor.py, visualization_generator.py. Should be defined once and imported.

server/chat/background/visualization_generator.py
```diff
@@ -0,0 +1,180 @@
...
+        else:
+            tool_calls = _fetch_recent_tool_calls(session_id, user_id, limit=10 if not force_full else 50)
+            logger.info(f"[Visualization] Fetched {len(tool_calls)} tool calls from state context")
+        
+        if not tool_calls:
+            return {"status": "skipped", "reason": "no_tool_calls"}
+        
+        existing_viz = _fetch_existing_visualization(incident_id)
+        
+        extractor = VisualizationExtractor()
```
Zarlanx: Instantiates a new VisualizationExtractor (which calls create_chat_model()) on every Celery task invocation. That's a new LLM client every 30 seconds per active incident maybe use a module-level singleton or at least try caching the LLM client.

server/chat/background/visualization_extractor.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+class VisualizationData(BaseModel):
+    """Complete visualization state."""
+    nodes: List[InfraNode] = Field(default_factory=list)
+    edges: List[InfraEdge] = Field(default_factory=list)
+    rootCauseId: Optional[str] = Field(default=None, description="Node ID of root cause")
+    affectedIds: List[str] = Field(default_factory=list, description="Affected node IDs")
+    version: int = Field(default=1, description="Incremented on each update")
+    updatedAt: str = Field(default_factory=lambda: datetime.utcnow().isoformat())
```
Zarlanx: datetime.utcnow() is deprecated in Python 3.12+, use datetime.now(datetime.timezone.utc) instead.

server/chat/background/visualization_extractor.py
```diff
@@ -0,0 +1,136 @@
...
+        prompt = self._build_prompt(recent_messages, existing_viz)
+        
+        try:
+            extractor = self.llm.with_structured_output(VisualizationData)
+            new_viz = extractor.invoke(prompt)
+            
+            if existing_viz:
+                merged = self._merge(existing_viz, new_viz)
+                merged.version = existing_viz.version + 1
+                merged.updatedAt = datetime.utcnow().isoformat()
```
Zarlanx: datetime.utcnow() is deprecated in Python 3.12+, use datetime.now(datetime.timezone.utc) instead.

server/chat/background/visualization_generator.py
```diff
@@ -0,0 +1,180 @@
...
+    """Store updated visualization in database."""
+    try:
+        with db_pool.get_admin_connection() as conn:
+            with conn.cursor() as cursor:
+                cursor.execute("""
+                    UPDATE incidents
+                    SET visualization_code = %s,
+                        visualization_updated_at = %s
+                    WHERE id = %s
+                """, (json_str, datetime.utcnow(), incident_id))
```
Zarlanx: datetime.utcnow() is deprecated in Python 3.12+, use datetime.now(datetime.timezone.utc) instead.

Noahcasarotto:
**UX Feedback: Pan/Zoom Navigation**

The diagram doesn't support click-and-drag panning. Users should be able to navigate around the visualization like Google Maps - e.g., look at the full graph, then zoom into a specific area, then pan/scroll to explore different sections.

Currently it feels static. Adding proper pan controls would make it much more usable for larger infrastructure graphs.

server/chat/background/visualization_extractor.py
damianloch: Thanks for pointing this out removed it

server/chat/background/visualization_extractor.py
damianloch: Was not being used removed it

server/chat/backend/agent/tools/cloud_tools.py
damianloch: uses TTL cache now

server/chat/background/task.py
damianloch: Fixed, now extracting tool_calls inside the async function before the context is lost, then passing them back via the return value.

server/routes/visualization_stream.py
damianloch: added a timeout

server/chat/background/visualization_generator.py
damianloch: switched to use get_redis_client

server/chat/background/visualization_generator.py
damianloch: Fixed by using a module-level singleton VisualizationExtractor instead of creating a new instance (and LLM client) on every task invocation

server/chat/background/visualization_extractor.py
damianloch: Fixed deprecated datetime.utcnow() calls by replacing with datetime.now(timezone.utc) for Python 3.12+ compatibility.

server/chat/background/visualization_extractor.py
damianloch: Fixed deprecated datetime.utcnow() calls by replacing with datetime.now(timezone.utc) for Python 3.12+ compatibility.

server/chat/background/visualization_generator.py
damianloch: Fixed deprecated datetime.utcnow() calls by replacing with datetime.now(timezone.utc) for Python 3.12+ compatibility.

server/chat/background/visualization_extractor.py
damianloch: Consolidated duplicate MAX_TOOL_OUTPUT_CHARS and INFRASTRUCTURE_TOOLS constants into a shared chat.backend.constants module

PR #27, opened 2026-01-27
Feat/confluence cql search

Zarlanx:
# Changes:
### Adds 
- Confluence CQL search tools so the RCA agent can automatically find and reference relevant postmortems, runbooks, and past incident docs during background investigations.

### Connector layer (server/connectors/confluence_connector/)
- auth.py: added search:confluence to OAuth scopes
- client.py: added search_content() method using v1 /content/search API (no v2 equivalent for CQL)
- cql_builder.py (new): pure-function CQL query builders with error message cleaning (strips UUIDs, IPs, timestamps, hex, base64 for stable matching)
- search_service.py (new): ConfluenceSearchService with automatic OAuth token refresh on 401 (follows existing runbook_utils.py pattern)
Agent tools (server/chat/backend/agent/tools/)
- confluence_search_tool.py (new): 3 tool functions with Pydantic arg schemas
- cloud_tools.py: registers the 3 tools, gated behind is_confluence_enabled() feature flag
### RCA integration
- prompt_builder.py: added Confluence search instructions to both build_system_invariant() and build_background_mode_segment()
- task.py: added Confluence to _get_connected_integrations() so the prompt builder knows when tools are available

### DIFF
--- a/server/chat/backend/agent/prompt/prompt_builder.py
+++ b/server/chat/backend/agent/prompt/prompt_builder.py
@@ -1452,6 +1452,20 @@ def build_background_mode_segment(state: Optional[Any]) -> str:
             "Look for: config changes, k8s manifests, Terraform, dependency updates.",
         ])
 
+    # Confluence search tools (if connected)
+    if integrations.get('confluence'):
+        parts.extend([
+            "",
+            "CONFLUENCE INVESTIGATION:",
+            "Use Confluence search tools to find prior incidents and runbooks:",
+            "- confluence_search_similar(keywords=['error msg'], service_name='svc') - Find postmortems / past incidents",
+            "- confluence_search_runbooks(service_name='svc') - Find runbooks / SOPs / playbooks",
+            "- confluence_fetch_page(page_id='12345') - Read full page content as markdown",
+            "",
+            "Workflow: search first, then fetch promising pages for detailed procedures.",
+            "Cross-reference Confluence findings with live infrastructure state.",
+        ])
+
     # Knowledge Base search (always available for authenticated users)
     parts.extend([
         "",
--- a/server/chat/background/task.py
+++ b/server/chat/background/task.py
@@ -76,6 +76,7 @@ def _get_connected_integrations(user_id: str) -> Dict[str, bool]:
     integrations = {
         'splunk': False,
         'github': False,
+        'confluence': False,
     }
 
     try:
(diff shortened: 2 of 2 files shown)


server/chat/backend/agent/tools/confluence_search_tool.py
```diff
@@ -0,0 +1,147 @@
...
+    session_id: Optional[str] = None,
+) -> str:
+    """Fetch a Confluence page by ID and return its markdown content."""
+    _ = session_id
+    if not user_id:
+        raise ValueError("user_id is required for Confluence page fetch")
+
+    try:
+        svc = ConfluenceSearchService(user_id)
+        result = svc.fetch_page_markdown(page_id, max_length=max_length)
```
OlivierTrudeau, line 140: Would it be a good idea to add some sort of offset? Like if the page is more than max_lenght, maybe the LLM realizes it needs more info and instead of remaking the request but with just a massive max_lenght, it could search the runbook itself? At the same time I thought this was what pages where for no?

server/chat/backend/agent/tools/confluence_search_tool.py
```diff
@@ -0,0 +1,147 @@
...
+
+def confluence_search_runbooks(
+    service_name: str,
+    operation: Optional[str] = None,
+    spaces: Optional[List[str]] = None,
+    user_id: Optional[str] = None,
+    session_id: Optional[str] = None,
+) -> str:
+    """Search Confluence for runbooks related to a service."""
+    _ = session_id
```
OlivierTrudeau, line 102: Actually just curious of why to do this like this? Can't we just remove it from the function parameters if we don't need it?

server/chat/backend/agent/tools/confluence_search_tool.py
Zarlanx: Not worth it right now, each extra fetch_page call burns an agent iteration (out of 240) and adds more tool output to a context window that already hits model limits on long RCA runs (we just shipped a context-trim middleware in #29 for exactly this). If we do need pagination later, adding an offset param to fetch_page_markdown() is trivial, but fewer tool knobs = more reliable LLM usage, so I'd rather not add it preemptively.

server/chat/backend/agent/tools/confluence_search_tool.py
Zarlanx: It's because of how tool injection works in our agent framework. The `with_user_context` and `with_forced_context` decorators in `cloud_tools.py` (lines 419-510) always inject **both** `user_id` and `session_id` into kwargs:

If the function signature doesn't accept `session_id`, the decorator call crashes with `TypeError: unexpected keyword argument 'session_id'`. Every tool in the codebase follows this pattern, check `confluence_runbook_tool.py` (line 14: `# user_id/session_id are injected by with_user_context`), `rag_indexer_tool.py`, `web_search_tool.py`, etc.

The `_ = session_id` is a Python convention to tell linters "yes I know this is unused, it's intentional." The parameter must exist in the signature; we just don't need its value in these particular tools.

PR #29, opened 2026-01-28
Add context trim middleware to prevent context overflow during long RCA runs

Zarlanx:
Implements automatic context trimming to manage token usage and prevent context window overflow when processing extended root cause analysis sessions. Includes middleware integration and context manager updates.

### DIFF
--- a/server/chat/backend/agent/middleware/context_trim.py
+++ b/server/chat/backend/agent/middleware/context_trim.py
@@ -1,0 +1,84 @@
+"""
+Context trimming middleware for LangChain create_agent.
+
+Prevents context overflow during long-running ReAct loops (e.g., RCA investigations
+with 240 recursion limit) by trimming messages before each LLM call. Only affects
+what the LLM sees — the full message history is preserved in graph state.
+"""
+
+import logging
+from typing import Awaitable, Callable
+
+from langchain.agents.middleware import AgentMiddleware
+from langchain.agents.middleware.types import ModelRequest, ModelResponse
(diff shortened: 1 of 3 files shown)


server/chat/backend/agent/middleware/context_trim.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def __init__(self, model_name: str):
+        self.model_name = model_name
+        self.max_tokens = int(
+            ChatContextManager.get_context_limit(model_name) * _CONTEXT_USAGE_RATIO
+        )
+        logger.info(
+            f"ContextTrimMiddleware initialized: model={model_name}, "
+            f"max_tokens={self.max_tokens}"
+        )
```
OlivierTrudeau, line 49: Could change this on debug level

server/chat/backend/agent/middleware/context_trim.py
```diff
@@ -0,0 +1,84 @@
...
+            end_on=("human", "tool"),
+        )
+
+        trimmed_count = len(trimmed)
+        trimmed_tokens = count_tokens_approximately(trimmed)
+
+        logger.info(
+            f"Context trimmed: {original_count} -> {trimmed_count} messages, "
+            f"~{estimated_tokens} -> ~{trimmed_tokens} tokens"
+        )
```
OlivierTrudeau, line 82: Same comment of debug for this one but this one is a little more informative so maybe could keep you tell me

PR #32, opened 2026-01-29
Feat/rca visualization frontend

damianloch:
Infrastructure Visualization for RCA Incidents

Adds real-time visual graph showing infrastructure entities and their relationships during investigations.

**Features:**
- Live SSE streaming with interactive graph (drag, zoom, fullscreen)
- Smart entity extraction from tool outputs with causality chains
- Status indicators (failed/degraded/healthy) highlighting root cause
- Parent-child relationships (pods → deployments)
- Icon-based node types

### DIFF
--- a/client/src/components/incidents/InfrastructureVisualization.tsx
+++ b/client/src/components/incidents/InfrastructureVisualization.tsx
@@ -484,14 +484,14 @@ export default function InfrastructureVisualization({ incidentId, className }: P
         elementsSelectable={true}
         panOnDrag={true}
         selectionOnDrag={false}
-        panOnScroll={true}
+        panOnScroll={false}
         panOnScrollMode="free"
-        zoomOnScroll={false}
+        zoomOnScroll={true}
         zoomOnPinch={true}
         zoomOnDoubleClick={true}
         fitView
-        minZoom={0.5}
-        maxZoom={1.5}
+        minZoom={0.1}
+        maxZoom={2}
         defaultEdgeOptions={{ 
           type: 'smoothstep',
           animated: false,
--- a/server/chat/background/task.py
+++ b/server/chat/background/task.py
@@ -33,21 +33,40 @@ from chat.backend.constants import MAX_TOOL_OUTPUT_CHARS, INFRASTRUCTURE_TOOLS
 logger = logging.getLogger(__name__)
 
 
-def _extract_tool_calls_for_viz() -> List[Dict]:
-    """Extract infrastructure tool calls from current state for final visualization."""
-    state = get_state_context()
-    if not state or not hasattr(state, 'messages'):
+def _extract_tool_calls_for_viz(session_id: str, user_id: str) -> List[Dict]:
+    """Extract infrastructure tool calls from database for final visualization."""
+    try:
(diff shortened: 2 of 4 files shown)


client/next.config.ts
damianloch, line 1: For @xyflow/react (ReactFlow library) I needed to do this. It prevents webpack from trying to polyfill web workers, which can cause build issues with ReactFlow.

client/src/app/api/incidents/[id]/visualization/stream/route.ts
```diff
@@ -0,0 +1,33 @@
...
+
+    const { id } = await params;
+
+    const response = await fetch(`${API_BASE_URL}/api/incidents/${id}/visualization/stream`, {
+      method: 'GET',
+      headers: authResult.headers,
+      credentials: 'include',
+    });
+
+    if (!response.ok) return new Response('Failed to connect to visualization stream', { status: response.status });
```
Comment (beng360) [inline client/src/app/api/incidents/[id]/visualization/stream/route.ts:24]: should log this like other routes

client/src/app/api/incidents/[id]/visualization/stream/route.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export async function GET(
+  request: NextRequest,
+  { params }: { params: { id: string } }
+) {
+  try {
+    if (!API_BASE_URL) return new Response('BACKEND_URL not configured', { status: 500 });
+
+    const authResult = await getAuthenticatedUser();
+    if (authResult instanceof Response) return authResult;
```
Comment (beng360) [inline client/src/app/api/incidents/[id]/visualization/stream/route.ts]: export async function getAuthenticatedUser(): Promise<AuthResult | NextResponse> {  
  const session = await auth()
  
  if (!session?.userId) {
    return NextResponse.json({ error: 'Unauthorized' }, { status: 401 })
  }

  return {
    userId: session.userId,
    headers: {
      'X-User-ID': session.userId,
    }
  }
}

getAuthenticateduser() doesnt return Response

Comment (beng360) [reply client/src/app/api/incidents/[id]/visualization/stream/route.ts]: check for nextReponse instead

client/src/hooks/useVisualizationStream.ts
```diff
@@ -0,0 +1,119 @@
...
+        } catch (err) {
+          console.error('[Visualization] Failed to parse SSE message:', err);
+        }
+      };
+
+      eventSource.onerror = () => {
+        if (mounted) {
+          setState(prev => ({ ...prev, isConnected: false }));
+        }
+      };
```
beng360, line 103: do we want to try and reconnect SSE or just mark it as disconnected

damianloch: The browser's EventSource API automatically reconnects in the background

client/src/app/api/incidents/[id]/visualization/stream/route.ts
Comment (damianloch) [reply client/src/app/api/incidents/[id]/visualization/stream/route.ts]: Fixed this

client/src/app/api/incidents/[id]/visualization/stream/route.ts
Comment (damianloch) [reply client/src/app/api/incidents/[id]/visualization/stream/route.ts:24]: Fixed this

PR #33, opened 2026-01-29
Oliverek/dev 743 deployable oss

OlivierTrudeau:
Extensive documentation and setup for deploying the app on any kubernetes cluster. More elaborate instructions for GKE specifically

### DIFF
--- a/deploy/helm/aurora/templates/server-deployment.yaml
+++ b/deploy/helm/aurora/templates/server-deployment.yaml
@@ -64,18 +64,18 @@ spec:
             httpGet:
               path: /health/readiness
               port: 5080
-            initialDelaySeconds: 10
+            initialDelaySeconds: 20
             periodSeconds: 10
             timeoutSeconds: 5
-            failureThreshold: 6
+            failureThreshold: 3
           livenessProbe:
             httpGet:
               path: /health/liveness
               port: 5080
-            initialDelaySeconds: 20
+            initialDelaySeconds: 60
             periodSeconds: 20
             timeoutSeconds: 5
-            failureThreshold: 6
+            failureThreshold: 3
           resources:
             {{- toYaml .Values.resources.server | nindent 12 }}
       volumes:
--- a/deploy/helm/aurora/values.yaml
+++ b/deploy/helm/aurora/values.yaml
@@ -253,11 +253,11 @@ secrets:
 resources:
   server:
     requests:
-      cpu: "100m"
+      cpu: "200m"
       memory: "512Mi"
     limits:
-      cpu: "500m"
-      memory: "1Gi"
+      cpu: "2000m"
+      memory: "2Gi"
   celeryWorker:
     requests:
       cpu: "100m"


deploy/helm/aurora/templates/secret.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ include "aurora.fullname" . }}-secrets
+  labels:
+    {{- include "aurora.labels" . | nindent 4 }}
+type: Opaque
+stringData:
+{{- range $key, $value := .Values.secrets }}
```
Zarlanx: This loops over **all** `.Values.secrets` and dumps them into one Secret. Every deployment that uses `secretRef: aurora-oss-secrets` gets every secret, the frontend pod sees `POSTGRES_PASSWORD`, `VAULT_TOKEN`, `STORAGE_SECRET_KEY`, etc. If the frontend container is compromised, the attacker gets DB creds, Vault root token, and S3 keys.

**Suggestion**: Split into scoped secrets:
```yaml
# secret-db.yaml — only mounted by server, celery, chatbot
POSTGRES_USER, POSTGRES_PASSWORD

# secret-vault.yaml — only mounted by server, chatbot
VAULT_TOKEN

# secret-storage.yaml — only mounted by server, celery
STORAGE_ACCESS_KEY, STORAGE_SECRET_KEY

# secret-app.yaml — shared app secrets
FLASK_SECRET_KEY, AUTH_SECRET, SEARXNG_SECRET

# secret-llm.yaml — only mounted by chatbot, celery
OPENROUTER_API_KEY, OPENAI_API_KEY, ANTHROPIC_API_KEY, GOOGLE_API_KEY
```
Such that each deployment mounts only the secrets it needs.

deploy/helm/aurora/templates/_helpers.tpl
```diff
@@ -0,0 +1,18 @@
+{{- define "aurora.name" -}}
+{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" -}}
+{{- end }}
+
+{{- define "aurora.fullname" -}}
+{{- if .Values.fullnameOverride -}}
+{{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" -}}
+{{- else -}}
+{{- printf "%s" (include "aurora.name" .) | trunc 63 | trimSuffix "-" -}}
```
Zarlanx: `aurora.name` ignores `.Release.Name` This just returns the chart name (`aurora-oss`), ignoring the release name entirely. If someone installs two releases (`helm install staging ./aurora` and `helm install prod ./aurora`), both generate identical resource names and collide.

The standard Helm pattern is:
```
{{- printf "%s-%s" .Release.Name (include "aurora.name" .) | trunc 63 | trimSuffix "-" -}}
```

Or use `helm create`'s default which checks if the release name already contains the chart name:
```
{{- if contains (include "aurora.name" .) .Release.Name -}}
{{- .Release.Name | trunc 63 | trimSuffix "-" -}}
{{- else -}}
{{- printf "%s-%s" .Release.Name (include "aurora.name" .) | trunc 63 | trimSuffix "-" -}}
{{- end -}}
```

deploy/helm/aurora/templates/vault-statefulset.yaml
```diff
@@ -0,0 +1,85 @@
...
+            - name: vault-data
+              mountPath: /vault/data
+            {{- if and .Values.vault.seal (eq .Values.vault.seal.type "gcpckms") .Values.vault.seal.gcpckms.credentials }}
+            - name: gcp-kms-credentials
+              mountPath: /vault/gcp
+              readOnly: true
+            {{- end }}
+          readinessProbe:
+            exec:
+              command: ["/bin/sh", "-c", "vault status || exit 0"]
```
Zarlanx: `vault status || exit 0` makes probes useless as both readiness and liveness probes use `vault status || exit 0`. The `|| exit 0` means the probe **always succeeds** even when the vault is sealed, initializing, or erroring. Kubernetes will route traffic to a sealed Vault and never restart a stuck one.

**Readiness** should fail when sealed (so no traffic is routed):
```yaml
readinessProbe:
  exec:
    command:
      - /bin/sh
      - -c
      - vault status -format=json | grep '"sealed": false'
  initialDelaySeconds: 10
  periodSeconds: 10
```

**Liveness** can be more lenient (Vault is running but maybe sealed):
```yaml
livenessProbe:
  httpGet:
    path: /v1/sys/health?standbyok=true&sealedcode=200&uninitcode=200
    port: 8200
  initialDelaySeconds: 30
  periodSeconds: 20
```

The HTTP endpoint with `sealedcode=200` lets the liveness probe pass while sealed (so K8s doesn't kill it during unseal), while the readiness probe correctly reports "not ready" until unsealed.

deploy/helm/aurora/templates/redis-statefulset.yaml
```diff
@@ -0,0 +1,30 @@
...
+    metadata:
+      annotations:
+        checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
+      labels:
+        app.kubernetes.io/name: {{ include "aurora.name" . }}-redis
+        app.kubernetes.io/instance: {{ .Release.Name }}
+        app.kubernetes.io/part-of: aurora
+    spec:
+      containers:
+        - name: redis
```
Zarlanx, line 29: Redis is a plain Deployment with no PersistentVolumeClaim, if the pod restarts, all Celery task state and cached data is lost. For Aurora this means background tasks may be silently dropped.

No readiness or liveness probe: K8s can't detect if Redis is unhealthy, and the pod will receive traffic immediately on start even if Redis isn't ready.

For persistence, either:
- Convert to StatefulSet with volumeClaimTemplates (like Postgres/Vault), or
- Add a PVC + volumeMount for `/data` with `redis-server --appendonly yes`

deploy/helm/aurora/templates/weaviate-deployment.yaml
```diff
@@ -0,0 +1,64 @@
+apiVersion: apps/v1
+kind: Deployment
```
Zarlanx: Deployment + static PVC breaks with replicas > 1. If Weaviate uses a Deployment referencing a single PVC (`weaviate-pvc.yaml` with `ReadWriteOnce`). If someone sets `replicaCounts.weaviate: 2`, both pods try to mount the same RWO PVC with only one succeeding and the other stays Pending forever.

deploy/helm/aurora/values.yaml
```diff
@@ -0,0 +1,361 @@
...
+# ------------------------------------------------------------------------------
+image:
+  # Container registry where images are stored
+  registry: "your-registry"  #Examples: docker.io/myorg, gcr.io/project-id, ghcr.io/myorg
+  
+  # Image tag (version) to deploy
+  tag: "latest"
+  
+  # Optional: Pull secrets for private registries
+  # pullSecrets:
```
Zarlanx, line 57: `pullSecrets` defined but never used in templates, this is commented out in the defaults (fine), but no template actually references `imagePullSecrets`. Even if someone uncomments this and sets it in `values.generated.yaml`, images from private registries will fail to pull.

deploy/helm/aurora/templates/server-deployment.yaml
```diff
@@ -0,0 +1,59 @@
...
+  template:
+    metadata:
+      annotations:
+        checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
+        checksum/secret: {{ include (print $.Template.BasePath "/secret.yaml") . | sha256sum }}
+      labels:
+        app.kubernetes.io/name: {{ include "aurora.name" . }}-server
+        app.kubernetes.io/instance: {{ .Release.Name }}
+        app.kubernetes.io/part-of: aurora
+    spec:
```
Zarlanx, line 24: No RBAC resources (ServiceAccount, Role, RoleBinding), all pods run as the `default` ServiceAccount in the namespace. This is an issue because:

1. The `default` SA may have broad permissions depending on cluster config
2. No way to scope permissions per-service (e.g., only the server needs Vault access)
3. Workload Identity (GKE) requires annotated ServiceAccounts — the Vault KMS docs reference `aurora-oss-vault` SA but no template creates it

**Suggestion**: Add a `serviceaccount.yaml` template (or per-service SAs) and reference them in pod specs:
```yaml
spec:
  serviceAccountName: {{ include "aurora.fullname" . }}-server
```

deploy/helm/aurora/templates/server-deployment.yaml
```diff
@@ -0,0 +1,59 @@
...
+      containers:
```
Zarlanx, line 37: ### Should-Fix: No `securityContext` on any pod

None of the deployments or statefulsets define a `securityContext`. This means containers may run as root (depending on the image), which is a security risk and violates Pod Security Standards.

**Suggestion** — add to each pod spec:
```yaml
spec:
  securityContext:
    runAsNonRoot: true
    fsGroup: 1000
  containers:
    - name: aurora-server
      securityContext:
        allowPrivilegeEscalation: false
        readOnlyRootFilesystem: true  # where feasible
        capabilities:
          drop: ["ALL"]
```
Note: Vault needs `IPC_LOCK` capability if mlock is enabled (but `disable_mlock = true` in your config, so it should be fine without it).

deploy/helm/aurora/templates/frontend-deployment.yaml
```diff
@@ -0,0 +1,38 @@
...
+        app.kubernetes.io/instance: {{ .Release.Name }}
+        app.kubernetes.io/part-of: aurora
+    spec:
+      containers:
+        - name: aurora-frontend
+          image: "{{ .Values.image.registry }}/aurora-frontend:{{ .Values.image.tag }}"
+          imagePullPolicy: IfNotPresent
+          ports:
+            - name: http
+              containerPort: 3000
```
Zarlanx, line 44: ### Should-Fix: Frontend has no health probes

No readiness or liveness probe. The frontend pod will receive traffic immediately, even if Next.js hasn't finished starting.

**Suggestion**:
```yaml
readinessProbe:
  httpGet:
    path: /
    port: 3000
  initialDelaySeconds: 10
  periodSeconds: 10
livenessProbe:
  httpGet:
    path: /
    port: 3000
  initialDelaySeconds: 20
  periodSeconds: 20
```

deploy/helm/aurora/templates/celery-beat-deployment.yaml
```diff
@@ -0,0 +1,34 @@
...
+      annotations:
+        checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
+        checksum/secret: {{ include (print $.Template.BasePath "/secret.yaml") . | sha256sum }}
+      labels:
+        app.kubernetes.io/name: {{ include "aurora.name" . }}-celery-beat
+        app.kubernetes.io/instance: {{ .Release.Name }}
+        app.kubernetes.io/part-of: aurora
+    spec:
+      containers:
+        - name: celery-beat
```
Zarlanx, line 37: ### Celery beat has no health probes

No readiness or liveness probe for celery-beat. If the beat scheduler hangs or crashes silently, K8s won't detect or restart it, and scheduled tasks stop running.

**Suggestion**: celery-beat doesn't serve HTTP, but you can use a file-based liveness probe:
```yaml
livenessProbe:
  exec:
    command:
      - /bin/sh
      - -c
      - "celery -A celery_config inspect ping -d celery@$HOSTNAME 2>/dev/null | grep -q pong"
  initialDelaySeconds: 30
  periodSeconds: 60
  timeoutSeconds: 10
```
Or use the `--pidfile` approach and check if the process is alive.

deploy/helm/aurora/templates/namespace.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: {{ .Release.Namespace }}
```
Zarlanx: ### Redundant namespace template

The `Makefile` already runs `helm upgrade --install ... --namespace aurora --create-namespace`, which creates the namespace automatically. This template is redundant and can cause issues:

1. `helm template` fails if the namespace already exists via `--create-namespace`
2. `helm uninstall` won't delete the namespace (Helm doesn't delete namespaces it didn't create)
3. If the namespace was created externally (e.g., by GitOps), this template may conflict

**Suggestion**: Delete this file. Let `--create-namespace` handle it.

deploy/helm/aurora/values.yaml
```diff
@@ -0,0 +1,361 @@
...
+  RCA_OPTIMIZE_COSTS: "false"
+  GEMINI_DISABLE_THINKING: "false"
+  
+  # --- Security ---
+  RATE_LIMITING_ENABLED: "false"
+  RATE_LIMIT_HEADERS_ENABLED: "true"
+  AURORA_VERIFY_CLI_IDENTITY: "false"
+  
+  # --- Optional Features (enable as needed) ---
+  ENABLE_POD_ISOLATION: "false"              # Isolate conversations in separate pods should be true for production
```
Zarlanx: ### `ENABLE_POD_ISOLATION` exposed but not implemented

This config value is passed through to the ConfigMap, but the chart has no templates that actually implement pod isolation with no RBAC, no NetworkPolicies, no separate namespace for untrusted pods, no RuntimeClass enforcement.

The related `TERMINAL_NAMESPACE`, `TERMINAL_RUNTIME_CLASS`, and `USE_UNTRUSTED_NODES` values below are also just passed as env vars with no corresponding K8s resources.

Either:
1. Add the actual isolation resources (NetworkPolicy, RBAC, RuntimeClass) gated on this flag, or
2. Remove these config values from the chart and document that pod isolation requires manual setup

deploy/helm/aurora/templates/frontend-deployment.yaml
```diff
@@ -0,0 +1,38 @@
...
+          envFrom:
+            - configMapRef:
+                name: {{ include "aurora.fullname" . }}-config
+            - secretRef:
```
Zarlanx, line 48: ### Frontend receives all secrets via `secretRef`

Related to the secret.yaml comment, the frontend container gets `POSTGRES_PASSWORD`, `VAULT_TOKEN`, `STORAGE_SECRET_KEY`, etc. via `envFrom`. The frontend only needs `AUTH_SECRET` and maybe `FLASK_SECRET_KEY`. This is also true for celery-beat, searxng, and other services that mount the full secret.

Zarlanx (changes requested):
## Helm Chart Review

Good start — the chart structure is solid, values are well-documented, the Vault KMS docs are thorough, and the `make deploy` workflow is clean. There are some issues that need fixing before this is safe for production use.

### Must-Fix (6 issues)
1. **Single Secret for all pods** — every pod (including frontend) gets every secret (DB password, Vault token, LLM keys)
2. **`aurora.fullname` ignores Release.Name** — can't install multiple releases in the same cluster
3. **Vault probes always pass** — `vault status || exit 0` swallows sealed/error states
4. **Redis has no PVC or probes** — data lost on restart, no health monitoring
5. **Weaviate uses Deployment + static PVC** — breaks with replicas > 1
6. **`imagePullSecrets` not wired** — pullSecrets is in values.yaml but never used in templates

### Should-Fix (6 issues)
7. No RBAC (ServiceAccount/Role/RoleBinding)
8. No NetworkPolicies — frontend can reach Vault/Postgres directly
9. No `securityContext` — containers may run as root
10. Frontend and celery-beat have no health probes
11. `namespace.yaml` is redundant with `--create-namespace`
12. `ENABLE_POD_ISOLATION` exposed in config but not implemented in templates

Makefile
```diff
@@ -187,5 +191,56 @@ prod-local-nuke:
...
+# Kubernetes deployment commands
+deploy-build:
+	@echo "Building and pushing images for Kubernetes deployment..."
+	@if [ ! -f deploy/helm/aurora/values.generated.yaml ]; then \
+		echo "Error: values.generated.yaml not found. Copy values.yaml to values.generated.yaml and configure it."; \
+		exit 1; \
+	fi
+	@echo "Extracting image registry and build args from values.generated.yaml..."
+	@set -e; \
+	IMAGE_REGISTRY=$$(grep -A2 '^image:' deploy/helm/aurora/values.generated.yaml | grep 'registry:' | sed 's/.*registry: *"\(.*\)"/\1/'); \
```
Zarlanx: ### Must-Fix: Fragile YAML parsing with grep/sed

This entire block parses `values.generated.yaml` using grep, sed, and cut. This is brittle and will silently break in several realistic scenarios:

**Breakage examples:**
```yaml
# 1. Unquoted value (valid YAML, breaks the sed regex)
image:
  registry: gcr.io/my-project  # sed expects "..."

# 2. Comment on same line
image:
  registry: "gcr.io/my-project"  # my registry
  # sed captures: gcr.io/my-project"  # my registry

# 3. Multi-word or special chars in NEXT_PUBLIC values
NEXT_PUBLIC_BACKEND_URL: "https://api.example.com/v1"  # slashes in sed delimiter

# 4. Indented NEXT_PUBLIC_* key (grep matches wrong line)
  NEXT_PUBLIC_ENABLE_OVH: "false"  # grep '$$var:' may match a comment too
```

Also, line 206 reads var names from `.env.example` but values from `values.generated.yaml` — two different sources with no validation that keys match.

**Suggestion**: Use `yq` (available as a single binary, MIT license) which properly parses YAML:
```makefile
IMAGE_REGISTRY=$$(yq '.image.registry' deploy/helm/aurora/values.generated.yaml); \
for var in $$(yq '.config | keys | .[] | select(test("^NEXT_PUBLIC_"))' deploy/helm/aurora/values.generated.yaml); do \
  value=$$(yq ".config.$$var" deploy/helm/aurora/values.generated.yaml); \
  BUILD_ARGS="$$BUILD_ARGS --build-arg $$var=$$value"; \
done; \
```

If `yq` is too much of a dependency, at minimum add a validation step that checks the extracted values are non-empty before proceeding with the docker builds.

deploy/helm/aurora/templates/configmap.yaml
```diff
@@ -0,0 +1,56 @@
...
+  STORAGE_ENDPOINT_URL: {{ .Values.config.STORAGE_ENDPOINT_URL | quote }}
+  STORAGE_REGION: {{ .Values.config.STORAGE_REGION | quote }}
+  STORAGE_USE_SSL: {{ .Values.config.STORAGE_USE_SSL | quote }}
+  STORAGE_VERIFY_SSL: {{ .Values.config.STORAGE_VERIFY_SSL | quote }}
+  STORAGE_CACHE_ENABLED: {{ .Values.config.STORAGE_CACHE_ENABLED | quote }}
+  STORAGE_CACHE_TTL: {{ .Values.config.STORAGE_CACHE_TTL | quote }}
+  WEAVIATE_HOST: {{ default (printf "%s-weaviate" (include "aurora.fullname" .)) .Values.config.WEAVIATE_HOST | quote }}
+  WEAVIATE_PORT: {{ .Values.config.WEAVIATE_PORT | quote }}
+  WEAVIATE_GRPC_PORT: {{ .Values.config.WEAVIATE_GRPC_PORT | quote }}
+  NEXT_PUBLIC_BACKEND_URL: {{ .Values.config.NEXT_PUBLIC_BACKEND_URL | quote }}
```
Zarlanx: ### Must-Fix: `NEXT_PUBLIC_*` env vars in ConfigMap don't work at runtime for Next.js

Next.js inlines `NEXT_PUBLIC_*` values at **build time** via `process.env.NEXT_PUBLIC_*` → they're replaced with string literals in the JS bundle during `next build`. Setting them as runtime env vars in a ConfigMap has **no effect** on the running frontend container.

This means:
1. Changing `NEXT_PUBLIC_BACKEND_URL` in `values.generated.yaml` and redeploying **will not update the frontend** — you must rebuild the Docker image
2. The `checksum/config` annotation on `frontend-deployment.yaml` will restart the frontend pod when config changes, but the pod will still use the old baked-in values
3. Users will think they can change URLs by updating values and redeploying, but it won't work

**The Makefile's `deploy-build` actually handles this correctly** — it passes `NEXT_PUBLIC_*` as `--build-arg` during `docker build`. But the ConfigMap here creates a false sense of runtime configurability.

**Options:**
1. **Document clearly** that `NEXT_PUBLIC_*` values require image rebuild (`make deploy` does this, but `helm upgrade` alone doesn't)
2. **Use Next.js runtime config** via `publicRuntimeConfig` in `next.config.js` (reads env vars at request time, not build time) — this is a bigger change but makes the ConfigMap actually work
3. **Remove `NEXT_PUBLIC_*` from the ConfigMap** and only set them as build args — at least it's not misleading

Related: `frontend-deployment.yaml` line 16 has `checksum/config` which will restart the pod on ConfigMap changes, but that restart is pointless for `NEXT_PUBLIC_*` vars.

deploy/helm/aurora/templates/ingress.yaml
```diff
@@ -0,0 +1,67 @@
+{{- if .Values.ingress.enabled }}
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: {{ include "aurora.fullname" . }}-ingress
+  labels:
+    {{- include "aurora.labels" . | nindent 4 }}
+  annotations:
+    nginx.ingress.kubernetes.io/proxy-read-timeout: "3600"
```
Zarlanx: ### Should-Fix: Hardcoded nginx-specific annotations

These annotations are nginx-ingress-controller specific. If someone uses Traefik, HAProxy, AWS ALB, or GCE ingress, these annotations are silently ignored and the 3600s timeouts, body size limits, and WebSocket support won't be configured.

**Specific issues:**
- `proxy-read-timeout: 3600` — essential for long-running RCA WebSocket connections (240 agent iterations can take 30+ minutes). Without this on a non-nginx controller, the connection drops at the default 60s timeout.
- `proxy-http-version: 1.1` — required for WebSocket upgrade. Without it, WebSocket connections fail entirely.
- `proxy-body-size: 50m` — needed for file uploads. Without it, large file uploads get 413'd.

**Suggestion**: Make annotations configurable via values:
```yaml
# values.yaml
ingress:
  annotations: {}
  # Default nginx annotations applied when className is "nginx"
  nginx:
    enabled: true  # auto-detect from className?
```

Template:
```yaml
annotations:
  {{- if eq .Values.ingress.className "nginx" }}
  nginx.ingress.kubernetes.io/proxy-read-timeout: "3600"
  nginx.ingress.kubernetes.io/proxy-send-timeout: "3600"
  nginx.ingress.kubernetes.io/proxy-body-size: "50m"
  nginx.ingress.kubernetes.io/proxy-http-version: "1.1"
  {{- end }}
  {{- with .Values.ingress.annotations }}
  {{- toYaml . | nindent 4 }}
  {{- end }}
```

At minimum, add a comment warning that these are nginx-only and need manual adjustment for other controllers.

deploy/helm/aurora/templates/ingress.yaml
```diff
@@ -0,0 +1,67 @@
...
+    nginx.ingress.kubernetes.io/proxy-send-timeout: "3600"
+    nginx.ingress.kubernetes.io/proxy-body-size: "50m"
+    # HTTP/1.1 for WebSocket support and session persistence
+    nginx.ingress.kubernetes.io/proxy-http-version: "1.1"
+    nginx.ingress.kubernetes.io/upstream-hash-by: "$remote_addr"
```
Zarlanx: ### Should-Fix: `upstream-hash-by: $remote_addr` doesn't provide reliable WebSocket sticky sessions

This attempts IP-based session affinity, but it breaks in common scenarios:

1. **Behind a cloud load balancer**: All traffic arrives from the LB's IP, so all users hash to the same upstream pod. One chatbot pod gets 100% of traffic.
2. **Corporate NAT**: Thousands of users behind one public IP all go to the same pod.
3. **Mobile users**: IP changes when switching networks, breaking mid-conversation WebSocket connections.

**For WebSocket connections, you need connection-level stickiness, not request-level.**

Nginx Ingress has proper cookie-based affinity:
```yaml
nginx.ingress.kubernetes.io/affinity: "cookie"
nginx.ingress.kubernetes.io/session-cookie-name: "aurora-ws-affinity"
nginx.ingress.kubernetes.io/session-cookie-max-age: "3600"
```

But note: this only helps for the HTTP upgrade request. Once the WebSocket is established, it's a persistent TCP connection that naturally stays on the same pod. The real issue is ensuring the upgrade request hits the right pod — cookie affinity handles that.

If chatbot replicas > 1, also consider whether Aurora's chatbot is designed for multi-instance (shared state via Redis?) or if it should always be replicas: 1.

deploy/helm/aurora/templates/searxng-deployment.yaml
```diff
@@ -0,0 +1,42 @@
...
+        checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
+        checksum/secret: {{ include (print $.Template.BasePath "/secret.yaml") . | sha256sum }}
+      labels:
+        app.kubernetes.io/name: {{ include "aurora.name" . }}-searxng
+        app.kubernetes.io/instance: {{ .Release.Name }}
+        app.kubernetes.io/part-of: aurora
+    spec:
+      containers:
+        - name: searxng
+          image: searxng/searxng:latest
```
Zarlanx: ### Must-Fix: `latest` tag — non-reproducible deployments

Using `searxng/searxng:latest` means:
1. **Non-reproducible**: Two deploys on different days pull different images. A rollback with `helm rollback` won't actually roll back the SearXNG version.
2. **Silent breakage**: SearXNG publishes breaking changes to `latest` without warning. The search integration could break after a pod restart (K8s pulls `latest` on every reschedule if `imagePullPolicy: Always`, or uses a stale cached image if `IfNotPresent`).
3. **No audit trail**: `kubectl describe pod` shows `latest` — you can't tell which version is running without `docker inspect`.

**Suggestion**: Pin to a specific version tag. SearXNG publishes dated tags:
```yaml
image: searxng/searxng:2024.12.23-a1b2c3d4f  # or whatever version matches docker-compose.yaml
```

Make it configurable:
```yaml
# values.yaml
images:
  searxng:
    repository: searxng/searxng
    tag: "2024.12.23-a1b2c3d4f"
```

Also: this deployment has no health probes. SearXNG exposes `GET /healthz` — add readiness/liveness probes using it.

deploy/helm/aurora/templates/t2v-transformers-deployment.yaml
```diff
@@ -0,0 +1,33 @@
...
+    metadata:
+      annotations:
+        checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
+      labels:
+        app.kubernetes.io/name: {{ include "aurora.name" . }}-t2v-transformers
+        app.kubernetes.io/instance: {{ .Release.Name }}
+        app.kubernetes.io/part-of: aurora
+    spec:
+      containers:
+        - name: transformers
```
Zarlanx, line 22: ### Should-Fix: No health probes for ML model service

The transformers inference server has no readiness or liveness probes. This service loads an ML model into memory on startup which takes 30-60 seconds. During that time:
- K8s considers the pod Ready immediately (no readiness probe to say otherwise)
- Weaviate starts sending vectorization requests
- Requests fail because the model isn't loaded yet
- Weaviate retries or returns errors → document ingestion fails silently

The Weaviate transformers inference container exposes a health endpoint:
```yaml
readinessProbe:
  httpGet:
    path: /.well-known/ready
    port: 8080
  initialDelaySeconds: 30
  periodSeconds: 10
  timeoutSeconds: 5
  failureThreshold: 6
livenessProbe:
  httpGet:
    path: /.well-known/live
    port: 8080
  initialDelaySeconds: 60
  periodSeconds: 30
```

The high `initialDelaySeconds` accounts for model loading time. Without this, the first few minutes after deploy will have broken vectorization.

deploy/helm/aurora/templates/celery-worker-deployment.yaml
```diff
@@ -0,0 +1,40 @@
...
+      annotations:
+        checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
+        checksum/secret: {{ include (print $.Template.BasePath "/secret.yaml") . | sha256sum }}
+      labels:
+        app.kubernetes.io/name: {{ include "aurora.name" . }}-celery-worker
+        app.kubernetes.io/instance: {{ .Release.Name }}
+        app.kubernetes.io/part-of: aurora
+    spec:
+      containers:
+        - name: celery-worker
```
Zarlanx, line 38: ### Should-Fix: Celery worker has no health probes

Celery workers can hang in several ways that don't crash the process:
- Worker connected to Redis but stopped consuming tasks (deadlock)
- Memory leak causing OOM thrashing but process stays alive
- Lost connection to broker but process keeps running with stale state

Without a liveness probe, K8s never restarts the hung worker. Tasks pile up in Redis and nobody processes them — RCA investigations stall silently.

**Suggestion**: Celery has a built-in health check:
```yaml
livenessProbe:
  exec:
    command:
      - /bin/sh
      - -c
      - "celery -A celery_config inspect ping -d celery@$HOSTNAME 2>/dev/null | grep -q pong"
  initialDelaySeconds: 30
  periodSeconds: 60
  timeoutSeconds: 10
  failureThreshold: 3
```

This asks the worker to respond to a ping via the broker. If the worker is hung or disconnected, the ping times out and K8s restarts it.

For readiness, you can use the same command but with a shorter period:
```yaml
readinessProbe:
  exec:
    command:
      - /bin/sh
      - -c
      - "celery -A celery_config inspect ping -d celery@$HOSTNAME 2>/dev/null | grep -q pong"
  initialDelaySeconds: 15
  periodSeconds: 30
  timeoutSeconds: 10
```

deploy/helm/aurora/templates/vault-config.yaml
```diff
@@ -0,0 +1,33 @@
...
+  vault.hcl: |
+    disable_mlock = true
+
+    storage "file" {
+      path = "/vault/data"
+    }
+
+    listener "tcp" {
+      address     = "0.0.0.0:8200"
+      tls_disable = 1
```
Zarlanx, line 18: ### Should-Fix: `tls_disable = 1` — Vault listener accepts plaintext HTTP

Vault's TCP listener has TLS disabled, meaning all communication (including tokens and secrets) travels as plaintext HTTP inside the cluster. This is acceptable only if:
1. You trust the cluster network completely (no multi-tenant, no untrusted workloads)
2. The ingress terminates TLS and internal traffic stays within the same node or trusted network

However, the chart also sets `ENABLE_POD_ISOLATION: "false"` with plans for isolated terminal pods in an `untrusted` namespace. If those untrusted pods can sniff network traffic (no NetworkPolicy, no encryption), they can intercept Vault tokens.

**For production, consider:**
- Enable TLS on the Vault listener with auto-generated certs (Vault can generate its own)
- Or use a service mesh (Istio/Linkerd) that provides mTLS between pods
- Or at minimum, add NetworkPolicies to restrict which pods can reach Vault's port 8200

**For now**: Add a comment in `values.yaml` documenting this tradeoff and that internal traffic is unencrypted.

deploy/helm/aurora/templates/weaviate-statefulset.yaml
```diff
@@ -0,0 +1,64 @@
...
+        - name: weaviate
+          image: cr.weaviate.io/semitechnologies/weaviate:1.27.6
+          imagePullPolicy: IfNotPresent
+          ports:
+            - name: http
+              containerPort: 8080
+            - name: grpc
+              containerPort: 50051
+          env:
+            - name: AUTHENTICATION_ANONYMOUS_ACCESS_ENABLED
```
Zarlanx, line 37: ### Should-Fix: `AUTHENTICATION_ANONYMOUS_ACCESS_ENABLED: true` — Weaviate has no auth

Anyone who can reach Weaviate's port 8080 can read, write, and delete all vector data. In the current chart:
- No NetworkPolicy restricts access to Weaviate
- The Weaviate Service is ClusterIP (not externally exposed), but **any pod in the namespace** (or any namespace without NetworkPolicy) can access it
- If `ENABLE_POD_ISOLATION` is enabled, untrusted terminal pods in the same cluster could query/delete all embeddings

**Mitigations (pick one or more):**
1. **NetworkPolicy**: Only allow traffic from server/celery/chatbot pods to Weaviate
2. **Weaviate API key auth**: Set `AUTHENTICATION_APIKEY_ENABLED: true` and configure an API key
3. **OIDC auth**: Weaviate supports OIDC if you have an identity provider

For the initial release, a NetworkPolicy is the simplest fix:
```yaml
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  name: weaviate-access
spec:
  podSelector:
    matchLabels:
      app.kubernetes.io/name: aurora-oss-weaviate
  ingress:
    - from:
        - podSelector:
            matchLabels:
              app.kubernetes.io/part-of: aurora
      ports:
        - port: 8080
        - port: 50051
```

deploy/helm/aurora/templates/frontend-deployment.yaml
```diff
@@ -0,0 +1,38 @@
...
+spec:
+  replicas: {{ .Values.replicaCounts.frontend }}
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: {{ include "aurora.name" . }}-frontend
+      app.kubernetes.io/instance: {{ .Release.Name }}
+  template:
+    metadata:
+      annotations:
+        checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
```
Zarlanx, line 16: ### Note: `checksum/config` triggers unnecessary restarts for frontend

The frontend only uses `NEXT_PUBLIC_*` values which are baked at build time (see configmap.yaml comment). When you change a backend-only config like `AGENT_RECURSION_LIMIT` or `REDIS_URL`, the ConfigMap checksum changes and the frontend pod restarts — but nothing about the frontend actually changed.

This causes:
- Unnecessary downtime for frontend users
- Wasted pod scheduling and image pull time
- False rollout events in monitoring

**Suggestion**: Either:
1. Create a separate `configmap-frontend.yaml` with only the vars the frontend actually needs, and reference that checksum
2. Remove `checksum/config` from frontend since it gets `NEXT_PUBLIC_*` baked at build time anyway
3. Use `checksum/secret` only (if frontend needs any runtime secrets like `AUTH_SECRET`)

Zarlanx (commented):
## Supplementary Review — Additional Findings

Expanding on the first review with more detailed findings across the Makefile, ingress, ConfigMap, and remaining deployment templates.

### Additional Must-Fix (3)
13. Makefile `deploy-build` uses fragile grep/sed to parse YAML
14. `NEXT_PUBLIC_*` vars in ConfigMap don't work — they're baked at Next.js build time
15. SearXNG uses `latest` tag — non-reproducible deploys

### Additional Should-Fix (7)
16. Ingress hardcodes nginx annotations — breaks other controllers
17. Ingress `upstream-hash-by: $remote_addr` is not real WebSocket sticky sessions
18. Vault listener has `tls_disable = 1` — plaintext HTTP inside the cluster
19. Weaviate has `AUTHENTICATION_ANONYMOUS_ACCESS_ENABLED: true`
20. Celery worker has no health probes
21. t2v-transformers has no health probes
22. No `PodDisruptionBudget` for any service

deploy/helm/aurora/templates/secret.yaml
OlivierTrudeau: Should be good

deploy/helm/aurora/templates/_helpers.tpl
OlivierTrudeau: Done

deploy/helm/aurora/templates/vault-statefulset.yaml
OlivierTrudeau: Done

deploy/helm/aurora/templates/redis-statefulset.yaml
OlivierTrudeau: Made it a stateful set

deploy/helm/aurora/templates/weaviate-deployment.yaml
OlivierTrudeau: keeping it as 1 but explain in a comment that can be scaled through external services and made it easily deactivated if needed. All the databases and statefulsets scaled at 1 instance but they can all support multiple connections. Can be scaled vertically for more concurrent connections or hosted on a third party for hyperscaling

deploy/helm/aurora/values.yaml
OlivierTrudeau: Fixed. Only added to images pulled from built images and not third party public images such as postgres

deploy/helm/aurora/templates/server-deployment.yaml
OlivierTrudeau: None of them orchestrate k8s workloads so wasn't much customization to be done, but now can track by service account. Also made automountserviceaccountoken false since they don't need it

deploy/helm/aurora/templates/server-deployment.yaml
OlivierTrudeau: Fixed and added comment for the PIC_LOCK to be activated depending on the mlock state

deploy/helm/aurora/templates/frontend-deployment.yaml
OlivierTrudeau: done

deploy/helm/aurora/templates/celery-beat-deployment.yaml
OlivierTrudeau: Added the probe check

deploy/helm/aurora/templates/namespace.yaml
OlivierTrudeau: Done

deploy/helm/aurora/templates/frontend-deployment.yaml
OlivierTrudeau: Yeah now only gets 3 secrets from the app-secrets

deploy/helm/aurora/values.yaml
OlivierTrudeau: Now uses the pod isolation and provisions the pods in the isolated namespace. Only thing manual is for the untrusted node which the user needs to provision themselves but it's clarified with comments in the values.yaml

Makefile
OlivierTrudeau: Added yq. Not really a dependency since it's installed on the user's machine.

deploy/helm/aurora/templates/configmap.yaml
OlivierTrudeau: Removed from configmap and added clear comment

deploy/helm/aurora/templates/ingress.yaml
OlivierTrudeau: Added comments and unhardcoded. Did not add native support for other configs tho. Could be a ticket in the future

deploy/helm/aurora/templates/ingress.yaml
OlivierTrudeau: Done through cookies and pod affinity is already true so we can scale to multiple replicas of chatbot instances

deploy/helm/aurora/templates/searxng-deployment.yaml
OlivierTrudeau: Fixed the latest tag and put a specific version. also added the health check

deploy/helm/aurora/templates/t2v-transformers-deployment.yaml
OlivierTrudeau: Added the probes

deploy/helm/aurora/templates/celery-worker-deployment.yaml
OlivierTrudeau: Added probes

deploy/helm/aurora/templates/frontend-deployment.yaml
OlivierTrudeau: it uses the value backend_url from configmap so not completely useless. however, I agree that it will restart on config map changes but realistically adding code to avoid this is such a marginal gain that I don't think it's even worth the effort

deploy/helm/aurora/templates/vault-config.yaml
OlivierTrudeau: Added network policies as well as comments to address the situation

OlivierTrudeau: Actually restricting untrusted namespace egress instead. The rest of the cluster can still sniff but not the untrusted pod

deploy/helm/aurora/templates/weaviate-statefulset.yaml
OlivierTrudeau: Also still accessible but isolated the untrusted namespace properly. Let me know if you don't think this is enough

Zarlanx: It should be enough for an initial deployment, we can revisit this later.

PR #34, opened 2026-01-30
Sms10221/dev 745 memgraph

isiddharthsingh:
## Memgraph Service Discovery Engine                                                                                                                                              
                                                                                                                                                                                    
  Adds a 3-phase infrastructure discovery pipeline backed by Memgraph graph database.                                                                                               
                                                                                                                                                                                    
  ### What's new                                                                                                                                                                    
  - **Memgraph** added to Docker stack (memgraph + memgraph-lab on port 3001)                                                                                                       
  - **6 cloud providers**: GCP (Cloud Asset API), AWS (Resource Explorer 2), Azure (Resource Graph), OVH, Scaleway, Tailscale                                                       
  - **4 enrichment modules**: Kubernetes internals, AWS details, Azure details, serverless env vars                                                                                 
  - **11 inference methods**: GCP relationships, security groups, load balancers, env vars, DNS, event sources, IAM, storage, secrets, CloudMap, network proximity                  
  - **9 REST API endpoints** under `/api/graph` for CRUD, impact analysis, and on-demand discovery                                                                                  
  - **Celery beat** schedules: hourly discovery + daily stale service detection                                                                                                     
                                                                                                                                                                                    
  ### API endpoints                                                                                                                                                                 
  | Method | Endpoint | Purpose |                                                                                                                                                   
  |--------|----------|---------|                                                                                                                                                   
  | GET | `/api/graph` | Full dependency graph |                                                                                                                                    
  | GET | `/api/graph/services` | List services (filterable) |                                                                                                                      
  | GET | `/api/graph/services/<name>` | Service with dependencies |                                                                                                                
  | GET | `/api/graph/services/<name>/impact` | Blast radius analysis |                                                                                                             
  | POST | `/api/graph/services` | Add/update service |                                                                                                                             
  | POST | `/api/graph/dependencies` | Add dependency |                                                                                                                             
  | DELETE | `/api/graph/dependencies/<id>` | Remove dependency |                                                                                                                   
  | POST | `/api/graph/discover` | Trigger discovery (async) |                                                                                                                      
  | GET | `/api/graph/stats` | Graph statistics |                                                                                                                                   
                                                                                                                                                                                    
  ### Testing                                                                                                                                                                       
  All verified in Docker:                                                                                                                                                           
  - 28/28 module imports pass                                                                                                                                                       
  - Memgraph CRUD lifecycle works (upsert, list, get, deps, impact, export, delete)                                                                                                 
  - Network proximity inference correctly detects vm→db and vm→cache edges                                                                                                          
  - 9 API routes registered in Flask app

### DIFF
--- a/server/routes/graph_routes.py
+++ b/server/routes/graph_routes.py
@@ -161,8 +161,18 @@ def trigger_discovery(user_id):
     from services.discovery.tasks import run_user_discovery
     from utils.cache.redis_client import get_redis_client
 
-    # Deduplicate: if a discovery task is already running for this user, return its ID
     redis_client = get_redis_client()
+
+    # Rate limit: 1 request per 30 seconds per user
+    rate_key = f"discovery:rate:{user_id}"
+    if redis_client and redis_client.get(rate_key):
+        return jsonify({
+            "error": "Rate limited. Please wait 30 seconds between discovery requests.",
+        }), 429
+    if redis_client:
+        redis_client.setex(rate_key, 30, "1")
+
+    # Deduplicate: if a discovery task is already running for this user, return its ID
     lock_key = f"discovery:running:{user_id}"
     if redis_client:
         existing_task_id = redis_client.get(lock_key)


docker-compose.yaml
```diff
@@ -522,6 +531,43 @@ services:
...
+    image: memgraph/memgraph-mage:latest
+    container_name: aurora-memgraph
+    ports:
+      - "7687:7687"
+      - "7444:7444"
+    volumes:
+      - memgraph-data:/var/lib/memgraph
+    environment:
+      - MEMGRAPH_USER=${MEMGRAPH_USER:-aurora}
+      - MEMGRAPH_PASSWORD=${MEMGRAPH_PASSWORD:-aurora_secure_password}
```
damianloch: remove these defaults. As long as they are in .env.example they should populate for users

docker-compose.yaml
```diff
@@ -522,6 +531,43 @@ services:
...
+      interval: 10s
+      timeout: 5s
+      retries: 5
+    restart: unless-stopped
+
+  memgraph-lab:
+    image: memgraph/lab:latest
+    container_name: aurora-memgraph-lab
+    ports:
+      - "3001:3000"
```
damianloch, line 563: 3000 is the frontend so we know it will never work there

docker-compose.yaml
```diff
@@ -86,6 +86,13 @@ x-common-env: &common-env
   VAULT_TOKEN: ${VAULT_TOKEN}
   VAULT_KV_MOUNT: ${VAULT_KV_MOUNT}
 
+  # Memgraph (Graph Database)
+  MEMGRAPH_HOST: ${MEMGRAPH_HOST:-memgraph}
+  MEMGRAPH_PORT: ${MEMGRAPH_PORT:-7687}
+  MEMGRAPH_USER: ${MEMGRAPH_USER:-aurora}
+  MEMGRAPH_PASSWORD: ${MEMGRAPH_PASSWORD:-aurora_secure_password}
+  DISCOVERY_INTERVAL_HOURS: ${DISCOVERY_INTERVAL_HOURS:-1}
```
damianloch: remove these defaults. As long as they are in .env.example they should populate for users

docker-compose.prod-local.yml
damianloch, line 1: view commments for docker-compose.yaml. Same ones apply

server/celery_config.py
```diff
@@ -70,6 +71,14 @@
             'task': 'knowledge_base.cleanup_stale_documents',
             'schedule': 180.0,  # Every 3 minutes
         },
+        'run-full-discovery': {
+            'task': 'services.discovery.tasks.run_full_discovery',
+            'schedule': float(os.getenv('DISCOVERY_INTERVAL_HOURS', '1')) * 3600,  # Default: every hour
```
damianloch, line 77: This should just be a constant not an env variable in my opinion. If you feel strongly about it then no default it needed since its already in .env.example but raise an error if that var is not detected so its traceable

server/services/graph/memgraph_client.py
```diff
@@ -0,0 +1,717 @@
...
+    """Singleton accessor for the Memgraph client."""
+    global _client_instance
+    if _client_instance is None:
+        with _client_lock:
+            if _client_instance is None:
+                _client_instance = MemgraphClient(
+                    host=os.getenv("MEMGRAPH_HOST", "memgraph"),
+                    port=int(os.getenv("MEMGRAPH_PORT", "7687")),
+                    username=os.getenv("MEMGRAPH_USER", "aurora"),
+                    password=os.getenv("MEMGRAPH_PASSWORD", "aurora_secure_password"),
```
damianloch: these are defined in .env.example so raise value error if not found instead of fallbacks

server/services/graph/memgraph_client.py
```diff
@@ -0,0 +1,717 @@
+"""
+Memgraph Client - Sole interface between Aurora and Memgraph graph database.
+All Cypher queries are encapsulated here. Uses gqlalchemy driver.
```
damianloch: Not sure if gqlalchemy driver is used anymore

server/services/graph/memgraph_client.py
```diff
@@ -0,0 +1,717 @@
...
+            uri = f"bolt://{self._host}:{self._port}"
+            self._driver = GraphDatabase.driver(
+                uri,
+                auth=(self._username, self._password),
+            )
+            logger.info(f"Connected to Memgraph at {uri}")
+        if not self._schema_initialized:
+            self._schema_initialized = True  # Set BEFORE calling to prevent recursion
+            self.ensure_schema()
+        return self._driver
```
damianloch, line 58: Driver connection is never closed. In long running processes this can leak connections

server/services/graph/memgraph_client.py
```diff
@@ -0,0 +1,717 @@
...
+            "team_owner": props.get("team_owner", ""),
+            "metadata": json.dumps(props.get("metadata", {})) if isinstance(props.get("metadata"), dict) else props.get("metadata", "{}"),
+        }
+        results = self._execute(query, params)
+        return self._node_to_dict(results[0]["s"]) if results else None
+
+    def batch_upsert_services(self, user_id, services):
+        """Upsert multiple services in a single transaction. Returns count."""
+        count = 0
+        for svc in services:
```
Zarlanx, line 174: N+1 query pattern — executes one Cypher round-trip per service. 200 discovered services = 200 network round-trips. Use `UNWIND $services AS svc` with a single MERGE query to batch the entire list into one call.

server/services/graph/memgraph_client.py
```diff
@@ -0,0 +1,717 @@
...
+            "confidence": confidence,
+            "discovered_from": discovered_list,
+        }
+        results = self._execute(query, params)
+        return results[0] if results else None
+
+    def batch_upsert_dependencies(self, user_id, deps):
+        """Upsert multiple edges. Returns count."""
+        count = 0
+        for dep in deps:
```
Zarlanx, line 352: Same N+1 pattern as `batch_upsert_services` above. Each dependency fires a separate round-trip including a `_resolve_service_id` lookup. Batch with `UNWIND`.

server/services/graph/memgraph_client.py
```diff
@@ -0,0 +1,717 @@
...
+        self._port = port
+        self._username = username
+        self._password = password
+        self._driver = None
+        self._schema_initialized = False
+
+    def _get_driver(self):
+        if self._driver is None:
+            uri = f"bolt://{self._host}:{self._port}"
+            self._driver = GraphDatabase.driver(
```
Zarlanx, line 49: Driver is created without pool configuration (defaults may be too small for concurrent Celery workers) and the class has no `close()` method, connections will leak on shutdown. Add `max_connection_pool_size` and a shutdown hook.

server/routes/graph_routes.py
```diff
@@ -0,0 +1,299 @@
...
+        return err, code
+    try:
+        from services.discovery.tasks import run_user_discovery
+        from utils.cache.redis_client import get_redis_client
+
+        # Deduplicate: if a discovery task is already running for this user, return its ID
+        redis_client = get_redis_client()
+        lock_key = f"discovery:running:{user_id}"
+        if redis_client:
+            existing_task_id = redis_client.get(lock_key)
```
Zarlanx: TOCTOU race: the gap between this `GET` and the `SETEX` on line 215 means two concurrent requests can both pass the check and dispatch duplicate tasks. Use `SET NX EX` for atomic lock acquisition instead of separate GET + SETEX.

server/services/discovery/tasks.py
```diff
@@ -0,0 +1,298 @@
...
+    except Exception as e:
+        logger.error(f"[Discovery Task] Fatal error: {e}")
+        return {"status": "error", "error": str(e)}
+
+
+@celery_app.task(
+    name="services.discovery.tasks.run_user_discovery",
+    bind=True,
+    max_retries=0,
+    soft_time_limit=7200,
```
Zarlanx, line 200: `soft_time_limit=7200` (2h) / `time_limit=10800` (3h) is very generous. More importantly, there's no `except SoftTimeLimitExceeded` handler in the task body — when the soft limit fires, the task crashes without cleanup (Redis lock won't be cleared, partial state left behind).

server/services/discovery/tasks.py
```diff
@@ -0,0 +1,298 @@
...
+        _clear_discovery_lock(user_id)
+        return summary
+
+    except Exception as e:
+        logger.error(f"[Discovery Task] Failed for user {user_id}: {e}")
+        _clear_discovery_lock(user_id)
+        return {"status": "error", "user_id": user_id, "error": str(e)}
+
+
+@celery_app.task(name="services.discovery.tasks.mark_stale_services", bind=True, max_retries=0)
```
Zarlanx: No `time_limit` or `soft_time_limit` at all. If the Memgraph query hangs (e.g. large graph scan), this task blocks the worker indefinitely.

server/services/discovery/tasks.py
```diff
@@ -0,0 +1,298 @@
...
+
+        # Group by user — providers fetch their own credentials at runtime
+        users = {}
+        for user_id, provider in rows:
+            users.setdefault(user_id, {})[provider] = {}
+
+        logger.info(f"[Discovery Task] Processing {len(users)} users")
+
+        results = []
+        for user_id, providers in users.items():
```
Zarlanx, line 176: Processes all users sequentially in a single task. User #50 waits for users 1-49 to finish. Fan out to per-user subtasks via `run_user_discovery.delay(user_id)` and collect results.

Zarlanx (changes requested):
7 must-fix issues below.

Positive: all Cypher queries use parameterized `$param` syntax (no injection), every query filters by `user_id` (correct data isolation), singleton with double-checked locking is thread-safe, and the 3-phase pipeline design is clean.

server/services/graph/memgraph_client.py
```diff
@@ -0,0 +1,717 @@
...
+    # Schema Initialization
+    # =========================================================================
+
+    def ensure_schema(self):
+        """Create constraints and indexes if they don't exist."""
+        statements = [
+            "CREATE CONSTRAINT ON (s:Service) ASSERT s.id IS UNIQUE;",
+            "CREATE CONSTRAINT ON (i:Incident) ASSERT i.id IS UNIQUE;",
+            "CREATE CONSTRAINT ON (c:Change) ASSERT c.id IS UNIQUE;",
+            "CREATE INDEX ON :Service(user_id);",
```
Zarlanx, line 99: Individual indexes on `user_id` and `name` exist, but the hot query path (`_resolve_service_id`, `get_service`, `list_services`) always filters on both. Add a composite index on `(user_id, name)` for those lookups.

server/services/discovery/tasks.py
```diff
@@ -0,0 +1,298 @@
...
+
+
+def _clear_discovery_lock(user_id):
+    """Remove the Redis dedup lock after a discovery task finishes."""
+    try:
+        from utils.cache.redis_client import get_redis_client
+        redis_client = get_redis_client()
+        if redis_client:
+            redis_client.delete(f"discovery:running:{user_id}")
+    except Exception:
```
Zarlanx: Silently swallowing all exceptions here means you'll never know if lock cleanup is failing. At minimum `logger.debug` the error, otherwise stale locks can pile up with no visibility.

server/celery_config.py
```diff
@@ -70,6 +71,14 @@
             'task': 'knowledge_base.cleanup_stale_documents',
             'schedule': 180.0,  # Every 3 minutes
         },
+        'run-full-discovery': {
```
Zarlanx, line 75: Discovery tasks are registered on the default queue, so they compete with RCA, chatbot, and knowledge-base tasks for worker capacity. Consider routing discovery to a dedicated queue.

docker-compose.yaml
```diff
@@ -522,6 +531,43 @@ services:
       seaweedfs-filer:
         condition: service_healthy
 
+  # Memgraph - In-memory Graph Database
+  # ═══════════════════════════════════════════════════════════════════════════
+
+  memgraph:
+    image: memgraph/memgraph-mage:latest
```
Zarlanx, line 538: Unpinned `latest` tag — non-reproducible builds. Pin to a specific version (e.g. `memgraph/memgraph-mage:2.x.y`).

docker-compose.yaml
```diff
@@ -522,6 +531,43 @@ services:
...
+    container_name: aurora-memgraph
+    ports:
+      - "7687:7687"
```
Zarlanx, line 541: Port 7687 exposed to the host with default credentials (`aurora` / `aurora_secure_password`). Fine for dev, but add a comment or guard so this doesn't ship to staging/prod.

docker-compose.yaml
```diff
@@ -522,6 +531,43 @@ services:
...
+      --log-level=WARNING
+      --also-log-to-stderr
+    healthcheck:
+      test: ["CMD-SHELL", "python3 -c \"import socket; s=socket.socket(); s.settimeout(2); s.connect(('localhost',7687)); s.close()\""]
+      interval: 10s
+      timeout: 5s
+      retries: 5
+    restart: unless-stopped
+
+  memgraph-lab:
```
Zarlanx, line 559: memgraph-lab UI exposed on port 3001 with no authentication. Same concern as above, should be dev-only or behind auth.

.env.example
```diff
@@ -167,6 +167,15 @@ WEAVIATE_HOST=weaviate
...
 WEAVIATE_GRPC_PORT=50051
 
+# -----------------------------------------------------------------------------
+# Memgraph (Graph Database)
+# -----------------------------------------------------------------------------
+MEMGRAPH_HOST=memgraph
+MEMGRAPH_PORT=7687
+MEMGRAPH_USER=aurora
+MEMGRAPH_PASSWORD=aurora_secure_password
+DISCOVERY_INTERVAL_HOURS=1
```
Zarlanx, line 177: `aurora_secure_password` looks real enough that someone will copy-paste it into production `.env` unchanged. Use an obviously placeholder value like `CHANGE_ME` or `<your-password-here>`.

server/routes/graph_routes.py
```diff
@@ -0,0 +1,299 @@
...
+    except Exception as e:
+        logger.error(f"Error deleting dependency: {e}")
+        return jsonify({"error": "Failed to delete dependency"}), 500
+
+
+# =========================================================================
+# Discovery
+# =========================================================================
+
+@graph_bp.route("/discover", methods=["POST"])
```
Zarlanx, line 157: No rate limiting on this endpoint. A misbehaving frontend or API consumer can spam discovery runs. The Redis dedup helps, but only if the previous task is still running — rapid retries after completion would re-trigger immediately.

client/src/hooks/use-graph-discovery-status.ts
```diff
@@ -0,0 +1,166 @@
...
+    const uid = userId;
+    const onProviderChange = () => checkTrigger(uid);
+    window.addEventListener("providerStateChanged", onProviderChange);
+
+    return () => {
+      stopPolling();
+      if (syncedTimerRef.current) clearTimeout(syncedTimerRef.current);
+      window.removeEventListener("providerStateChanged", onProviderChange);
+    };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
Zarlanx, line 178: `eslint-disable` hides that `checkTrigger` and `stopPolling` are missing from the deps array. If their identities change across renders this effect won't re-run. Either add them to deps or document why it's safe to omit.

Zarlanx (commented):
9 should-fix items — supplementary to the must-fix review above.

damianloch (changes requested):

server/services/graph/memgraph_client.py
isiddharthsingh: rewrote both to use UNWIND with a single cypher round-trip. also added a fallback to individual inserts in case the batch query fails for any reason.

server/services/graph/memgraph_client.py
isiddharthsingh: rewrote both to use UNWIND with a single cypher round-trip. also added a fallback to individual inserts in case the batch query fails for any reason.

server/services/graph/memgraph_client.py
isiddharthsingh: fixed, added max_connection_pool_size=50 and a close() method on the client.

server/routes/graph_routes.py
isiddharthsingh: switched to SET NX EX for atomic lock acquisition.

server/services/discovery/tasks.py
isiddharthsingh: the existing except Exception actually does catch it since SoftTimeLimitExceeded inherits from Exception, so the lock was being cleared. but you're right it should be explicit.  added a dedicated except block for clarity and a cleaner error message.

server/services/discovery/tasks.py
isiddharthsingh: added soft_time_limit=300 and time_limit=600

server/services/graph/memgraph_client.py
isiddharthsingh: added it to the schema init.

server/services/discovery/tasks.py
isiddharthsingh: changed to logger.debug.

.env.example
isiddharthsingh: changed to CHANGE_ME

server/routes/graph_routes.py
isiddharthsingh: the redis dedup already blocks concurrent runs and the endpoint requires auth. the frontend only fires on provider connection events not user clicks, so spamming isn't a realistic scenario right now.

client/src/hooks/use-graph-discovery-status.ts
isiddharthsingh: the callbacks are all useCallback-wrapped with stable deps so their identity never changes after mount.

server/services/discovery/tasks.py
isiddharthsingh: do you mean adding a dedicated queue for this?

docker-compose.yaml
isiddharthsingh: this i will remove

docker-compose.yaml
isiddharthsingh: this already has a password protection thing. i dont think it will be required. it is similar to postgres db.

server/services/graph/memgraph_client.py
isiddharthsingh: lready fixed — added a close() method and explicit pool size config.

server/services/graph/memgraph_client.py
isiddharthsingh: oh yes, i had to change from gqalchemy because i was having a lot of dependency issues and cmake problems. and neo4juses the same bolt mechanism so i replaced it with that.

server/services/graph/memgraph_client.py
isiddharthsingh: dpne

server/celery_config.py
isiddharthsingh: i feel the user should be in control of it and also lets us tune it without much code changes

docker-compose.yaml
isiddharthsingh: done

docker-compose.prod-local.yml
isiddharthsingh: done

docker-compose.yaml
isiddharthsingh: the port mapping is 3001:3000. the right side (3000) is just memgraph-lab's internal default port that's hardcoded in their image, we can't change that. the left side (3001) is the host port, which is what actually matters. so on localhost, the frontend is on 3000 and memgraph-lab is on 3001, no conflict.

docker-compose.yaml
isiddharthsingh: done

server/services/discovery/tasks.py
Zarlanx: Yes, so basically try to process these in a parallel manner, by having a queue the result collection will be faster.

docker-compose.yaml
Zarlanx: Was written before taking a look at the .env, you're good.

server/celery_config.py
isiddharthsingh: It would not be necessary i feel. Having a dedicated queue would mean one more container and can increase complexity.

docker-compose.yaml
isiddharthsingh: Following the pattern like other images

server/services/discovery/tasks.py
isiddharthsingh: Can you elaborate more on this part? All the graph syncs happen in parallel already and not sure if it would block any other task.

server/routes/graph_routes.py
Zarlanx: Redis dedup can be bypassed at initial trigger, if a malicious agent tries to abuse the api, that could happen.

server/celery_config.py
Zarlanx: If it's going to get too much time, you can discard my comment, but I do think that long term it might be a better solution as we scale.

server/services/discovery/tasks.py
Zarlanx: The current code runs all users sequentially inside a single Celery task. Even if each user’s internal graph sync does parallel work, only one user is active at a time in this outer loop. So user #50 waits for users 1–49 to finish before any work starts for them.  

Fan out would help as if you enqueue run_user_discovery.delay(user_id) per user, Celery can run multiple users in parallel across workers, which reduces total wall‑clock time and tail latency. It also isolates failures/retries per user instead of failing the whole batch. Implementing it using a dedicated queue is optional but useful. It lets you limit concurrency and prevent discovery jobs from starving other background tasks.

server/celery_config.py
isiddharthsingh: okay lets do this later

server/services/discovery/tasks.py
isiddharthsingh: okay as discussed lets implement this later and also mark it like it is required later

server/routes/graph_routes.py
isiddharthsingh: based on discussion, added rate limit 1 request 30 sec

PR #45, opened 2026-02-02
feat(api): expose correlation data in incidents API and add alerts endpoint

Zarlanx:
Adds proper correlation classification once an alert comes through

### DIFF
--- a/server/routes/datadog/tasks.py
+++ b/server/routes/datadog/tasks.py
@@ -302,8 +302,8 @@ def process_datadog_event(
                             """INSERT INTO incident_alerts
                                (user_id, incident_id, source_type, source_alert_id, alert_title, alert_service,
                                 alert_severity, correlation_strategy, correlation_score,
-                                correlation_details, alert_metadata)
-                               VALUES (%s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s)""",
+                                correlation_details, alert_metadata, received_at)
+                               VALUES (%s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s)""",
                             (
                                 user_id,
                                 incident_id,
@@ -316,6 +316,7 @@ def process_datadog_event(
                                 correlation_result.score,
                                 json.dumps(correlation_result.details),
                                 json.dumps(alert_metadata),
+                                received_at,
                             ),
                         )
                         cursor.execute(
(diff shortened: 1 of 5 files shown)


server/routes/incidents_routes.py
```diff
@@ -181,126 +264,214 @@ def get_incident(incident_id: str):
...
+                logger.debug(
+                    "[INCIDENTS] Incident %s: rawPayload length=%d, sourceUrl=%s",
+                    incident_id,
+                    len(incident["alert"]["rawPayload"]),
+                    incident["alert"].get("sourceUrl", ""),
+                )
+
+                cursor.execute(
+                    """SELECT id, source_type, alert_title, alert_service, alert_severity,
+                              correlation_strategy, correlation_score, correlation_details, received_at
```
damianloch, line 449: Might be mistaken but the insert doesn't insert a received_at column

server/routes/incidents_routes.py
```diff
@@ -181,126 +264,214 @@ def get_incident(incident_id: str):
...
+                       FROM incident_alerts
+                       WHERE incident_id = %s
+                       ORDER BY received_at ASC""",
+                    (incident_id,),
+                )
```
damianloch, line 454: Doesn't filter by user id, which doesn't seem to be a vulnerability yet but if the incident ownership check which happens first ever gets removed/modified this will be vulnerability. I would add RLS for safety

server/routes/incidents_routes.py
```diff
@@ -401,72 +582,161 @@ def get_incident(incident_id: str):
...
+                alert_rows = cursor.fetchall()
```
damianloch, line 660: No index on incident_alerts means full table scans here. Fine for now but will get slow at scale.

server/routes/incidents_routes.py
Zarlanx: Fixed. We now explicitly pass `received_at` on all `incident_alerts` inserts (both correlated + primary). Schema still keeps the DEFAULT for safety, but inserts now use the alert’s timestamp.

server/routes/incidents_routes.py
Zarlanx: Added safety at both query + DB level:
- Queries now filter `incident_alerts` by `user_id`.
- `incident_alerts` now has `user_id` + RLS policies (select/insert/update/delete) and a backfill from `incidents`, with NOT NULL enforced.

server/routes/incidents_routes.py
Zarlanx: Added index `idx_incident_alerts_incident_received` on `(incident_id, received_at)` to cover the ORDER BY query.

PR #47, opened 2026-02-03
Bgervais/dev 754 mid rca update

beng360:
This PR adds the ability to provide up to date context to an ongoing RCA investigation from a correlated alert (this will be implemented further). If an incoming alert is marked as "correlated" while its associated RCA is still running, the agent will make a tool call that will fetch context and inject it into the ongoing RCA investigation. As the framework for how alerts will be marked as "correlated" is still pending, a testing document has been provided and is outlined below.

# Testing RCA Context Updates

Verifies that correlated alerts inject context into an ongoing RCA and appear chronologically in the chat UI.

## Setup

Use a slow RCA model in `server/chat/backend/agent/llm.py`:
```python
RCA_MODEL = "google/gemini-2.5-pro-preview"  # ~3-5 min
```

Restart: `docker compose restart celery_worker`

### 1. Trigger initial alert
```bash
curl -s -X POST "http://localhost:5080/pagerduty/webhook/<USER_ID>" \
  -H "Content-Type: application/json" \
  -d '{"event":{"id":"test-001","event_type":"incident.triggered","resource_type":"incident","data":{"id":"INC-001","type":"incident","number":1001,"status":"triggered","title":"Test incident","service":{"id":"SVC001","summary":"test-service"},"urgency":"high"}}}'
```

### 2. Get session ID (wait ~5s)
```bash
docker exec -i aurora-postgres psql -U aurora -d aurora_db -t -c \
  "SELECT id, status FROM chat_sessions ORDER BY created_at DESC LIMIT 1;"
```

### 3. Send correlated alert (wait 15-30s after step 1)
```bash
curl -s -X POST "http://localhost:5080/pagerduty/webhook/<USER_ID>" \
  -H "Content-Type: application/json" \
  -d '{"event":{"id":"corr-001","event_type":"incident.triggered","resource_type":"incident","correlation":{"rca_session_id":"<SESSION_ID>","id":"corr-001"},"data":{"id":"INC-CORR","type":"incident","number":1002,"status":"triggered","title":"Related issue","service":{"id":"SVC002","summary":"service"},"urgency":"high","body":{"details":"Context to inject into RCA"}}}}'
```

## Key Fields

| Field | Description |
|-------|-------------|
| `correlation.rca_session_id` | Target RCA session UUID |
| `data.body.details` | Context message to inject |

## Expected Result

`rca_context_update` appears mid-conversation, not at start/end:
```
1: user - text
2: bot - cloud_exec
3: bot - rca_context_update  <-- here
4: bot - web_search
...
```

Check logs: `docker logs aurora-celery_worker-1 2>&1 | grep "RCA-UPDATE"`

### DIFF
--- a/server/chat/background/context_updates.py
+++ b/server/chat/background/context_updates.py
@@ -6,7 +6,7 @@ import uuid
 from datetime import datetime, timezone
 from typing import Any, Dict, List, Optional
 
-from langchain_core.messages import HumanMessage, SystemMessage
+from langchain_core.messages import HumanMessage
 from utils.cloud.cloud_utils import get_workflow_context
 from utils.cache.redis_client import get_redis_client
 


server/chat/backend/agent/middleware/context_trim.py
```diff
@@ -53,6 +55,20 @@ async def awrap_model_call(
...
+        # Strategy: Append at the END as the most recent message (highest priority).
+        # Only inject ONCE when the update first arrives.
+        state = get_state_context()
+        update_message = apply_rca_context_updates(state)
+        if update_message:
+            logger.info(f"[ContextTrimMiddleware] Got update_message, appending to END of request")
+            try:
+                # Append at end - most recent message has highest priority
+                request = request.override(messages=[*request.messages, update_message])
+                logger.info(f"[ContextTrimMiddleware] ✅ Appended context update to END (now {len(request.messages)} messages)")
```
damianloch: Emoji

server/chat/background/context_updates.py
```diff
@@ -0,0 +1,321 @@
+"""RCA context update queue for mid-run background investigations."""
+
+import json
+import logging
+import uuid
+from datetime import datetime, timezone
+from typing import Any, Dict, List, Optional
+
+from langchain_core.messages import HumanMessage, SystemMessage
```
damianloch: HumanMessage unused

server/chat/background/context_updates.py
```diff
@@ -0,0 +1,321 @@
...
+from utils.cloud.cloud_utils import get_workflow_context
+from utils.cache.redis_client import get_redis_client
+
+logger = logging.getLogger(__name__)
+
+_RCA_UPDATE_KEY_PREFIX = "rca_context_updates"
+_RCA_UPDATE_TTL_SECONDS = 6 * 60 * 60  # 6 hours
+_MAX_UPDATE_CHARS = 8000
```
damianloch: Unused

server/chat/background/context_updates.py
```diff
@@ -0,0 +1,321 @@
...
+    }
+
+    # Check if session is already completed
+    session_status = _get_session_status(session_id)
+    if session_status in ("completed", "failed"):
+        logger.info(
+            "[RCA-UPDATE] Session %s is %s, appending context update directly to database",
+            session_id, session_status
+        )
+        return _append_context_update_to_completed_session(user_id, session_id, update_payload)
```
damianloch, line 174: Context update gets enqueued to Redis for a completed session, then never processed. The update is permanently lost.

server/chat/background/context_updates.py
```diff
@@ -0,0 +1,321 @@
...
+
+def _get_session_status(session_id: str) -> Optional[str]:
+    """Get the current status of a chat session."""
+    try:
+        from utils.db.connection_pool import db_pool
+        with db_pool.get_admin_connection() as conn:
+            with conn.cursor() as cursor:
+                cursor.execute(
+                    "SELECT status FROM chat_sessions WHERE id = %s",
+                    (session_id,)
```
damianloch: get_admin connectoin bypasses RLS for this and we do not filter by userid. We should dd user_id to status check and add RLS

server/chat/background/context_updates.py
```diff
@@ -0,0 +1,321 @@
...
+            event_data = payload.get("event", {}).get("data", {})
+            title = event_data.get("title", "")
+            if title:
+                titles.append(title)
+            body_details = event_data.get("body", {}).get("details", "")
+            if body_details:
+                details.append(body_details)
+    
+    title_str = titles[0] if titles else "Correlated incident"
+    details_str = details[0] if details else ""
```
damianloch, line 241: Won't context be lost by only taking the first value?

server/chat/background/context_updates.py
```diff
@@ -0,0 +1,321 @@
...
+    logger.info(f"[RCA-UPDATE] Created HumanMessage with {len(content)} chars for session {session_id}")
+
+    try:
+        if hasattr(state, "messages") and isinstance(state.messages, list):
+            state.messages.append(update_message)
+            logger.info(f"[RCA-UPDATE] ✅ INJECTED HumanMessage into state.messages (now {len(state.messages)} messages) for session {session_id}")
+            
+            # Store UI update payload for injection during UI conversion
+            tool_call_id = f"rca_context_update_{uuid.uuid4().hex}"
+            injected_at = updates[0].get("received_at") if updates else None
```
damianloch, line 326: if updates is always True at this point (we return None earlier if empty), but updates[0].get("received_at") can return None if the key is missing.

server/chat/background/context_updates.py
```diff
@@ -0,0 +1,321 @@
...
+        "Please incorporate this information into your investigation. Continue your analysis and include this context in your final report.",
+    ])
+
+    return "\n".join(parts)
+
+
+def apply_rca_context_updates(state: Any) -> Optional[HumanMessage]:
+    """Inject queued updates into the in-flight RCA state as a HumanMessage.
+    
+    This function is called on EVERY LLM call. We check Redis for new updates
```
damianloch, line 266: Every LLM call a bit greedy

server/chat/background/context_updates.py
```diff
@@ -0,0 +1,321 @@
...
+                # Renumber all messages
+                for idx, msg in enumerate(messages):
+                    msg["message_number"] = idx + 1
+                
+                # Update the database
+                cursor.execute(
+                    "UPDATE chat_sessions SET messages = %s, updated_at = %s WHERE id = %s AND user_id = %s",
+                    (json.dumps(messages), datetime.now(), session_id, user_id)
+                )
+                conn.commit()
```
damianloch, line 127: DB update without row lock that can cause read-modify-write race conditions

server/chat/background/context_updates.py
damianloch, line 1: If a correlated alert arrives while the RCA is in-progress (enqueued to Redis at line 174), then the session completes, and PagerDuty retries the same webhook, the alert gets written to the database again (line 164). Consider adding deduplication using incident_id + event_id to prevent the same alert from appearing twice in the UI. Not sure how common or important this is but it was flagged. Redis and don't communicate which causes this

server/routes/pagerduty/tasks.py
```diff
@@ -522,8 +561,49 @@ def process_pagerduty_event(
                             alert_metadata=alert_metadata,
                         )
                         
+                        # If correlated, enqueue a context update instead of starting a new RCA
+                        if correlated_session_id or correlated_incident_id:
+                            try:
+                                from chat.background.context_updates import enqueue_rca_context_update
+
+                                session_id_to_update = correlated_session_id
```
damianloch, line 569: No explicit check that this belongs to same user as webhook. Coudl potentially be dangerous

server/chat/backend/agent/middleware/context_trim.py
beng360: removed

server/chat/background/context_updates.py
beng360: fixed in next PR

server/chat/background/context_updates.py
beng360: I dont know about this, like yeah you're probably right it would be better to enqueue

server/routes/pagerduty/tasks.py
beng360: you're absolutely right, but this is depreciated function in next PR. no longer there

server/chat/background/context_updates.py
beng360: basically for testing this was just used for getting the title of the alert, in my new PR it gets pretty much all the thoughts using the pipeline olivier built for "chat with RCA"

server/chat/background/context_updates.py
beng360: fixed

server/chat/background/context_updates.py
beng360: good catch

server/chat/background/context_updates.py
beng360: removed

server/chat/background/context_updates.py
beng360: System Message unused*, removed system message

server/chat/background/context_updates.py
beng360: true, fixed in next pR

server/chat/background/context_updates.py
beng360: fixed

damianloch:
Included in https://github.com/Arvo-AI/aurora/pull/48

PR #64, opened 2026-02-05
Add SSE keepalive heartbeat and stop incident polling on terminal state

isiddharthsingh:
## Summary
  - Add SSE keepalive heartbeat to prevent `/api/incidents/stream` timeout errors
  - Stop 1-second incident polling once RCA investigation completes

  ## Problem
  - The SSE endpoint blocks indefinitely without sending any data, causing the Next.js proxy to timeout and log `TimeoutError` (code 23) in a reconnect loop
  - Incident detail page polls `GET /api/incidents/{id}` every second and never stops, even after the incident is resolved
  - 
image

  ## Changes
  - `server/routes/incidents_sse.py` — Send `: keepalive` SSE comment every 15s to keep connection alive
  - `client/src/app/incidents/[id]/page.tsx` — Stop polling when incident status is no longer `investigating`

  ## Test plan
  - [ ] Open incidents list page, verify no `TimeoutError` in console
  - [ ] Trigger an incident update, verify SSE event still arrives
  - [ ] Open an incident detail page during active RCA, verify thoughts stream in
  - [ ] Wait for RCA to finish, verify polling stops

### DIFF
--- a/server/routes/incidents_sse.py
+++ b/server/routes/incidents_sse.py
@@ -48,7 +48,7 @@ def incident_stream():
             # Continuously read messages and yield as SSE events
             while True:
                 try:
-                    message = message_queue.get(timeout=15)
+                    message = message_queue.get(timeout=10)
                     yield f"data: {json.dumps(message)}\n\n"
                 except queue.Empty:
                     # Send keepalive comment to prevent proxy/client timeouts


server/routes/incidents_sse.py
```diff
@@ -47,8 +47,12 @@ def generate_sse_events():
         try:
             # Continuously read messages and yield as SSE events
             while True:
-                message = message_queue.get()
-                yield f"data: {json.dumps(message)}\n\n"
+                try:
+                    message = message_queue.get(timeout=15)
```
beng360: use 10 second timeout for more agressive CDNs

beng360 (dismissed):

server/routes/incidents_sse.py
isiddharthsingh: okay

PR #146, opened 2026-03-09
feat: add Traefik reverse proxy and security hardening

Zarlanx:
## Summary
- Adds Traefik v3.3 as a TLS-terminating reverse proxy for `docker-compose.prod-local.yml`
- Routes frontend, API, and WebSocket through Traefik with automatic Let's Encrypt certificates
- Hardens the Docker Compose stack with network segmentation, 127.0.0.1 port bindings, and container security settings
- Fixes CORS origin reflection vulnerability in `cors_utils.py`

## Changes
- **`config/traefik/traefik.yml`** — Static config: HTTP→HTTPS redirect, ACME cert resolver, ping healthcheck, access logging
- **`config/traefik/dynamic/tls.yml`** — TLS 1.2+ minimum with strong cipher suites
- **`config/traefik/dynamic/middlewares.yml`** — Security headers (HSTS, frame deny, nosniff, permissions policy) + rate limiting
- **`docker-compose.prod-local.yml`** — Traefik service (hardened: read-only, no-new-privileges, cap_drop ALL), Docker labels on public services, proxy/backend network segmentation, internal services bound to 127.0.0.1
- **`server/utils/web/cors_utils.py`** — Validates Origin against FRONTEND_URL whitelist instead of reflecting arbitrary origins
- **`.env.example`** — Added DOMAIN/API_DOMAIN/WS_DOMAIN/ACME_EMAIL vars and URL configuration guidance

## Test plan
- [ ] `docker compose -f docker-compose.prod-local.yml config` validates without errors
- [ ] Traefik boots cleanly with `read_only` filesystem and healthcheck passes
- [ ] Frontend, API, and WebSocket accessible via Traefik-routed domains
- [ ] Internal services (postgres, redis, vault, etc.) not reachable from outside 127.0.0.1
- [ ] CORS preflight only reflects allowed origins

## Summary by CodeRabbit

* **New Features**
  * Added Traefik reverse proxy with automatic HTTPS (Let’s Encrypt), domain-based routing for frontend, API and WebSocket, and TLS hardening.
  * Added security middleware: HSTS, XSS/frame/content-type protections, permissions policy, custom headers, and rate limiting.
  * Improved CORS handling to whitelist frontend and localhost variants.

* **Chores**
  * New compose/make targets and network defaults to support local and Traefik-backed production runs.

### DIFF
--- a/docker-compose.prod-local.yml
+++ b/docker-compose.prod-local.yml
@@ -107,6 +107,40 @@ x-common-env: &common-env
   FRONTEND_URL: ${FRONTEND_URL}
 
 services:
+  traefik:
+    image: traefik:v3.3
+    container_name: aurora-traefik
+    ports:
+      - "80:80"
+      - "443:443"
+    environment:
+      TRAEFIK_CERTIFICATESRESOLVERS_LETSENCRYPT_ACME_EMAIL: ${ACME_EMAIL}
+    volumes:
+      - /var/run/docker.sock:/var/run/docker.sock:ro
+      - ./config/traefik/traefik.yml:/etc/traefik/traefik.yml:ro
+      - ./config/traefik/dynamic:/etc/traefik/dynamic:ro
+      - traefik-certs:/letsencrypt
+      - traefik-logs:/var/log/traefik
+    security_opt:
+      - no-new-privileges:true
+    cap_drop:
+      - ALL
+    cap_add:
+      - NET_BIND_SERVICE
+    read_only: true
+    tmpfs:
+      - /tmp
+    healthcheck:
+      test: ["CMD", "traefik", "healthcheck"]
+      interval: 15s
+      timeout: 5s
+      retries: 3
+      start_period: 10s
+    restart: unless-stopped
+    networks:
+      - proxy
+      - backend
+
   aurora-server:
     build:
       context: ./server
@@ -114,8 +148,13 @@ services:
       target: prod
     image: aurora_server:latest
     container_name: aurora-server
-    ports:
-      - "${FLASK_PORT}:${FLASK_PORT}"
+    labels:
+      - "traefik.enable=true"
+      - "traefik.http.routers.api.rule=Host(`${API_DOMAIN}`)"
+      - "traefik.http.routers.api.entrypoints=websecure"
+      - "traefik.http.routers.api.tls.certresolver=letsencrypt"
+      - "traefik.http.routers.api.middlewares=security-headers@file,rate-limit@file"
+      - "traefik.http.services.api.loadbalancer.server.port=${FLASK_PORT}"
     volumes:
       - terraform-workdir:/app/terraform_workdir
       - /var/run/docker.sock:/var/run/docker.sock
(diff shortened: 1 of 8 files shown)


docker-compose.prod-local.yml
damianloch, line 1: I don't think traefik should be added to prod-local. If anythign we should have a make deploy or smtg, because having a domain for local dev is a bit of a push.

damianloch: make a make deploy which runs this prod-local and a docker-compose.traefik.yml as well

docker-compose.prod-local.yml
```diff
@@ -107,15 +107,54 @@ x-common-env: &common-env
   FRONTEND_URL: ${FRONTEND_URL}
 
 services:
+  traefik:
```
damianloch: cosmetically, traefik depends on other services so should probs wait for those before being booted up, but not a big deal

docker-compose.prod-local.yml
```diff
@@ -455,11 +522,18 @@ services:
...
+      - "traefik.http.routers.frontend.rule=Host(`${DOMAIN}`)"
+      - "traefik.http.routers.frontend.entrypoints=websecure"
+      - "traefik.http.routers.frontend.tls.certresolver=letsencrypt"
+      - "traefik.http.routers.frontend.middlewares=security-headers@file,rate-limit@file"
+      - "traefik.http.services.frontend.loadbalancer.server.port=3000"
     depends_on:
       - aurora-server
     restart: unless-stopped
+    networks:
+      - proxy
```
damianloch: The frontend service needs to reach aurora-server internally (for Next.js API routes / server-side calls via BACKEND_URL). Being only on proxy means it can't reach any backend services. It should be on both networks like aurora-server and chatbot are. This would cause SSR/API route failures at runtime

damianloch (changes requested):

docker-compose.prod-local.yml
Zarlanx: Extracted Traefik into a separate docker-compose.traefik.yml overlay file. prod-local is now clean with direct port exposure for frontend (3000), API (FLASK_PORT), and chatbot (5006). No domain config needed for local dev.

Zarlanx: Added make prod-traefik target that runs docker compose -f docker-compose.prod-local.yml -f docker-compose.traefik.yml up --build -d. Also updated make down to tear down the traefik overlay.

docker-compose.prod-local.yml
Zarlanx: Added depends_on for aurora-server, chatbot, and frontend (condition: service_started) to the traefik service in docker-compose.traefik.yml.

docker-compose.prod-local.yml
Zarlanx: Fixed. Frontend is now on the backend network in prod-local so it can reach aurora-server for SSR/API routes. When the traefik overlay is used, it also gets the proxy network via the merge.

damianloch (changes requested):
There should be additons to docs to explain how to set this up and what it does. Wasn't able to get it working

PR #167, opened 2026-03-24
Add Tailscale manage page

isiddharthsingh:
## Summary
- Add `/tailscale/manage` page with device list, SSH setup, and disconnect flow — consistent with the kubectl manage page pattern
- Route ConnectorCard to manage page when connected, onboarding when disconnected
- Update ConnectorRegistry path from `/tailscale/onboarding` to `/tailscale/manage`

## Details
Tailscale had no manage page — clicking "Manage" when connected just reopened the onboarding form. The backend APIs (devices, SSH setup, disconnect) and all Next.js proxy routes already existed but had no frontend consumer.

The new manage page shows:
- Connected tailnet name and device count
- Device list with online/offline status, IP, OS, version, tags
- Authorize and remove actions per device (with confirmation dialogs)
- Embedded SSH public key and quick setup command with copy buttons
- Disconnect with full cleanup (Vault credentials, localStorage, provider preferences, events)

## Test plan
- [ ] Disconnected state: "Connect" button on Tailscale card navigates to `/tailscale/onboarding`
- [ ] Connected state: "Manage" button navigates to `/tailscale/manage`
- [ ] Manage page loads device list and SSH setup
- [ ] Refresh button reloads devices
- [ ] Device authorize and remove actions work with confirmation dialogs
- [ ] Disconnect clears state and redirects to `/connectors`

## Summary by CodeRabbit

* **New Features**
  * Added a Tailscale management UI to view devices (online/offline), authorize or remove devices, access SSH setup info with copy-to-clipboard, refresh device list, and securely disconnect.

* **Bug Fixes / UX**
  * Connector navigation now routes users directly into the appropriate Tailscale onboarding or management flow.

* **Documentation**
  * Updated Tailscale onboarding instructions and button text for credential creation and scope guidance.

### DIFF
--- a/client/src/app/tailscale/manage/page.tsx
+++ b/client/src/app/tailscale/manage/page.tsx
@@ -64,7 +64,12 @@ export default function ManageTailscalePage() {
     const init = async () => {
       try {
         const res = await fetch("/api/tailscale/status");
-        if (!res.ok) return;
+        if (!res.ok) {
+          console.error("Tailscale status check failed:", res.status);
+          setLoadingDevices(false);
+          setLoadingSSH(false);
+          return;
+        }
         const data = await res.json();
         setTailnetName(data.tailnetName || data.tailnet || "");
         if (!data.connected) {
@@ -73,6 +78,8 @@ export default function ManageTailscalePage() {
         }
       } catch (error) {
         console.error("Error loading Tailscale status:", error);
+        setLoadingDevices(false);
+        setLoadingSSH(false);
         return;
       }
       loadDevices();
--- a/client/src/app/tailscale/manage/page.tsx
+++ b/client/src/app/tailscale/manage/page.tsx
@@ -66,6 +66,7 @@ export default function ManageTailscalePage() {
         const res = await fetch("/api/tailscale/status");
         if (!res.ok) {
           console.error("Tailscale status check failed:", res.status);
+          toast({ title: "Error", description: "Failed to load Tailscale status", variant: "destructive" });
           setLoadingDevices(false);
           setLoadingSSH(false);
           return;
@@ -78,6 +79,7 @@ export default function ManageTailscalePage() {
         }
       } catch (error) {
         console.error("Error loading Tailscale status:", error);
+        toast({ title: "Error", description: "Failed to load Tailscale status", variant: "destructive" });
         setLoadingDevices(false);
         setLoadingSSH(false);
         return;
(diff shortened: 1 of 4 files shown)


client/src/app/tailscale/manage/page.tsx
```diff
@@ -0,0 +1,525 @@
...
+  const [deviceToRemove, setDeviceToRemove] = useState<{ id: string; hostname: string } | null>(null);
+  const [copied, setCopied] = useState<"key" | "command" | null>(null);
+  const copyTimeoutRef = useRef<NodeJS.Timeout | null>(null);
+
+  useEffect(() => {
+    const init = async () => {
+      try {
+        const res = await fetch("/api/tailscale/status");
+        if (!res.ok) {
+          console.error("Tailscale status check failed:", res.status);
```
damianloch, line 68: If the status endpoint returns a non-200 response, the user just sees an empty page with loading spinners that stop. No toast or error message is shown. Worth showing something visible.

client/src/app/tailscale/manage/page.tsx
```diff
@@ -0,0 +1,525 @@
...
+  };
+
+  const confirmRemoveDevice = async () => {
+    if (!deviceToRemove) return;
+
+    try {
+      setRemovingDevice(deviceToRemove.id);
+      setDeviceToRemove(null);
+
+      const res = await fetch(`/api/tailscale/devices/${deviceToRemove.id}`, {
```
damianloch: setDeviceToRemove(null) is called on line 191, then deviceToRemove.id is referenced on line 193. This works only because setDeviceToRemove is async (state update batched), so the local closure still holds the old value. It's safe in practice but fragile and confusing

damianloch (changes requested):
The tailscale connection instruction have the wrong path to get oauth and wrong instructions for creating the key these should be updated as well

client/src/app/tailscale/manage/page.tsx
isiddharthsingh: now shows an error toast instead of silently leaving spinners

client/src/app/tailscale/manage/page.tsx
isiddharthsingh: fixed, destructured deviceToRemove into local variables before clearing state to avoid relying on stale closure behavior

isiddharthsingh:
updated the new steps

client/src/app/tailscale/manage/page.tsx
```diff
@@ -0,0 +1,528 @@
...
+      if (!res.ok) throw new Error("Failed to disconnect");
+
+      localStorage.removeItem("isTailscaleConnected");
+      window.dispatchEvent(new CustomEvent("providerStateChanged"));
+
+      try {
+        await providerPreferencesService.removeProvider("tailscale");
+        window.dispatchEvent(
+          new CustomEvent("providerPreferenceChanged", { detail: { providers: [] } })
+        );
```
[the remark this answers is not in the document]
isiddharthsingh: The pattern is actually:                                               
                                                                                                                           
  - On connect: { providers: ["<provider_name>"] } (Grafana, Datadog, PagerDuty, ThousandEyes)                             
  - On disconnect: { providers: [] } (all of them)                                                                         
                                                                                                                           
  So all disconnects consistently use { providers: [] }. The Tailscale manage page only dispatches on disconnect, and it  follows the same pattern. Not fixing — the code is correct and consistent.

PR #191, opened 2026-03-30
feat: add Prometheus connector for Alertmanager webhook integration

Zarlanx:
## Summary

Add a Prometheus connector so Aurora users can connect their existing Prometheus/Alertmanager instance and receive alerts through Aurora's incident pipeline.

## Changes

**Backend:**
- `server/connectors/prometheus_connector/api_client.py` — API client for Prometheus HTTP API (validate, alerts, targets, PromQL query)
- `server/routes/prometheus/prometheus_routes.py` — REST endpoints: connect, status, disconnect, alerts, targets, query, webhook, webhook-url
- `server/routes/prometheus/tasks.py` — Celery task for Alertmanager webhook processing through correlation pipeline + RCA triggering
- `server/main_compute.py` — Blueprint registration + CORS
- `server/utils/db/db_utils.py` — `prometheus_alerts` table with RLS
- `server/chat/background/rca_prompt_builder.py` — `build_prometheus_rca_prompt()` for AI root cause analysis
- `server/routes/incidents_routes.py` — Handle `source_type="prometheus"` in source URL and payload fetch

**Frontend:**
- `client/src/app/prometheus/auth/page.tsx` — Connection page with URL + optional token input, inline Alertmanager YAML config
- `client/src/lib/services/prometheus.ts` — API service client
- `client/src/components/connectors/ConnectorRegistry.ts` — Registered under Monitoring category
- `client/public/prometheus.svg` — Prometheus icon

## How it works

1. User connects their Prometheus instance URL (with optional Bearer token)
2. Aurora validates by hitting `/api/v1/status/buildinfo`
3. After connecting, user gets a webhook URL + Alertmanager YAML config
4. Alertmanager sends alerts to Aurora via webhook
5. Alerts are stored, correlated, and optionally trigger automated RCA

## Test plan

- [ ] Connect to a Prometheus instance via the auth page
- [ ] Verify webhook URL is displayed with Alertmanager YAML config
- [ ] Send a test webhook payload and verify it appears in prometheus_alerts table
- [ ] Verify incidents are created from firing alerts
- [ ] Verify resolved alerts are stored but don't create new incidents
- [ ] Disconnect and verify credentials are removed

## Summary by CodeRabbit

* **New Features**
  * Prometheus integration: connect/disconnect Prometheus instances with optional API token via a new settings page
  * Webhook support with copyable webhook URL and instructions for Alertmanager
  * View Prometheus alerts, scrape targets, and run PromQL queries from the UI
  * Automatic alert ingestion, incident correlation, and Root Cause Analysis workflows for Prometheus-sourced alerts


client/src/app/prometheus/auth/page.tsx
```diff
@@ -0,0 +1,248 @@
...
+    } finally {
+      setLoading(false);
+      setApiToken("");
+    }
+  };
+
+  const handleDisconnect = async () => {
+    setLoading(true);
+    try {
+      await prometheusService.disconnect();
```
damianloch, line 79: prometheus" isn't in the provider allowlist in connected-accounts/[provider]/route.ts so the DELETE will always 400. Either add it to the allowlist with a special handler that calls the backend /prometheus/disconnect endpoint, or call the dedicated backend disconnect route directly.

client/src/app/prometheus/auth/page.tsx
damianloch, line 1: Missing the /api/provider-preferences POST and providerPreferenceChanged event dispatch on both connect and disconnect. This wont show up as a connected provider in RCA

client/src/app/prometheus/auth/page.tsx
```diff
@@ -0,0 +1,248 @@
...
+  const [isCheckingStatus, setIsCheckingStatus] = useState(true);
+  const [webhookInfo, setWebhookInfo] = useState<PrometheusWebhookUrlResponse | null>(null);
+  const [copied, setCopied] = useState(false);
+
+  useEffect(() => {
+    const cached = localStorage.getItem("isPrometheusConnected") === "true";
+    if (cached) {
+      setStatus({ connected: true });
+      setIsCheckingStatus(false);
+    }
```
damianloch, line 37: The localStorage cache only stores a boolean flag while every other connector caches the full status JSON. This causes a visible flicker on load where the connected card shows no URL or version until the API responds. Cache the full status object like the other connectors do.

client/src/lib/services/prometheus.ts
```diff
@@ -0,0 +1,80 @@
...
+      return null;
+    }
+  },
+
+  async getAlerts(): Promise<{ alerts: PrometheusAlert[]; count: number }> {
+    return await jsonFetch<{ alerts: PrometheusAlert[]; count: number }>(`${API_BASE}/alerts`);
+  },
+
+  async disconnect(): Promise<void> {
+    const response = await fetch('/api/connected-accounts/prometheus', { method: 'DELETE', credentials: 'include' });
```
damianloch, line 75: This endpoint will 400 because "prometheus" was never added to the provider allowlist in connected-accounts/[provider]/route.ts. This is the root cause of the disconnect bug flagged on the auth page.

client/src/lib/services/prometheus.ts
```diff
@@ -0,0 +1,80 @@
...
+    }
+  },
+
+  async connect(baseUrl: string, apiToken?: string): Promise<PrometheusStatus> {
+    const raw = await jsonFetch<{ success: boolean; connected: boolean; error?: string; version?: string; baseUrl?: string }>(
+      `${API_BASE}/connect`,
+      { method: 'POST', body: JSON.stringify({ baseUrl, apiToken: apiToken || undefined }) },
+    );
+    if (!raw.success && !raw.connected) {
+      throw new Error(raw.error || 'Connection failed');
```
damianloch, line 56: this is dead code

client/src/lib/services/prometheus.ts
damianloch, line 1: Are requests being proxied to the backend? I dont see a route.ts anywhere

server/routes/prometheus/tasks.py
damianloch, line 1: The docstring says "SSE broadcast" but there's no actual broadcast_incident_update_to_user_connections call after incident creation

server/routes/prometheus/tasks.py
```diff
@@ -0,0 +1,259 @@
...
+                       VALUES (%s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s)
+                       RETURNING id""",
+                    (
+                        user_id, alert_name, alert_status,
+                        labels.get("severity", "unknown"),
+                        instance, group_labels.get("alertname"),
+                        fingerprint, json.dumps(labels),
+                        json.dumps(alert.get("annotations", {})),
+                        alert.get("generatorURL"),
+                        json.dumps(raw_payload), received_at,
```
damianloch, line 137: datadog and grafana store only json.dumps(alert) instead, not entire raw_payload (since a webhook can have multiple alerts)

server/routes/prometheus/prometheus_routes.py
```diff
@@ -0,0 +1,307 @@
...
+    ngrok_url = os.getenv("NGROK_URL", "").rstrip("/")
+    backend_url = os.getenv("NEXT_PUBLIC_BACKEND_URL", "").rstrip("/")
+    base_url = ngrok_url if ngrok_url and backend_url.startswith("http://localhost") else backend_url
+    if not base_url:
+        base_url = request.host_url.rstrip("/")
+
+    return jsonify({
+        "webhookUrl": f"{base_url}/prometheus/webhook/{user_id}",
+        "signatureHeader": "X-Aurora-Signature",
+        "signatureAlgorithm": "HMAC-SHA256 of request body using your webhook secret",
```
damianloch, line 295: webhook-url response tells the user about X-Aurora-Signature and the HMAC algorithm but doesn't include the actual secret needed to compute signatures. This makes the instructions misleading

server/routes/prometheus/prometheus_routes.py
```diff
@@ -0,0 +1,307 @@
...
+            return jsonify({"error": "Missing X-Aurora-Signature header"}), 401
+        expected = hmac.new(webhook_secret.encode(), request.get_data(), hashlib.sha256).hexdigest()
+        if not hmac.compare_digest(expected, signature):
+            logger.warning("[PROMETHEUS] Webhook rejected: invalid signature for user %s", user_id[:50])
+            return jsonify({"error": "Invalid webhook signature"}), 401
+
+    payload = request.get_json(silent=True) or {}
+    logger.info("[PROMETHEUS] Received webhook for user %s (%d alerts)", user_id, len(payload.get("alerts", [])))
+
+    _REDACTED_HEADERS = {"authorization", "cookie", "set-cookie", "proxy-authorization", "x-api-key"}
```
damianloch, line 262: make a module level constant

server/routes/prometheus/prometheus_routes.py
```diff
@@ -0,0 +1,307 @@
...
+@prometheus_bp.route("/webhook/<user_id>", methods=["POST", "OPTIONS"])
+def webhook(user_id: str):
+    """Receive webhook events from Prometheus Alertmanager."""
+    if request.method == "OPTIONS":
+        return create_cors_response()
+
+    if not user_id:
+        return jsonify({"error": "user_id is required"}), 400
+
+    if not _verify_webhook_user(user_id):
```
damianloch, line 234: _verify_webhook_user on line 234 and get_token_data on line 239 both query user_tokens for the same user/provider combo

server/routes/prometheus/prometheus_routes.py
```diff
@@ -0,0 +1,307 @@
...
+    logger.info("[PROMETHEUS] Connecting user %s to %s", user_id, base_url)
+
+    client = PrometheusClient(base_url, api_token)
+    try:
+        validation = client.validate()
+    except PrometheusAPIError as exc:
+        logger.error("[PROMETHEUS] Validation failed for user %s: %s", user_id, exc)
+        return jsonify({"error": str(exc)}), 502
+
+    webhook_secret = secrets.token_hex(32)
```
damianloch, line 59: cursor is flagging this: This is broken -- webhook_secret is always generated on connect (line 59) and always present in creds, so the HMAC check on line 250-253 always runs. But the secret is never returned to the user anywhere (not in /connect, not in /webhook-url, not in the frontend). Alertmanager also doesn't natively support custom HMAC signature headers. Every real webhook will be rejected with 401 "Missing X-Aurora-Signature header". Either expose the secret so users can build a signing proxy, make signature validation opt-in, or remove it.

damianloch (changes requested):
Reviewed a good amount but not all files since some needed rebase

PR #239, opened 2026-04-10
ci: publish multi-arch (amd64 + arm64) images natively

isiddharthsingh:
## Summary

- Rewrite `.github/workflows/publish-images.yml` to a native matrix: four parallel `build` jobs (server/frontend × amd64/arm64) run on `ubuntu-24.04` and `ubuntu-24.04-arm`, push per-arch digests, then two `merge` jobs stitch them into a single manifest list per tag and verify both platforms are present. Replaces the previous QEMU cross-build.
- Make `make deploy-build` multi-arch by default (`linux/amd64,linux/arm64`, overridable with `PLATFORMS=...`) via a `docker-container` buildx builder, and inspect each image's manifest list after push.
- Add `website/docs/multi-arch-images.md` explaining host auto-resolve, `--platform` overrides, Kubernetes behavior, and per-`make`-target arch behavior; link it in the Docusaurus sidebar.

Result: a single image tag (`:edge`, `:v1.2.3`, `:sha-abc123`) transparently serves both `linux/amd64` (x86_64 dev / CI / EC2) and `linux/arm64` (Apple Silicon / Graviton / Ampere) via an OCI manifest list. No Dockerfile, compose, or Helm changes are required — `server/Dockerfile` was already arch-safe and `client/Dockerfile` uses a multi-arch Bun base.

### Build time

Native matrix wall-clock on this branch: **10m 20s** total (4 builds in parallel; slowest was frontend arm64 at ~9:44).
Previous QEMU runs on `main`: ~19m (#233 python upgrade) and ~20m (#229 cache fixes).
Roughly **2× faster**, even on a cold GHA cache — subsequent runs will be faster still.

## Test plan

- [x] YAML / Makefile lint: `python3 -c 'import yaml; yaml.safe_load(...)'`, `make -n deploy-build`
- [x] CI workflow on this branch (run `24259339854`) — 4 native builds + 2 merges + summary all green
- [x] Manifest inspection — both `aurora-server:sha-0869ee9` and `aurora-frontend:sha-0869ee9` list `linux/amd64` and `linux/arm64` with distinct digests
- [x] Native pull on Apple Silicon → `aarch64` inside the container
- [x] Forced `docker pull --platform linux/amd64 ...` → `x86_64` inside the container (same tag)
- [x] `make prod-prebuilt VERSION=sha-0869ee9` smoke test on Apple Silicon: all 4 Aurora services (`aurora-server`, `aurora-frontend`, `aurora-celery_worker`, `aurora-chatbot`) run `aarch64`; backend `/health` and frontend `/` both return 200
- [x] `make deploy-build` end-to-end against a test registry + multi-arch manifest verification — see comment below
- [x] Mixed-arch EKS smoke — added a t4g.medium arm64 nodegroup to the test cluster; same image reference resolved to x86_64 on the amd64 node and aarch64 on the arm64 node. See comment below.

## Summary by CodeRabbit

* **New Features**
  * Multi-architecture container images for Linux AMD64 and ARM64 are published for server and frontend; Docker/Kubernetes will select the appropriate variant on pull.

* **Chores**
  * CI now builds per-architecture images, uploads per-arch digests, merges them into unified multi-arch manifests, and verifies manifest contents.
  * Build tooling updated to support configurable target platforms and post-publish inspections; logs now reflect multi-arch behavior.

* **Documentation**
  * New “Multi-arch Images” guide covering usage, verification, Kubernetes behavior, and air-gapped workflows.

### DIFF
--- a/.github/workflows/publish-images.yml
+++ b/.github/workflows/publish-images.yml
@@ -112,6 +112,7 @@ jobs:
       - name: Export digest
         env:
           DIGEST: ${{ steps.build-server.outputs.digest || steps.build-frontend.outputs.digest }}
+          ARCH: ${{ matrix.arch }}
         run: |
           set -euo pipefail
           if [ -z "${DIGEST:-}" ]; then
@@ -119,8 +120,11 @@ jobs:
             exit 1
           fi
           mkdir -p /tmp/digests
-          # Strip the "sha256:" prefix so the filename is just the hex digest.
-          touch "/tmp/digests/${DIGEST#sha256:}"
+          # File name = arch, content = sha256:<hex>. Per-arch filenames let the
+          # merge job validate that every expected arch is present *before*
+          # calling `imagetools create`, so we never publish a half-populated
+          # manifest list to the registry.
+          printf '%s' "$DIGEST" > "/tmp/digests/${ARCH}"
 
       - name: Upload digest artifact
         uses: actions/upload-artifact@v4
@@ -175,16 +179,31 @@ jobs:
           IMAGE_NAME: ghcr.io/arvo-ai/aurora-${{ matrix.image }}
         run: |
           set -euo pipefail
-          tags=$(jq -cr '.tags | map("-t " + .) | join(" ")' <<< "$DOCKER_METADATA_OUTPUT_JSON")
-          sources=""
-          for digest_file in *; do
-            [ -f "$digest_file" ] || continue
-            sources="$sources ${IMAGE_NAME}@sha256:${digest_file}"
+
+          # Fail fast if any expected arch's digest is missing, *before* we
+          # touch the registry. Keep this list in sync with the build matrix
+          # (currently amd64 + arm64).
+          expected_arches="amd64 arm64"
+          missing=""
+          for arch in $expected_arches; do
+            if [ ! -s "$arch" ]; then
+              missing="${missing} $arch"
+            fi
           done
-          if [ -z "$sources" ]; then
-            echo "No digest files found to merge" >&2
+          if [ -n "$missing" ]; then
+            echo "Missing digest file(s) for arch(es):$missing" >&2
+            echo "Contents of /tmp/digests:" >&2
+            ls -la >&2
             exit 1
           fi
+
+          tags=$(jq -cr '.tags | map("-t " + .) | join(" ")' <<< "$DOCKER_METADATA_OUTPUT_JSON")
+          sources=""
+          for arch in $expected_arches; do
+            digest=$(cat "$arch")
+            sources="${sources} ${IMAGE_NAME}@${digest}"
+          done
+
           # shellcheck disable=SC2086
           docker buildx imagetools create $tags $sources
 
--- a/.github/workflows/publish-images.yml
+++ b/.github/workflows/publish-images.yml
@@ -24,18 +24,41 @@ permissions:
   packages: write
 
 jobs:
-  publish-server:
-    name: Publish aurora-server
-    runs-on: ubuntu-24.04
-    # Only run on the main repository, not on forks
+  build:
+    name: Build ${{ matrix.image }} (${{ matrix.platform }})
     if: github.repository == 'Arvo-AI/aurora'
-
+    runs-on: ${{ matrix.runner }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - image: server
+            context: ./server
+            dockerfile: ./server/Dockerfile
+            platform: linux/amd64
+            arch: amd64
+            runner: ubuntu-24.04
+          - image: server
+            context: ./server
+            dockerfile: ./server/Dockerfile
+            platform: linux/arm64
+            arch: arm64
+            runner: ubuntu-24.04-arm
+          - image: frontend
+            context: ./client
+            dockerfile: ./client/Dockerfile
+            platform: linux/amd64
+            arch: amd64
+            runner: ubuntu-24.04
+          - image: frontend
+            context: ./client
+            dockerfile: ./client/Dockerfile
+            platform: linux/arm64
+            arch: arm64
+            runner: ubuntu-24.04-arm
     steps:
       - name: Checkout repository
         uses: actions/checkout@v5
-
-      - name: Set up QEMU (for multi-arch builds)
-        uses: docker/setup-qemu-action@v3
 
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3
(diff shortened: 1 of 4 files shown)


isiddharthsingh:
## End-to-end verification complete

Finished the two remaining test-plan items: `make deploy-build` against a real registry, and a mixed-arch Kubernetes smoke on EKS.

### `make deploy-build` → ECR

Ran against a private test registry. Both images landed as proper OCI manifest lists:

```
Name:      <registry>/aurora-server:0869ee90
MediaType: application/vnd.oci.image.index.v1+json
Manifests:
  linux/amd64    sha256:62637f1afa2da9570a57fe23c39c79d1f133f533166d1df50c891b24ad51585e
  linux/arm64    sha256:85640e9eb46fbb2a20e9b64650cd1c3c39d209c3a7187182fa48fee4cffaa919

Name:      <registry>/aurora-frontend:0869ee90
MediaType: application/vnd.oci.image.index.v1+json
Manifests:
  linux/amd64    sha256:3599d68ba120168ccadb385c1211a3870a18b3bacbba895c7c75ba535a9e1504
  linux/arm64    sha256:f0f1598dfdc9541048fe85673468e5182b9b3a5b96f68d7bfb68237913b8d1c8
```

Wall-clock: **27:58** on an Apple Silicon host. This is expected and is the dominant cost of local multi-arch: the amd64 leg runs under QEMU emulation while arm64 is native, so it's ~3× slower than CI's ~10 min (where each arch runs on its native runner). The new inline warning in the Makefile comment block calls this out and points users to CI as the preferred publish path.

Note: the new `docker-container` buildx builder is a separate buildkit instance with its own storage inside Docker Desktop's VM — if the VM disk is tight on space, the non-native leg can fail mid-build with `E: You don't have enough free space in /var/cache/apt/archives/`. Fix: `docker builder prune -af` (or a broader `docker system prune -a --volumes`) before running `deploy-build`. Not a regression from this PR — the same symptom would have hit the old target under equivalent pressure — but worth knowing because multi-arch makes it more likely to trigger.

### Mixed-arch EKS smoke

Temporarily added a `t4g.medium` (arm64) nodegroup to an existing EKS cluster so the cluster had both amd64 and arm64 nodes:

```
NAME                                           STATUS   ARCH    INSTANCE-TYPE
ip-192-168-3-120.us-east-2.compute.internal    Ready    arm64   t4g.medium
ip-192-168-70-121.us-east-2.compute.internal   Ready    amd64   t3.large
ip-192-168-9-147.us-east-2.compute.internal    Ready    amd64   t3.large
```

Applied two smoke pods pinned by `nodeSelector: kubernetes.io/arch` using the **same** image reference (`aurora-server:0869ee90`) and the **same** pulled digest (`sha256:73c7cc5e…`, the manifest list). The EKS kubelet on each node transparently resolved the matching per-platform variant with zero Helm, zero platform config, zero override flags:

| Pod | Node | Node arch | `uname -m` | Kernel |
|---|---|---|---|---|
| `arch-smoke-amd64` | `ip-192-168-70-121` (t3.large) | amd64 | `x86_64` | `6.12.73-95.123.amzn2023.x86_64` |
| `arch-smoke-arm64` | `ip-192-168-3-120` (t4g.medium) | arm64 | `aarch64` | `6.12.77-99.140.amzn2023.aarch64` |

Both pods started cleanly with no architecture warnings, no exec-format errors, and the Aurora server image reported the correct AL2023 kernel for its host arch. This validates the full chain:

1. `server/Dockerfile`'s `uname -m` branches (Terraform/AWS CLI/Azure CLI/kubectl per-arch downloads) build cleanly on arm64 natively.
2. `docker buildx imagetools create` produces an ECR manifest list consumable by the EKS kubelet.
3. No Helm `nodeSelector` or `nodeAffinity` is needed for transparent mixed-arch scheduling — when a workload replica lands on an arm64 Graviton node, it just gets the arm64 variant.

### Cleanup

- Smoke pods deleted.
- `arm64-smoke` nodegroup being torn down (`eksctl delete nodegroup`).
- `deploy/helm/aurora/values.generated.yaml` restored from a pre-`deploy-build` backup — the file is byte-identical to the baseline so the `0869ee90` tag update made by `deploy-build` doesn't leak into an accidental commit.

Test plan in the PR body can be marked fully green.

website/docs/multi-arch-images.md
```diff
@@ -0,0 +1,125 @@
...
+## Which images are multi-arch?
+
+- `ghcr.io/arvo-ai/aurora-server` — Flask API, Celery worker, Celery beat, and chatbot are all built from `server/Dockerfile`.
+- `ghcr.io/arvo-ai/aurora-frontend` — Next.js app built from `client/Dockerfile`.
+
+Both are published as **OCI image indexes** (a.k.a. manifest lists) containing `linux/amd64` and `linux/arm64` variants under the same tag.
+
+## How it works (OCI manifest lists)
+
+When you run `docker pull` against a multi-arch image, the registry returns a manifest list describing every available architecture. The Docker client reads your host's `GOARCH` and pulls only the matching variant. This is the same mechanism used by official images like `python:3.12-slim`, `node:20-alpine`, and `postgres:15-alpine`.
```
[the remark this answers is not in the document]
isiddharthsingh: its acceptable

Zarlanx (commented):
Neither of these blocks merging.

.github/workflows/publish-images.yml
```diff
@@ -47,42 +70,84 @@ jobs:
...
+        uses: actions/upload-artifact@v4
+        with:
+          name: digests-${{ matrix.image }}-${{ matrix.arch }}
+          path: /tmp/digests/*
+          if-no-files-found: error
+          retention-days: 1
+
+  merge:
+    name: Merge ${{ matrix.image }} manifest
+    needs: build
```
Zarlanx, line 139: needs: build gates on all 4 matrix jobs so if one arch of one image fails neither manifest gets created, not a bug but worth confirming its intentional

.github/workflows/publish-images.yml
```diff
@@ -47,42 +70,84 @@ jobs:
...
         uses: docker/metadata-action@v5
         with:
-          images: ${{ env.SERVER_IMAGE }}
-          tags: |
-            type=semver,pattern={{version}}
-            type=semver,pattern={{major}}.{{minor}}
-            type=raw,value=latest,enable=${{ startsWith(github.ref, 'refs/tags/v') && !contains(github.ref_name, '-') }}
-            type=edge,branch=main
-            type=sha,prefix=sha-,format=short
+          images: ghcr.io/arvo-ai/aurora-${{ matrix.image }}
```
Zarlanx, line 75: SERVER_IMAGE and FRONTEND_IMAGE env vars (lines 19-20) are now dead, everything uses this inline pattern instead, safe to drop them

.github/workflows/publish-images.yml
isiddharthsingh: yes this is intentional

.github/workflows/publish-images.yml
isiddharthsingh: removed

Zarlanx (approved):

PR #240, opened 2026-04-11
feat(gcp): add service account JSON auth alongside OAuth

isiddharthsingh:
## Summary

Adds a second way to connect GCP alongside the existing Google OAuth flow: users can now upload a Service Account JSON key directly. Meant for enterprise users whose Google Workspace doesn't allow OAuth, service-to-service setups, or anyone who prefers managing their own credentials rather than letting Aurora provision them.

On connect, Aurora validates the key against Google, auto-discovers every project the SA can see (so one key with org-level browser/viewer access unlocks all accessible projects in one step), and stores it encrypted in Vault. The uploaded SA becomes Aurora's identity directly — no per-user service accounts get created in the user's GCP projects, no impersonation hop, no extra IAM bindings to set up. Whatever roles the user already granted the SA in GCP IAM are exactly the roles Aurora has.

The connect UI now has tabs (OAuth | Service Account) so users pick at connect time, and can switch modes later via a confirm dialog. The manage dialog shows a badge so it's clear which mode is active. The uploaded key file is read in the browser and never actually transmitted — only the parsed JSON string is sent to Aurora.

End to end, every cloud tool path works in SA mode: `gcloud`, `gsutil`, `kubectl` (via `gcloud container clusters get-credentials`), and `terraform`. Non-GCP connector code is untouched.

## Test plan

- [x] SA key upload (file picker + paste) with inline validation of required fields
- [x] Multi-project auto-discovery after granting org-level `roles/viewer` — 3 projects stored
- [x] Manage dialog renders "Service Account" badge and full project list
- [x] 21 gcloud tool calls across a single heavy chat session — every one in SA mode, zero impersonation errors
- [x] `gcloud compute instances list` against the billing-enabled root project returns real VM data in under 2s
- [x] `kubectl` via GKE credential helper works end-to-end
- [x] `gsutil ls` works in chat (no stray `-i` impersonation flag injection)
- [x] Project-management routes short-circuit correctly in SA mode (no spurious IAM calls)
- [x] Disconnect + reconnect clears stale state cleanly
- [ ] Revoked-key path returns the stable user-safe error message
- [x] Switch-auth-method confirm dialog flow

## Summary by CodeRabbit

* **New Features**
  * Added direct Google Cloud Platform service account authentication via service account key upload (JSON file or paste).
  * Introduced authentication method indicator displaying current GCP connection type (OAuth or Service Account).
  * Enhanced GCP project management with improved handling for service account credentials.

* **Chores**
  * Improved credential caching and cleanup mechanisms.


server/routes/gcp/auth.py
```diff
@@ -224,6 +224,236 @@ def force_disconnect_gcp(user_id):
...
+    # per-user Aurora SA and grants impersonation, which only applies to OAuth
+    # mode. The uploaded SA key IS the working identity for SA mode.
+
+    # Clear any OAuth-era preferences so stale Aurora-created SA metadata and
+    # the previous root project don't leak into the SA-mode UI. The user can
+    # set a new root project via the manage dialog after connecting.
+    try:
+        from utils.auth.stateless_auth import store_user_preference
+        store_user_preference(user_id, "gcp_service_accounts", None)
+        store_user_preference(user_id, "gcp_root_project", None)
```
OlivierTrudeau: All of this is only useful for direct switching but you removed it now. So this and the redis cleanup can both be removed

server/routes/gcp/auth.py
```diff
@@ -224,6 +224,236 @@ def force_disconnect_gcp(user_id):
...
+        logging.warning(
+            "GCP SA connect: failed to clear stale GCP preferences (error_type=%s)",
+            type(exc).__name__,
+        )
+
+    # Clear Redis GCP caches for the user (mirrors /api/gcp/force-disconnect).
+    org_id = get_org_id_from_request()
+    conn = None
+    cursor = None
+    secret_ref_to_clear = None
```
OlivierTrudeau: Remove this as well. This and the step before should be done on disconnect not on connect

server/routes/gcp/auth.py
```diff
@@ -224,6 +224,236 @@ def force_disconnect_gcp(user_id):
...
+                "GCP SA connect: failed to clear Redis cache for user %s (error_type=%s)",
+                user_id,
+                type(exc).__name__,
+            )
+
+    # Also clear the per-user GCP setup cache (Redis + in-process env vars +
+    # temp cred files). Critical on OAuth → SA switches so stale impersonation
+    # env vars from a prior OAuth run can't leak into SA-mode gcloud calls.
+    try:
+        from chat.backend.agent.tools.auth.gcp_cached_auth import clear_gcp_cache_for_user
```
OlivierTrudeau: This should also be removed I think. all should be on disconnect

OlivierTrudeau (changes requested):

server/chat/backend/agent/tools/cloud_exec_tool.py
```diff
@@ -565,24 +602,38 @@ def setup_gcp_environment_isolated(user_id: str, selected_project_id: str | None
...
-            "HOME": "/home/appuser",  # Use terminal pod's writable home directory
+            "HOME": "/home/appuser",
             "USER": os.environ.get("USER", ""),
             "GOOGLE_OAUTH_ACCESS_TOKEN": access_token,
             "CLOUDSDK_AUTH_ACCESS_TOKEN": access_token,
             "GOOGLE_CLOUD_PROJECT": project_id,
-            "CLOUDSDK_AUTH_IMPERSONATE_SERVICE_ACCOUNT": sa_email,
-            "CLOUDSDK_IMPERSONATE_SERVICE_ACCOUNT": sa_email,
             "CLOUDSDK_CONFIG": "/tmp/.gcloud",
         }
```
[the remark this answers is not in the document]
isiddharthsingh: Fixed in b9fad701. setup_gcp_environment_isolated now creates a per-user /tmp/.gcloud-{user_id} directory via os.makedirs(..., exist_ok=True) and sets CLOUDSDK_CONFIG to it. User IDs are UUIDs so they're safe to embed in the path directly. Falls back to the shared /tmp/.gcloud with a logged warning only if mkdir raises OSError (i.e. /tmp itself is unwritable, which would be a bigger infra problem). Each concurrent request now gets its own gcloud config/cache dir — no more cross-user state leak.

OlivierTrudeau (approved):

damianloch (approved):

PR #242, opened 2026-04-11
fix: gate confluence_runbook_parse behind credential check

Zarlanx:
confluence_runbook_parse was registered in the unconditional tool_functions list, so the agent saw it for every user regardless of whether they had Confluence credentials. When an alert payload included a runbook_url, the agent would call this tool, fail because no Confluence token existed, and the entire RCA would error out.

Moved the tool registration into the existing credential-gated Confluence block (alongside confluence_search_similar, confluence_search_runbooks, and confluence_fetch_page) so it only appears when the user actually has a Confluence connection. Also removed the now-dead elif branch from the unconditional loop that handled ConfluenceRunbookArgs schema binding.

## Summary by CodeRabbit

* **Bug Fixes**
  * Confluence tools no longer activate unless valid Confluence credentials are present.

* **Improvements**
  * Confluence integration now registers the full set of tools only when authenticated, with consistent user-context handling, completion notifications, and optional capture.
  * Logging updated to reflect the correct number of Confluence tools when enabled.

### DIFF
--- a/server/chat/backend/agent/tools/cloud_tools.py
+++ b/server/chat/backend/agent/tools/cloud_tools.py
@@ -1163,7 +1163,6 @@ Once you identify which account has the issue, pass account_id (e.g. 'account') 
         (cloud_exec_wrapper, "cloud_exec"),
         (terminal_exec, "terminal_exec"),
         (tailscale_ssh, "tailscale_ssh"),
-        (confluence_runbook_parse, "confluence_runbook_parse"),
         (on_prem_kubectl, "on_prem_kubectl"),
         (analyze_zip_file, "analyze_zip_file"),
         # (web_search, "web_search"),  # Moved to dedicated registration below with explicit args_schema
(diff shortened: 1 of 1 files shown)


server/chat/backend/agent/tools/cloud_tools.py
```diff
@@ -1319,12 +1318,19 @@ def cloud_exec_wrapper(provider: str, command: str, output_file: Optional[str] =
...
-                args_schema=ConfluenceRunbookArgs,
+                description=(
+                    "Trigger a full automated Root Cause Analysis investigation. "
+                    "Use this when the user reports an operational incident or describes symptoms "
+                    "that warrant investigation (e.g. high CPU, errors, latency spikes, outages). "
+                    "Creates an incident and dispatches a background RCA using all connected integrations. "
+                    "Parameters: issue_description (required), title (optional), service (optional), "
+                    "severity (optional: critical/high/medium/low)."
+                ),
+                args_schema=TriggerRCAArgs,
```
damianloch, line 1333: this is a duplicate code block

damianloch (changes requested):

server/chat/backend/agent/tools/cloud_tools.py
Zarlanx: Fixed in 75c88a99. Removed the duplicate elif block, the first match always wins so the second was dead code.

damianloch (approved):

PR #251, opened 2026-04-13
fix: create active session for Next Step execution

beng360:
## Summary
- `mark_suggestion_executed` was using `create_background_chat_session` which creates sessions with `status='in_progress'` — designed for Celery background tasks that update the status when done
- But Next Step execution happens via the live WebSocket chatbot, which never updates the DB session status — leaving it stuck as "in progress" with a permanent loading spinner in the chat sidebar
- Replaced with a direct INSERT that creates the session with `status='active'` and `selectedMode='agent'`, matching the actual execution flow

## Root cause
Introduced in PR #243 during the CodeRabbit fix "stop navigation on mark-executed failure & persist executionSessionId" — `create_background_chat_session` was the wrong abstraction for this use case.

## Test plan
- [ ] Click "Execute" on a Next Step suggestion
- [ ] Verify the new chat session appears in the sidebar WITHOUT a loading spinner
- [ ] Verify the command executes normally via WebSocket
- [ ] Verify "View Output" button works after execution

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed chat session creation when executing incident suggestions. Sessions are now properly initialized with agent mode configuration, active status tracking, and correct security context to ensure proper incident access control.


isiddharthsingh:
@claude review this pr

beng360:
Addressed all review items in 7e1b7db:

**#1 `execution_session_id` never set — propagation broken** ✅ Fixed
The frontend now passes `chatSessionId` in the request body to `mark-executed`. The backend reads it and writes it to `execution_session_id`, so `_propagate_suggestion_status` can match on session completion and transition `executed` → `completed`/`failed`.

**#2 Fetched `execution_status` unused in SELECT** ✅ Fixed
Removed from the SELECT — now only fetches `s.id, s.incident_id`.

**#3 No idempotency guard** ✅ Documented
Re-execution is intentional (the UI has a "Re-execute" button). Added a docstring note explaining this is an idempotent update path.

**#4 "View Output" points to RCA session** — Acknowledged, no change needed
Correct — execution now happens in the RCA session itself, so navigating there is the right behavior. Scrolling to the execution output is a valid future UX improvement.

**#5 `isExecuting` never reset on success** ✅ Fixed
Added `setIsExecuting(false)` on the success path before `onClose()`.

**#6 `hasBlockChild` heuristic too broad** ✅ Fixed
Tightened the check to `c.type === CodeBlock` instead of `typeof c.type !== 'string'`, so only the known block-level component triggers the `<div>` wrapper.

isiddharthsingh (approved):

PR #257, opened 2026-04-14
refactor: skills-based prompt architecture

Zarlanx:
## Summary

- **Replaces the monolithic prompt builder** (1,986 lines) with a modular, file-based skills architecture
- **prompt_builder.py**: 1,986 → 62 lines (97% reduction), decomposed into 8 focused sub-modules
- **rca_prompt_builder.py**: refactored to load integration guidance from skill files via SkillRegistry
- **21 integration SKILL.md files** — one per integration (Datadog, Splunk, GitHub, Jira, OpsGenie, etc.)
- **8 core system prompt segments** as markdown files under `skills/core/`
- **17 RCA segment files** — background mode + investigation templates as markdown
- **9 provider investigation guides** under `skills/rca/`
- **Token-budgeted RCA loading** (4,000 token cap with priority-based skill loading)
- **`load_skill` LangGraph tool** — agent loads integration details on-demand instead of always-in-context

### Architecture

```
skills/
  core/               8 markdown files (system prompt segments)
  integrations/       21 SKILL.md files (one per integration)
  rca/
    background/       10 background mode segment files
    segments/         7 RCA investigation segment files
    *.md              9 provider investigation guides
  registry.py         Singleton skill registry + token budgeting
  loader.py           YAML frontmatter parser + template engine + core loader
  load_skill_tool.py  LangGraph on-demand loading tool

prompt/
  schema.py           PromptSegments dataclass
  provider_rules.py   Provider constraints, context, prerequisites
  context_fetchers.py Manual VM access, knowledge base memory
  background.py       Background RCA mode segment builder
  composer.py         Segment assembly + build_prompt_segments
  cache_registration.py  Prefix cache breakpoint registration
  prompt_builder.py   Thin facade re-exporting public API
```

### Why

1. **Context bloat** — every integration's full instructions were in every prompt regardless of relevance; now only a ~300-token index is always-loaded
2. **Prompt in Python strings** — tweaking investigation steps meant editing Python; now anyone can edit markdown
3. **No token budgeting** — adding integrations degraded quality; now RCA has a 4,000-token budget with priority loading
4. **Duplicated patterns** — Jenkins/CloudBees were copy-pasted; single SkillRegistry handles both interactive + RCA

## Test plan

- [ ] Verify `build_system_invariant()` loads all 8 core segments correctly
- [ ] Verify `SkillRegistry.build_index()` returns compact index for connected integrations
- [ ] Verify `load_skill` tool returns full skill content for connected integrations
- [ ] Verify `load_skills_for_rca()` respects token budget and priority ordering
- [ ] Verify connection checks dispatch correctly for all 3 methods (get_credentials_from_db, get_token_data, is_connected_function)
- [ ] Test RCA prompt generation for webhook-triggered alerts (Datadog, Grafana, PagerDuty, OpsGenie)
- [ ] Test interactive chat prompt assembly with multiple integrations connected
- [ ] Verify prefix cache breakpoints still register correctly
- [ ] Verify backward compatibility — existing imports from `prompt_builder` still resolve

## Summary by CodeRabbit

* **New Features**
  * Modular prompt assembly with reusable segments and an in-chat "load skill" tool for on‑demand integrations.
  * Enhanced background RCA mode with structured investigation flows, provider-aware guidance, and VM/SSH access helpers.

* **Documentation**
  * Large collection of RCA playbooks, per-integration guides, tool-selection, error-handling, and investigation templates.

* **Bug Fixes**
  * Retry behavior for initial message auto-send improved; suggestion modal no longer aborts flow on execution failures.


server/chat/backend/agent/skills/registry.py
```diff
@@ -0,0 +1,541 @@
...
+        """Fetch recent deployment records from the database."""
+        try:
+            from utils.db.connection_pool import db_pool
+
+            with db_pool.get_admin_connection() as conn:
+                with conn.cursor() as cur:
+                    cur.execute(
+                        """SELECT service, environment, result, build_number,
+                                  commit_sha, trace_id, webhook_received_at
+                           FROM deployment_events
```
beng360: **Bug: wrong table name and column** — `deployment_events` does not exist in the schema (only `jenkins_deployment_events` and `spinnaker_deployment_events` are defined). The `except Exception: return []` silently swallows the SQL error, so `{jenkins_deploys_section}`/`{cloudbees_deploys_section}` template vars in skill files would always render as "No recent deployments found."\n\nAlso the column name is wrong: the real table uses `received_at`, not `webhook_received_at`.\n\nFix:\n```python\ncur.execute(\n    """SELECT service, environment, result, build_number,\n              commit_sha, trace_id, received_at\n       FROM jenkins_deployment_events\n       WHERE user_id = %s AND provider = %s\n             AND received_at >= NOW() - INTERVAL '24 hours'\n       ORDER BY received_at DESC\n       LIMIT 10""",\n    (user_id, provider),\n)\n```

server/chat/backend/agent/skills/registry.py
```diff
@@ -0,0 +1,541 @@
...
+        lines = [
+            "CONNECTED INTEGRATIONS:",
+            "Call load_skill('id') for detailed guidance before using an integration's tools.",
+            "",
+        ]
+        for meta in sorted(connected, key=lambda m: m.name):
+            tools_str = ", ".join(meta.tools[:4])
+            if len(meta.tools) > 4:
+                tools_str += ", ..."
+            lines.append(f"- {meta.id}: {meta.index} [tools: {tools_str}]")
```
beng360: **Cosmetic: grafana shows `[tools: ]` in the agent index** — since grafana has `tools: []`, this line renders as:\n\n`- grafana: Grafana — observation-only alert ingestion, no CLI tools [tools: ]`\n\nThe trailing `[tools: ]` is noise. Suggest skipping the suffix when the tools list is empty:\n\n```python\nfor meta in sorted(connected, key=lambda m: m.name):\n    if meta.tools:\n        tools_str = ', '.join(meta.tools[:4])\n        if len(meta.tools) > 4:\n            tools_str += ', ...'\n        lines.append(f'- {meta.id}: {meta.index} [tools: {tools_str}]')\n    else:\n        lines.append(f'- {meta.id}: {meta.index}')\n```

server/chat/backend/agent/skills/load_skill_tool.py
```diff
@@ -0,0 +1,54 @@
...
+    if not user_id:
+        return json.dumps({"error": "No user context available"})
+
+    try:
+        from .registry import SkillRegistry
+
+        registry = SkillRegistry.get_instance()
+        result = registry.load_skill(skill_id, user_id)
+
+        if not result.is_connected:
```
beng360, line 68: **Agent UX: unknown skill ID and disconnected skill return the same error** — if the agent calls `load_skill('typo_id')` it gets `"Integration 'typo_id' is not connected."`, which is the same message as a valid-but-disconnected skill. The agent can't self-correct a typo.\n\nThe `SkillLoadResult.content` already has the right message (`"Unknown skill: 'x'"`) but `load_skill_tool.py` overrides it.\n\nFix:\n```python\nif not result.is_connected:\n    all_ids = registry.get_all_skill_ids()\n    if skill_id not in all_ids:\n        return json.dumps({\n            "error": f"Unknown skill '{skill_id}'. Check the CONNECTED INTEGRATIONS index for valid skill IDs.",\n            "available": registry.get_connected_skill_ids(user_id),\n        })\n    return json.dumps({\n        "error": f"Integration '{skill_id}' is not connected for this user.",\n        "hint": "Check the CONNECTED INTEGRATIONS index for available skills.",\n        "available": registry.get_connected_skill_ids(user_id),\n    })\n```

server/chat/backend/agent/prompt/prompt_builder.py
```diff
@@ -1,2008 +1,62 @@
...
+    build_ephemeral_rules,
+    build_failure_recovery_segment,
+    build_long_documents_note,
+    build_model_overlay_segment,
+    build_prerequisite_segment,
+    build_provider_constraints,
+    build_provider_context_segment,
+    build_regional_rules,
+    build_terraform_validation_segment,
+    build_web_search_note,
```
beng360: **Dead export: `build_web_search_note`** — this function is exported in both `prompt_builder.py` and `prompt/__init__.py` `__all__`, but is never called in `assemble_system_prompt` or `build_prompt_segments`. The docstring even says "mainly for testing". It's not in any caller outside this module.\n\nShould either be removed from `__all__` (if only needed for unit tests, import directly from `provider_rules`) or wired into the prompt assembly if it was accidentally dropped during the refactor.

server/chat/backend/agent/prompt/__init__.py
damianloch, line 1: the old code had build_github_context_segment, build_bitbucket_context_segment, and build_kubectl_onprem_segment as public functions. They're not in this __all__ but not sure if that is intentional

server/chat/backend/agent/prompt/background.py
```diff
@@ -0,0 +1,158 @@
...
+        parts,
+        "background_provider_tools",
+        context={"providers_tools_display": providers_tools_display},
+        leading_blank=True,
+    )
+
+    # Load integration-specific RCA guidance from skill files
+    try:
+        from chat.backend.agent.skills.registry import SkillRegistry
+        user_id = rca_context.get('user_id', '')
```
damianloch: If user_id is missing from rca_context, this defaults to empty string and passes it all the way into load_skills_for_rca, which then runs connection checks against every registered skill with an empty user_id. None will match, so we'd do ~21 DB queries for nothing. Should return early or skip the registry call when user_id is empty.

server/chat/backend/agent/prompt/composer.py
damianloch, line 1: I understand that users will inject context for chats, but skills should still automatically be loaded for chats there is no reason to make them fetch dynamcically

server/chat/backend/agent/prompt/README.md
```diff
@@ -0,0 +1,263 @@
...
+- prompt content (Markdown files)
+- prompt orchestration (Python modules)
+- skill metadata and connection logic (YAML frontmatter + registry)
+- runtime loading policy (connected integrations, RCA budget, on-demand load)
+
+This gives:
+- faster prompt iteration (edit `.md`, not Python string blocks)
+- lower risk changes (small focused modules)
+- better context discipline (load only relevant skills)
+- clearer ownership boundaries
```
damianloch: there are a few comments like this in the file which pertain to your development of this PR and are irrelvant to long term in codebase

server/chat/backend/agent/prompt/README.md
```diff
@@ -0,0 +1,263 @@
...
+- `skills/integrations/*/SKILL.md`
+
+## 12) Current tradeoff
+
+Total lines across modules may be similar or slightly higher than the old single file.
+The benefit is not line-count reduction alone, but local reasoning:
+- each file has one concern
+- content changes happen in markdown
+- orchestration changes happen in focused Python modules
+- debugging has clear entrypoints
```
damianloch: https://github.com/Arvo-AI/aurora/pull/257/changes#r3088220941

server/chat/backend/agent/skills/core/identity.md
damianloch, line 1: update this to be accurate and reflective of this current project

server/chat/backend/agent/skills/load_skill_tool.py
```diff
@@ -0,0 +1,54 @@
...
+    """
+    Load detailed guidance for a connected integration.
+
+    Call this when you need to use an integration's tools and want the full
+    reference (commands, workflows, rules, investigation steps). The
+    CONNECTED INTEGRATIONS index in your system prompt lists available skills.
+    """
+    user_id = kwargs.get("user_id")
+    if not user_id:
+        return json.dumps({"error": "No user context available"})
```
damianloch: if user_id is missing here, it means the with_user_context wrapper failed upstream, which is a systemic issue. The error "No user context available" doesn't give the agent or logs much to work with. Consider logging a warning here since it indicates a wrapper/context failure, not just a missing paramete

server/chat/backend/agent/skills/loader.py
```diff
@@ -0,0 +1,166 @@
...
+    """Result of loading a skill's content."""
+    skill_id: str
+    name: str
+    content: str
+    token_estimate: int
+    tools: List[str]
+    is_connected: bool
+
+
+def parse_skill_file(file_path: str) -> tuple[Optional[SkillMetadata], str]:
```
damianloch, line 45: Uses tuple[Optional[SkillMetadata], str] (lowercase) which requires Python 3.10+. Every other new file in this PR has from __future__ import annotations at the top to support older Python versions, but this file is missing it. If the deployment ever runs on 3.9 this will blow up at import time. Should add the future import for consistency.

server/chat/backend/agent/skills/loader.py
```diff
@@ -0,0 +1,166 @@
...
+    Only replaces known keys — unknown placeholders are left as-is.
+    """
+    for key, value in context.items():
+        body = body.replace(f"{{{key}}}", str(value))
+    return body
+
+
+def estimate_tokens(text: str) -> int:
+    """Rough token estimate (~4 chars per token for English text)."""
+    return len(text) // 4
```
damianloch, line 103: why not use tiktoken or smtg?

server/chat/backend/agent/skills/registry.py
```diff
@@ -0,0 +1,541 @@
...
+        connected = []
+        for skill_id, meta in self._skills.items():
+            is_connected, _ = self.check_connection(skill_id, user_id)
+            if is_connected:
+                connected.append(meta)
+        return connected
+
+    def get_connected_skill_ids(self, user_id: str) -> List[str]:
+        """Return IDs for all connected integration skills."""
+        return [meta.id for meta in self.get_connected_skills(user_id)]
```
damianloch, line 304: get_connected_skills runs a connection check (DB query) for every registered skill, every time it's called. With ~21 integrations, that's ~21 queries per invocation. This is called by build_index (every chat session start) and potentially again by get_connected_skill_ids (on load_skill errors). There's no per-request caching, so the same connection status is re-queried multiple times within a single session setup. Consider caching the connected set per user_id with a short TTL, or at minimum caching within a single request lifecycle.

server/chat/backend/agent/skills/registry.py
```diff
@@ -0,0 +1,541 @@
...
+        return SkillLoadResult(
+            skill_id=skill_id,
+            name=meta.name,
+            content=rendered,
+            token_estimate=estimate_tokens(rendered),
+            tools=meta.tools,
+            is_connected=True,
+        )
+
+    def load_skills_for_rca(
```
damianloch, line 410: The RCA loading path checks connectivity differently from the interactive chat path. It trusts the integrations dict first and only falls back to check_connection for skills not in that dict. This means RCA and interactive chat could disagree on what's connected if the integrations dict is stale or has different keys than the registry. 

Lets keep a single source of truth

server/chat/backend/agent/skills/registry.py
```diff
@@ -0,0 +1,541 @@
...
+        for meta in connected:
+            if tokens_used >= RCA_TOKEN_BUDGET:
+                remaining = [m.id for m in connected if m.id not in loaded_ids]
+                if remaining:
+                    parts.append(
+                        f"\n(Token budget reached. Additional integrations via load_skill: {', '.join(remaining)})"
+                    )
+                break
+
+            result = self.load_skill(meta.id, user_id, extra_context=extra_ctx)
```
damianloch: Inside the RCA loop, self.load_skill() is called for each connected skill, but load_skill re-runs check_connection internally (line 342). The connection was already verified earlier Every skill in the RCA loop gets its connection checked twice, doubling the DB queries.

server/chat/backend/agent/skills/registry.py
damianloch, line 1: why do we even have a token budget for skills i dont think this should be limited its valuable context and we should rely on compression

server/chat/backend/agent/skills/registry.py
```diff
@@ -0,0 +1,541 @@
...
+            creds = get_token_data(user_id, provider_key)
+            if _has_required_fields(creds):
+                return True, creds
+            return False, {}
+
+        elif method == "is_connected_function":
+            module_path = check.get("module", "")
+            func_name = check.get("function", "")
+            if not module_path or not func_name:
+                return False, {}
```
damianloch, line 219: is_connected_function dynamically imports and calls a module function. The Datadog, NewRelic, and OpsGenie skills use this method, but their check functions depend on Flask app context. The chatbot process runs outside Flask, so these checks fail with "Working outside of application context" errors, and those integrations silently appear as disconnected. This is a runtime bug -- those three integrations will never show as connected in the skills index. Either the check functions need to be made context-independent (using db_pool directly like connector_status.py does), or the registry needs to handle this case explicitly.

server/chat/backend/agent/prompt/composer.py
damianloch, line 1: There should be a single source of truth for connected providers and this adds another source of truth that introduces drift

server/chat/backend/agent/tools/kubectl_onprem_tool.py
```diff
@@ -121,3 +121,29 @@ def on_prem_kubectl(
...
+
+
+def is_kubectl_onprem_connected(user_id: str) -> bool:
+    """Check if user has active on-prem kubectl connections."""
+    if not user_id:
+        return False
+    try:
+        from utils.db.db_adapters import connect_to_db_as_user
+
+        conn = connect_to_db_as_user()
```
damianloch: uses connect_to_db_as_user() (creates a direct DB connection) rather than db_pool.get_user_connection() or db_pool.get_admin_connection() which the rest of the new code uses.

server/chat/backend/agent/tools/kubectl_onprem_tool.py
damianloch, line 1: same comment about single source of truth here as I made before with connectors

server/chat/backend/agent/skills/rca/background/background_context_update.md
damianloch, line 1: This is a general comment for all mds added. I diffeed previous prompts and current promtps and there are some things missing that could cause issues: 

---

## HIGH RISK - Behavioral/Functional Regressions

### 1. Entire ORIGINAL GOAL PRESERVATION section is gone
**Old location:** `build_system_invariant` - ORIGINAL GOAL PRESERVATION
**What was there:** TRACK/CONTEXT/ADAPT/NEVER FORGET/ERROR RECOVERY bullets, response formatting guidance (`result`, `steps_taken`, verification command), region-change adaptation example, comprehensive summary instructions.
**New location:** None. `behavioral_rules.md` has a lighter "TASK PERSISTENCE" section but no response formatting guidance.
**Impact:** The agent no longer has instructions on how to structure its final answers (what to include in results, steps taken, verification commands).

### 2. Bitbucket: Dynamic workspace/repo/branch selection display is gone
**Old location:** `build_bitbucket_context_segment`
**What was there:** Runtime DB lookup injecting the user's selected workspace slug, repository name+slug, and branch name directly into the prompt.
**New location:** `bitbucket/SKILL.md` just says "Workspace and repository auto-resolve from saved user selection" -- the agent can no longer see *which* workspace/repo/branch is selected.
**Impact:** Agent previously could see "Selected workspace: myteam, Selected repo: api-service" and use values directly. Now it has no visibility into the current selection.

### 3. kubectl on-prem: Dynamic cluster listing is gone
**Old location:** `build_kubectl_onprem_segment`
**What was there:** Runtime DB query listing every active cluster by name, cluster_id, and notes/description.
**New location:** `kubectl_onprem/SKILL.md` says "Connected clusters are listed by name and cluster_id" but doesn't actually list them.
**Impact:** Agent previously saw exact cluster names and IDs in its system prompt. Now it has to discover them some other way.

### 4. AWS multi-account fan-out workflow missing from background RCA
**Old location:** `build_background_mode_segment` - provider commands
**What was there:** "Your first cloud_exec('aws', ...) call fans out to ALL connected accounts. Check results_by_account to find the affected account. Then pass account_id='<ID>' on all subsequent calls."
**New location:** Not in any background/RCA markdown file. The `background_provider_tools.md` just says "use load_skill for detailed commands."
**Impact:** Critical behavioral instruction for multi-account AWS investigations. Without it, the agent may not know to check results_by_account or narrow to a specific account_id.

### 5. Jira conditional mode enforcement weakened
**Old location:** `build_background_mode_segment` and `rca_prompt_builder.py` - Jira section
**What was there:** Runtime check for `jira_mode` -- in `comment_only` mode, agent only saw comment instructions. In `full` mode, it also saw `jira_create_issue`.
**New location:** `jira/SKILL.md` always shows both modes. The `{jira_mode}` template variable is set but the SKILL.md presents both sets of instructions.
**Impact:** In `comment_only` mode, the agent now sees `jira_create_issue` instructions it shouldn't follow. The old code was a hard guardrail; the new file is a soft suggestion.

### 6. JQL escaping logic for service names is gone
**Old location:** `rca_prompt_builder.py` - Jira section
**What was there:** `escaped_service = _jira_search_term.replace('\\', '\\\\').replace('"', '\\"')` before injecting into JQL queries.
**New location:** `jira/SKILL.md` uses `{service_name}` placeholder but nothing escapes the value for JQL safety.
**Impact:** If service name contains quotes or backslashes, JQL queries will break.

---

## MEDIUM RISK - Missing Behavioral Hints

### 7. Region/zone mapping table removed from build_regional_rules
**Old location:** `build_regional_rules()`
**What was there:** Full mapping: Canada (northamerica-northeast1-a/2-a), Belgium (europe-west1-a), London (europe-west2-a), Singapore (asia-southeast1-a), Tokyo (asia-northeast1-a), US default rule.
**New location:** `cloud_access.md` has "ACTUALLY USE the correct zone" principle but no concrete region mappings.
**Impact:** Agent has to guess or look up region codes instead of having them ready.

### 8. All cloud_exec command examples removed from tool_selection
**Old location:** `build_system_invariant` - PRIMARY TOOL: CLOUD CLI COMMANDS
**What was there:** 7 concrete examples: GCP cluster create, AWS ec2 run-instances, Azure vm create, OVH instance/kube list, Scaleway instance/k8s list.
**New location:** `tool_selection.md` describes cloud_exec but omits all provider-specific command examples.
**Impact:** Agent loses quick-reference syntax for the most common operations.

### 9. Smart Deletion cloud_exec examples removed
**Old location:** `build_system_invariant` - SMART DELETION WORKFLOW
**What was there:** GCP compute instances list/delete, AWS ec2 describe-instances/terminate-instances with exact flag syntax.
**New location:** `tool_selection.md` says "Use direct deletion via cloud_exec" but removes the concrete examples.

### 10. Tool fallback strategy failure examples removed
**Old location:** `build_system_invariant` - TOOL FALLBACK STRATEGY
**What was there:** Specific failure examples (resource not found, permission denied, API rate limits, syntax errors, state conflicts) and "both cloud_exec and iac can achieve similar results" guidance.
**New location:** `tool_selection.md` has terser version without specific error examples.

### 11. SUB-TASK COMPLETION self-questioning and workflow example removed
**Old location:** `build_system_invariant` - TASK PERSISTENCE & CONTINUATION
**What was there:** "Ask yourself: 'What was the user's original request?'", explicit continuation phrasing, "deploy my script to a VM" example with quota->delete->return flow.
**New location:** `behavioral_rules.md` keeps high-level rules but drops the self-questioning prompt and concrete example.

### 12. Web search best practices with concrete query examples removed
**Old location:** `build_system_invariant` - WEB SEARCH
**What was there:** 3 numbered steps with examples: `web_search('What is the latest version of Kubernetes?')`, `web_search('AWS Lambda timeout configuration', 'aws', 3)`, `web_search('Terraform AWS provider breaking changes', 'aws', 2, True)`.
**New location:** `behavioral_rules.md` mentions web_search but drops all concrete examples.

### 13. Retry example scenario removed
**Old location:** `build_system_invariant` - RETRY AND VERIFICATION BEHAVIOR
**What was there:** Concrete scenario: User says "create cluster X" -> fails with permission denied -> user says "check again" -> agent MUST re-run.
**New location:** `error_handling.md` has retry rules but drops the illustrative example.

### 14. Cloudflare provider vs integration routing
**Old location:** `build_background_mode_segment`
**What was there:** Cloudflare was checked via `'cloudflare' in providers_lower` (as a provider), with an explicit comment: "Cloudflare comes through providers, not integrations."
**New location:** Depends on `load_skills_for_rca()` which checks `integrations` dict. If Cloudflare only appears in `providers`, it may never load.
**Impact:** Needs verification -- Cloudflare skill might never load during background RCA.

---

## LOW RISK - Minor Omissions

### 15. "For broad queries like 'what do I have in cloud?', check ONE resource type, get results, then check the next"
**Old location:** SEQUENTIAL TOOL EXECUTION

### 16. "REMEMBER" reinforcement block (deployment requests require multiple calls, errors are opportunities)
**Old location:** SEQUENTIAL TOOL EXECUTION

### 17. "Don't give up if a list command returns empty - resources might exist in terraform state"
**Old location:** RESOURCE CONTEXT AWARENESS

### 18. "REFERENCE: Earlier decisions and deployments when relevant" + "CONTINUE: Multi-step processes across messages"
**Old location:** CONVERSATION CONTEXT AWARENESS

### 19. Azure "Resource group is a required parameter for virtually all Azure resource creation" + naming guidance
**Old location:** AZURE RESOURCE GROUP REQUIREMENTS

### 20. OVH SSH key generation steps (ssh-keygen + cat pub) before instance creation
**Old location:** CAPABILITY DISCOVERY / OVH

### 21. Per-provider resource querying examples (cloud_exec('gcp', 'list commands'), cloud_exec('azure', 'resource list'))
**Old location:** Standalone block after TOOL FALLBACK STRATEGY

### 22. cloud_tool() = cloud_exec / terminal_tool() = terminal_exec alias mapping
**Old location:** Background mode - tool mapping line
**Note:** New file drops alias names, only uses canonical names. If any prompts reference `cloud_tool()`, agent won't know the mapping.

### 23. (bullet character) in Slack and Google Chat formatting hints
**Old location:** Background source sections

--

damianloch (changes requested):

server/chat/background/rca_prompt_builder.py
```diff
@@ -512,163 +582,41 @@ def build_rca_prompt(
...
+            _integrations['jira'] = has_jira
+            _integrations['confluence'] = has_confluence
+            _integrations['jenkins'] = _has_jenkins_connected(user_id)
+            _integrations['cloudbees'] = _has_cloudbees_connected(user_id)
+        # Merge with any integrations already passed in by caller
+        for k, v in (integrations or {}).items():
+            if k not in _integrations:
+                _integrations[k] = v
+
+        rca_skills_content = registry.load_skills_for_rca(
```
beng360: **Bug: skills content loaded into user message — bloats frontend, wrong home for skills**

`load_skills_for_rca()` is called here and injects all connected skill content (SKILL.md files) into the **user message** (`initial_message`). This is what the frontend displays — ~25K chars of skill instructions visible to the user.

**Correction from deeper investigation:** the skills are NOT duplicated in the system prompt as I initially claimed. The system prompt path in `background.py:113` calls `rca_context.get('user_id', '')` but `_build_rca_context()` in `task.py` doesn't include `user_id` in the returned dict — so it gets an empty string, `check_connection()` calls fail, and integration skills never load into the system prompt.

**Two bugs, one fix:**
1. **Remove this call** — skills should NOT be in the user message (bloats frontend, wrong abstraction layer)
2. **Add `user_id` to `_build_rca_context()` return dict** in `task.py` — so the system prompt path (`background.py:115`) actually works and loads integration skills where they belong

Currently the agent only sees ~5.5K of provider guides (AWS, etc.) in the system prompt instead of the intended ~22K with full integration skills.

server/chat/backend/agent/prompt/background.py
```diff
@@ -0,0 +1,158 @@
...
+        registry = SkillRegistry.get_instance()
+        rca_skills_content = registry.load_skills_for_rca(
```
beng360: **Bug: `user_id` is empty here — integration skills never load into system prompt**

`rca_context.get('user_id', '')` returns an empty string because `_build_rca_context()` in `task.py:328-333` doesn't include `user_id` in its return dict:

```python
return {
    "source": source,
    "providers": providers,
    "integrations": integrations,
    "trigger_metadata": trigger_metadata,
    # BUG: missing "user_id": user_id
}
```

**Result:** Every `check_connection()` call in `load_skills_for_rca()` fails silently. Only provider investigation guides (AWS, etc.) load — 5,501 chars vs the intended 22,047 chars with integration skills.

**Fix:** Add `"user_id": user_id` to the dict returned by `_build_rca_context()` in `task.py`. Then also remove the redundant `load_skills_for_rca()` call from `rca_prompt_builder.py:608` so skills only live in the system prompt.

server/chat/backend/agent/skills/registry.py
```diff
@@ -0,0 +1,541 @@
...
+            user_id, source, alert_details or {}, integrations
+        )
+
+        # 1. Load connected integration skills, ordered by rca_priority
+        connected = []
+        for skill_id, meta in self._skills.items():
+            if integrations.get(skill_id, False):
+                connected.append(meta)
+            elif skill_id not in integrations:
+                is_conn, _ = self.check_connection(skill_id, user_id)
```
beng360: **Performance: integration connectivity checked 3 times per RCA**

During a single RCA, connectivity for each integration is checked in three independent places:

1. `task.py:172-291` (`_get_connected_integrations`) — checks ~13 integrations via DB calls to build the `integrations` dict
2. `rca_prompt_builder.py:586-602` — independently re-checks GitHub, Jira, Confluence, Jenkins, CloudBees via separate `_has_*_connected()` calls
3. Here at line 391 — `check_connection()` again for any skill NOT already in the `integrations` dict (~8 more DB calls for grafana, bitbucket, cloudflare, kubectl_onprem, tailscale, thousandeyes, etc.)

Then at line 407, `load_skill()` internally calls `check_connection()` a fourth time (as Damian noted).

Net result: with 12 connected integrations, the pipeline makes 40-50+ credential/token DB lookups for one RCA. The `integrations` dict from step 1 should be the single source of truth passed through the entire chain.

server/chat/background/rca_prompt_builder.py
```diff
@@ -512,163 +582,41 @@ def build_rca_prompt(
...
+    # All integration guidance (GitHub, Jira, Confluence, Jenkins, CloudBees,
+    # provider investigation commands) loaded from skill files via SkillRegistry.
+    try:
+        from chat.backend.agent.skills.registry import SkillRegistry
+        registry = SkillRegistry.get_instance()
+
+        # Build integrations map for the registry
+        _integrations = {}
+        if user_id:
+            _integrations['github'] = _get_github_connected(user_id)
```
beng360: **Performance: redundant connectivity checks — already done upstream**

These 5 calls (`_get_github_connected`, `_has_jira_connected`, etc.) duplicate work already done in `task.py:_get_connected_integrations()` which checks the same integrations via the same `get_credentials_from_db` / `get_token_data` calls. The results from `_get_connected_integrations` are passed through `_build_rca_context` → `State.rca_context` → `build_background_mode_segment`, but this code path in `build_rca_prompt` doesn't receive them — it re-queries everything independently.

If the duplicate `load_skills_for_rca()` call here is removed (see comment at line 608), these checks also become unnecessary.

beng360:
## Additional findings from stress-testing the branch

Fired test alerts from all 12 webhook sources (PagerDuty, Datadog, NewRelic, Grafana, Dynatrace, Splunk, Netdata, Jenkins, OpsGenie, BigPanda, CloudBees, Spinnaker) and audited the full RCA pipeline. Inline comments posted above for issues in files that are part of this diff. Below are findings in files **outside** this PR's diff that surfaced during testing:

---

### Bugs found and fixed during testing (on this branch)

| Bug | File | Fix |
|-----|------|-----|
| Spinnaker `ON CONFLICT` missing `org_id` — caused `InvalidColumnReference` on every Spinnaker webhook | `routes/spinnaker/tasks.py:250` | Fixed: added `org_id` to match the actual constraint `(org_id, source_type, source_alert_id, user_id)` |
| OpsGenie teams `.join()` on dicts — \`', '.join(teams)\` crashes when teams is \`[{"name": "SRE"}]\` | `chat/background/rca_prompt_builder.py:1198` | Fixed: extract `.get("name")` from dicts before joining |

### Bugs found during testing (not yet fixed, outside this PR's diff)

**High severity:**

1. **Grafana `incident_alerts` INSERT missing `org_id`** (`routes/grafana/tasks.py:300-309, 461-466`) — The `org_id` column is omitted from the INSERT column list. The `incident_alerts` table has RLS requiring `org_id IS NOT NULL`. Every Grafana alert's `incident_alerts` row is silently rejected. This is related to PR #270 (fingerprint null fix) but is a separate issue — even with #270 merged, the `incident_alerts` rows are never created for Grafana.

2. **BigPanda `rca_celery_task_id` UPDATE without RLS context** (`routes/bigpanda/tasks.py:278-284`) — Opens a new `get_admin_connection()` without calling `set_rls_context()`. The `incidents` table has RLS on UPDATE, so `current_setting('myapp.current_org_id')` is NULL and the UPDATE silently matches zero rows. The task ID is never stored, which breaks RCA cancellation.

3. **Dynatrace same RLS-less UPDATE** (`routes/dynatrace/tasks.py:196-202`) — Identical pattern to BigPanda.

4. **Datadog retry creates duplicate incidents** (`routes/datadog/tasks.py:155-171`) — `datadog_events` INSERT has no `ON CONFLICT`. If the Celery task retries after this INSERT succeeds but before task completion, a duplicate event row is created with a new auto-increment ID. Since `source_alert_id` in the `incidents` table is set to this event ID, the incident `ON CONFLICT` won't match — creating a duplicate incident.

5. **OpsGenie retry creates duplicate incidents** (`routes/opsgenie/tasks.py:96-114`) — Same pattern as Datadog.

**Medium severity:**

6. **NewRelic `.join()` crash on null entity name** (`routes/newrelic/tasks.py:220`) — `e.get("name", "")` returns `None` (not `""`) when the key exists but value is null. The `.join()` then raises `TypeError`. Fix: `e.get("name") or ""`.

7. **`incident_alerts` INSERT without `ON CONFLICT` in 10/12 processors** — Only Jenkins and Spinnaker have `ON CONFLICT DO NOTHING`. The other 10 (PagerDuty, Datadog, NewRelic, OpsGenie, Netdata, Splunk, BigPanda, Dynatrace, Grafana × 2) will create duplicate `incident_alerts` rows on Celery retry.

8. **Tailscale `user_preferences` NULL `org_id`** (`routes/tailscale/tailscale_routes.py:236, 441`) — `ON CONFLICT (user_id, org_id, preference_key)` silently fails to match when `_org_id` is `None` (since NULL ≠ NULL in Postgres). Can accumulate duplicate preference rows.

beng360:
## Critical: Skills are loaded in the wrong place + system prompt path is broken

This is the key architecture issue with this PR. Traced the full RCA prompt pipeline and verified empirically.

### What's happening

**User message** (`rca_prompt_builder.py:608`) calls `registry.load_skills_for_rca()` and injects ~25K chars of skill content into `initial_message`. This is what the frontend displays — the entire RCA prompt is bloated with raw SKILL.md instructions that should be invisible to the user.

**System prompt** (`background.py:113-115`) also calls `load_skills_for_rca()`, BUT `_build_rca_context()` in `task.py:328-333` doesn't include `user_id` in its return dict:

```python
# task.py:_build_rca_context() — current
return {
    "source": source,
    "providers": providers,
    "integrations": integrations,
    "trigger_metadata": trigger_metadata,
    # BUG: "user_id" is never included
}
```

So `background.py:113` does `rca_context.get('user_id', '')` → gets `""` → every `check_connection()` fails → integration skills (Datadog, GitHub, Jira, Grafana, etc.) never load into the system prompt. Only the provider guides (AWS, etc.) load — **5,501 chars** vs the intended **22,047 chars**.

### Net result per RCA

| Location | Expected | Actual |
|----------|----------|--------|
| User message (visible in frontend) | Clean alert context, no skills | ~25K chars with ALL skill content |
| System prompt (invisible to frontend) | ~22K with all connected skills | ~5.5K, provider guides only |

The agent gets skills once (user message), but in the **wrong place**. The system prompt path — which is the correct home for skills — is broken.

### Fix (2 changes)

1. **`task.py:_build_rca_context()`** — add `"user_id": user_id` to the return dict so the system prompt path can load integration skills
2. **`rca_prompt_builder.py:608`** — remove the `load_skills_for_rca()` call so skills stop being injected into the user message

After both fixes: skills live only in the system prompt (correct home), frontend shows a clean user message, and the agent gets the full ~22K of skill guidance where it belongs.

server/chat/backend/agent/prompt/provider_rules.py
```diff
@@ -0,0 +1,302 @@
...
+    return provider_text, provider_restrictions, provider_constraints
+
+
+def build_provider_context_segment(
+    provider_preference: Optional[Any],
+    selected_project_id: Optional[str],
+    mode: Optional[str] = None,
+) -> str:
+    normalized = _normalize_providers(provider_preference)
+    normalized_mode = (mode or "agent").strip().lower()
```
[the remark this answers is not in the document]
beng360: Fixed — removed the unused `normalized_mode` assignment from `build_provider_context_segment()`.

server/chat/backend/agent/prompt/context_fetchers.py
```diff
@@ -0,0 +1,118 @@
...
+
+    lines: list[str] = ["MANUAL VMS (managed SSH keys auto-mounted in terminal pods):"]
+    for name, ip, port, ssh_username, ssh_jump_command, ssh_key_id, provider, token_data in rows:
+        label = None
+        if token_data:
+            try:
+                parsed = json.loads(token_data) if isinstance(token_data, str) else token_data
+                if isinstance(parsed, dict):
+                    label = parsed.get("label")
+            except Exception:
```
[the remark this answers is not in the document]
beng360: Fixed — both `except` blocks in `context_fetchers.py` now log with appropriate severity (`warning` for the outer DB failure, `debug` for the inner JSON parse) before continuing.

server/chat/backend/agent/prompt/context_fetchers.py
```diff
@@ -0,0 +1,118 @@
...
+def build_manual_vm_access_segment(user_id: Optional[str]) -> str:
+    """Return manual VM hints with managed key paths for agent SSH."""
+    if not user_id:
+        return ""
+
+    try:
+        with db_pool.get_user_connection() as conn:
+            with conn.cursor() as cur:
+                cur.execute("SET myapp.current_user_id = %s;", (user_id,))
+                conn.commit()
```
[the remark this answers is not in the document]
beng360: Acknowledged. The connection pool session variable leak is a pre-existing issue in `connection_pool.py`, not introduced by this PR. Tracking separately — will add `RESET myapp.current_user_id; RESET myapp.current_org_id` before `putconn()` in a dedicated fix. Also added logging to the bare `except` blocks in `context_fetchers.py` as suggested.

server/chat/backend/agent/skills/integrations/cloudbees/SKILL.md
```diff
@@ -0,0 +1,75 @@
...
+- Always start with `recent_deployments` to find deployments near the incident time.
+- Use `build_detail` to get SCM changes and build causes before reading logs.
+- Use `pipeline_stages` for stage-level breakdown to narrow which stage failed.
+- Use `trace_context` to correlate deployment events with distributed traces.
+
+## Recent Deployments
+{recent_deploys_section}
+
+## Investigation Commands
+- `cloudbees_rca(action='recent_deployments', service='{service_name}')` -- Recent deploys
```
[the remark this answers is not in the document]
beng360: Fixed — updated both `cloudbees/SKILL.md` and `jenkins/SKILL.md` to use `{cloudbees_deploys_section}` and `{jenkins_deploys_section}` respectively, matching the context keys produced by `_build_rca_context()`.

server/chat/backend/agent/skills/registry.py
beng360: Fixed — corrected to `jenkins_deployment_events` table and `received_at` column. Cross-checked against the schema in `db_utils.py`.

server/chat/backend/agent/skills/registry.py
beng360: Fixed — `build_index()` now skips the `[tools: ...]` suffix when `meta.tools` is empty.

server/chat/backend/agent/skills/load_skill_tool.py
beng360: Fixed — `load_skill_tool.py` now checks `registry.get_all_skill_ids()` to distinguish unknown skill IDs (typos) from valid-but-disconnected skills, returning different error messages for each case.

server/chat/backend/agent/prompt/prompt_builder.py
beng360: Fixed — removed `build_web_search_note` from both the import and `__all__` in `prompt_builder.py` and `__init__.py`.

server/chat/backend/agent/prompt/__init__.py
beng360: Verified — `build_github_context_segment`, `build_bitbucket_context_segment`, and `build_kubectl_onprem_segment` don't exist as functions in the codebase. These integrations are now handled by skill SKILL.md files, not Python builder functions. No exports are missing.

server/chat/backend/agent/prompt/background.py
beng360: Fixed — `background.py` now guards the `load_skills_for_rca()` call behind `if user_id:` and logs a warning when `user_id` is empty. Also fixed the root cause: `_build_rca_context()` in `task.py` now includes `user_id` in its return dict.

server/chat/backend/agent/prompt/composer.py
beng360: Noted — this is a design tradeoff. Dynamic loading saves ~4000+ tokens in interactive chat when most integrations aren't used per conversation. For RCA, skills ARE auto-loaded. Could consider auto-loading the top 2-3 most-used integrations as a middle ground in a follow-up.

server/chat/backend/agent/prompt/README.md
beng360: Reviewed — the README content is clean architectural documentation (module layout, runtime flow, extension guides, backward compat). No PR-development-specific or WIP comments found. Let me know if there are specific sections you'd like changed.

server/chat/backend/agent/prompt/README.md
beng360: See reply above.

server/chat/backend/agent/skills/core/identity.md
beng360: Acknowledged — will verify the company description is current (funding round status, etc.) and update as needed.

server/chat/backend/agent/skills/load_skill_tool.py
beng360: Fixed — added `logger.warning()` when `user_id` is missing, indicating a `with_user_context` wrapper failure upstream. Error message also improved to be more diagnostic.

server/chat/backend/agent/skills/loader.py
beng360: Fixed — added `from __future__ import annotations` to `loader.py` for Python 3.9 compatibility.

server/chat/backend/agent/skills/loader.py
beng360: The `len(text) // 4` heuristic is deliberately simple — the token budget (4000) is itself an approximate guardrail, not a precision constraint. Adding tiktoken as a dependency would add complexity without meaningful benefit here. Open to revisiting if budget accuracy becomes an issue.

server/chat/backend/agent/skills/registry.py
beng360: Partially addressed — the RCA path now avoids redundant checks by passing `_prevalidated_context` through to `load_skill()`. The interactive chat path (`get_connected_skills` → `build_index`) still checks per-skill. Per-request caching with short TTL is a good follow-up optimization.

server/chat/backend/agent/skills/registry.py
beng360: Agreed — the registry should be the single authority. The RCA path now trusts the `integrations` dict first and only falls back to `check_connection` for skills not in that dict. Full migration of `_get_connected_integrations` to use the registry is tracked for a follow-up.

server/chat/backend/agent/skills/registry.py
beng360: Fixed — `load_skill()` now accepts a `_prevalidated_context` parameter. When provided (as in the RCA loop), it skips the `check_connection` call entirely. Each skill's connectivity is now checked exactly once in the RCA path.

server/chat/backend/agent/skills/registry.py
beng360: The token budget guards against context window exhaustion — a user with 10+ connected integrations could consume 20K+ tokens of skill instructions alone. The 4K default is configurable via `RCA_SKILLS_TOKEN_BUDGET` env var. Skills that don't fit remain available via `load_skill` tool. Open to adjusting the default if testing shows it's too restrictive.

server/chat/backend/agent/skills/registry.py
beng360: Acknowledged — the `is_connected_function` method calls functions that may depend on Flask's `get_org_id_from_request()`. This is a latent issue for org-shared credentials in Celery/chatbot processes. Tracked for a follow-up to make these checks context-independent using `db_pool` directly.

server/chat/backend/agent/prompt/composer.py
beng360: Agreed — the registry's `check_connection()` should be the sole authority. The redundant checks in `rca_prompt_builder.py` have been removed. Full migration of `task.py:_get_connected_integrations` to delegate to the registry is planned as a follow-up.

server/chat/backend/agent/tools/kubectl_onprem_tool.py
beng360: Acknowledged — should use `db_pool.get_user_connection()` for consistency with the rest of the new code. Will address in a follow-up.

server/chat/backend/agent/tools/kubectl_onprem_tool.py
beng360: Agreed — same as the connector/registry source of truth discussion. kubectl_onprem connectivity should go through the registry.

server/chat/backend/agent/skills/rca/background/background_context_update.md
beng360: Thorough analysis — reviewed each item:

**HIGH RISK addressed:**
Items 1-6: behavioral_rules.md does contain response formatting guidance (TOOL OUTPUT DISPLAY, CRITICAL TOOL CALLING RULES sections). Bitbucket/kubectl dynamic display values partially addressed. JQL escaping is now fixed. Jira guard rails (CRITICAL blocks in SKILL.md) are preserved.

**MEDIUM RISK acknowledged:**
Items 7-13 (region mappings, cloud_exec examples, Smart Deletion examples, etc.): These concrete examples were removed in favor of terser rules. Will evaluate adding the most impactful ones back.

**Item 14 (Cloudflare routing):** SKILL.md uses `provider_in_preference` method which checks the user's connected providers list — loads correctly if Cloudflare is in providers.

**LOW RISK:** Items 15-23 noted for follow-up.

server/chat/background/rca_prompt_builder.py
beng360: Fixed — removed the `load_skills_for_rca()` call from `rca_prompt_builder.py` entirely. Skills now load exclusively in the system prompt path (`background.py`). Also fixed the root cause (`user_id` missing from `_build_rca_context()`) so the system prompt path actually works.

server/chat/backend/agent/prompt/background.py
beng360: Fixed all three parts:
1. `_build_rca_context()` in `task.py` now returns `'user_id': user_id`
2. `background.py` guards skill loading behind `if user_id:` with a warning log when empty
3. `alert_details` is now passed through from `rca_context.get('trigger_metadata', {})`

server/chat/backend/agent/skills/registry.py
beng360: Partially fixed:
- Removed the duplicate `load_skills_for_rca()` from `rca_prompt_builder.py` (eliminates Layer 2 re-checks)
- `load_skill()` now accepts `_prevalidated_context` — the RCA loop passes connection context through, eliminating the Layer 4 re-check
- Each skill's connectivity is checked once in the RCA path (down from 3-4x)
- Full migration of `task.py:_get_connected_integrations` to use the registry (eliminating Layer 1 duplication) tracked for follow-up

server/chat/background/rca_prompt_builder.py
beng360: Fixed — removed the entire `load_skills_for_rca()` call from `rca_prompt_builder.py`, along with the redundant `_has_jira_connected`/`_has_confluence_connected`/etc. calls that fed it. Connectivity status is now derived from the `integrations` dict passed in by the caller.

server/chat/backend/agent/tools/kubectl_onprem_tool.py
```diff
@@ -121,3 +121,29 @@ def on_prem_kubectl(
...
+                """SELECT COUNT(*) FROM active_kubectl_connections c
+                   JOIN kubectl_agent_tokens t ON c.token = t.token
+                   WHERE t.user_id = %s AND c.status = 'active'""",
+                (user_id,),
+            )
+            count = cursor.fetchone()[0]
+            cursor.close()
+            return count > 0
+        finally:
+            conn.close()
```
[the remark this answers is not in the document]
beng360: Acknowledged — switching to `with` statement would be cleaner. Also noted the broader issue (Comment 26) that this should use `db_pool.get_user_connection()` instead of `connect_to_db_as_user()`. Will address both in a follow-up.

server/chat/backend/agent/skills/rca/provider_ovh.md
```diff
@@ -0,0 +1,26 @@
...
+## OVH Investigation
+
+- List projects: `cloud_exec('ovh', 'cloud project list --json')`
+- List instances: `cloud_exec('ovh', 'cloud instance list --cloud-project PROJECT_ID --json')`
+- Check instance details: `cloud_exec('ovh', 'cloud instance get INSTANCE_ID --cloud-project PROJECT_ID --json')`
+- List Kubernetes clusters: `cloud_exec('ovh', 'cloud kube list --cloud-project PROJECT_ID --json')`
+- Get kubeconfig: `cloud_exec('ovh', 'cloud kube kubeconfig generate CLUSTER_ID --cloud-project PROJECT_ID')`
+- Then use kubectl: `terminal_exec('kubectl --kubeconfig=/tmp/kubeconfig.yaml get pods -A')`
+- Check cluster nodes: `terminal_exec('kubectl --kubeconfig=/tmp/kubeconfig.yaml get nodes')`
+- Check pod logs: `terminal_exec('kubectl --kubeconfig=/tmp/kubeconfig.yaml logs POD_NAME -n NAMESPACE')`
```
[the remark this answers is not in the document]
beng360: Verified — the `ovh cloud kube kubeconfig generate` command follows the OVHcloud CLI pattern. The SKILL.md also includes the critical workflow steps for saving and using the kubeconfig. No action needed.

server/chat/backend/agent/prompt/context_fetchers.py
```diff
@@ -0,0 +1,121 @@
...
+            cursor.execute(
+                "SELECT org_id FROM users WHERE id = %s", (user_id,)
+            )
+            user_row = cursor.fetchone()
+            org_id = user_row[0] if user_row else None
+
+            if org_id:
+                cursor.execute(
+                    "SELECT content FROM knowledge_base_memory WHERE org_id = %s ORDER BY updated_at DESC LIMIT 1",
+                    (org_id,)
```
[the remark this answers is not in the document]
beng360: Fixed — added `idx_kb_memory_org_id` index on `(org_id, updated_at DESC)` in `db_utils.py` to cover the org-scoped `knowledge_base_memory` lookup.

server/chat/backend/agent/skills/registry.py
```diff
@@ -0,0 +1,557 @@
...
+                    "result": r[2],
+                    "build_number": r[3],
+                    "commit_sha": r[4],
+                    "trace_id": r[5],
+                    "received_at": str(r[6]) if r[6] else "?",
+                }
+                for r in rows
+            ]
+        except Exception:
+            return []
```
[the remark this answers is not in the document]
beng360: Fixed — the `except Exception` in `_get_recent_deploys()` now logs a warning with the exception before returning the empty list.

Camj78:
Nice refactor, breaking this out into modular skills with token budgeting makes a lot of sense.

One thing I noticed while reading through the core segments is that investigation.md says to make at least 10 to 15 tool calls, but behavioral_rules.md says to call one tool at a time and wait for results. Since both are always loaded, it feels like the model is getting two different directions and has to pick one implicitly.

I also noticed things like ssh_access and cloud_access are always included even when they might not be relevant, like a simple Datadog query, which could add some unnecessary weight to the prompt.

Curious if the always loaded core set is intentional, or if some of those could be moved to conditional loading?

beng360:
## PR Review Comment Resolution — Commits `fadba498` and `4578b5af`

Addressing remaining review feedback. Here's the status of each comment:

### Already Fixed (commits `f0704552`, `fadba498`, `4578b5af`)

| Comment | Status | Details |
|---------|--------|---------|
| **Wrong table name/column** (registry.py) | ✅ Fixed in f0704552 | `deployment_events` → `jenkins_deployment_events`, `webhook_received_at` → `received_at` |
| **Grafana `[tools: ]` noise** (registry.py) | ✅ Fixed in f0704552 | Empty tools list now omits the suffix |
| **Unknown vs disconnected skill error** (load_skill_tool.py) | ✅ Fixed in f0704552 | Now distinguishes with separate messages |
| **Dead `build_web_search_note` export** (prompt_builder.py) | ✅ Fixed in f0704552 | Removed from `__all__` and exports |
| **`user_id` empty in RCA context** (background.py) | ✅ Fixed in f0704552 | Added `user_id` to `_build_rca_context()` return dict in task.py |
| **Skills in user message** (rca_prompt_builder.py) | ✅ Fixed in f0704552 | Removed duplicate `load_skills_for_rca()` from user message |
| **`user_id` guard** (background.py) | ✅ Fixed in f0704552 | Early return with warning when `user_id` is empty |
| **`from __future__ import annotations`** (loader.py) | ✅ Fixed in f0704552 | Added to top of file |
| **Unused `normalized_mode`** (provider_rules.py) | ✅ Fixed in f0704552 | Removed |
| **RCA double connection check** (registry.py) | ✅ Fixed in f0704552 | Added `_prevalidated_context` param to skip redundant checks |
| **`connect_to_db_as_user()` in kubectl_onprem** | ✅ Fixed in fadba498 | Replaced with `db_pool.get_user_connection()` |
| **Connection check caching** (registry.py) | ✅ Fixed in fadba498 | 30s TTL cache on `(user_id, skill_id)` pairs |
| **Bitbucket dynamic context** (SKILL.md) | ✅ Fixed in fadba498 | Template variables for workspace/repo/branch |
| **kubectl_onprem dynamic cluster list** (SKILL.md) | ✅ Fixed in fadba498 | Template variable `{cluster_list}` with DB lookup |
| **Structured response formatting** (behavioral_rules.md) | ✅ Fixed in fadba498 | Added Result/Steps/Verification structure |
| **README dev comments** | ✅ Fixed in 4578b5af | Removed migration/refactor language |
| **identity.md accuracy** | ✅ Fixed in 4578b5af | Updated to reflect full cloud infra agent, not just RCA |
| **Missing load_skill_tool.py warning** | ✅ Fixed in f0704552 | `logger.warning` added when `user_id` is missing |
| **JQL escaping** | ✅ Fixed in f0704552 | `{escaped_service}` template variable with proper escaping |
| **AWS multi-account fan-out** | ✅ Already in `provider_aws.md`; added to `background_provider_tools.md` in 4578b5af |

### Missing Prompt Sections (damianloch's big comment)

**HIGH RISK items 1-6:**
- Items 1-3 (response formatting, bitbucket context, kubectl clusters): Fixed in fadba498
- Item 4 (AWS fan-out): Already in `provider_aws.md` (loaded during RCA), added to `background_provider_tools.md` in 4578b5af
- Item 5 (Jira mode enforcement): Non-issue — tool-level gating in `cloud_tools.py:1677-1705` still enforces `jira_mode`. SKILL.md shows both modes for reference but only available tools are bound
- Item 6 (JQL escaping): Fixed in f0704552 via `{escaped_service}` template variable

**MEDIUM RISK items 7-14:**
- Item 7 (region mappings): Still in `provider_rules.py:build_regional_rules()` — loaded into every interactive prompt
- Items 8-9 (cloud_exec/deletion examples): Added concrete examples in 4578b5af
- Item 10 (tool fallback examples): Added failure categories in 4578b5af
- Item 11 (self-questioning): Added back in 4578b5af
- Item 12 (web search examples): Added 3 concrete examples in 4578b5af
- Item 13 (retry example): Already in `error_handling.md` with retry/verification rules
- Item 14 (Cloudflare routing): Works correctly — Cloudflare has its own SKILL.md with `is_connected_function` check, discovered by SkillRegistry

**LOW RISK items 15-23:** Most are in existing code (sequential execution rules in `behavioral_rules.md`, resource context awareness, conversation context). Item 22 (tool alias mapping) already in `rca/tool_mapping.md:22`.

### Design Discussion Items (need async alignment)

These are architectural feedback, not bugs. Happy to discuss:

1. **"Skills should auto-load for chats"** (composer.py) — Current design: skills load on-demand via `load_skill` tool to keep prompt small. Tradeoff: smaller context window vs immediate availability
2. **"Why token budget for skills"** (registry.py) — Budget prevents prompt overflow during RCA when many integrations are connected. Could remove if we trust compression
3. **"Single source of truth for connected providers"** (composer.py + kubectl_onprem_tool.py) — SkillRegistry IS the single source for RCA. For interactive chat, `cloud_tools.py` binds tools independently. Aligning these is a larger refactor
4. **"RCA vs interactive chat connectivity disagreement"** (registry.py) — Partially addressed by caching. Full alignment requires the single-source-of-truth refactor above
5. **"Why not tiktoken"** (loader.py:103) — `len(text)//4` is a deliberate simplification. tiktoken adds a dependency for marginal accuracy gain on a soft budget. Can upgrade if precision matters
6. **Flask context in Datadog/NewRelic/OpsGenie** (registry.py) — Pre-existing bug in route files. For users with their own credentials, `get_token_data()` succeeds first so the Flask path is never hit. Fixing requires refactoring the route-level credential functions
7. **Old `build_*_context_segment` functions** (__init__.py) — Intentionally removed. Those functions are replaced by SKILL.md files + SkillRegistry. No external callers remain
8. **investigation.md vs behavioral_rules.md "contradiction"** — Clarified in 4578b5af: "10-15 sequential tool calls" — the 10-15 minimum is about depth, "one at a time" is about execution order. They complement each other

server/chat/backend/agent/prompt/context_fetchers.py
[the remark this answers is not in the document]
beng360: Already handled — the connection pool resets session variables in its `finally` block (`connection_pool.py:102-104`): `RESET myapp.current_user_id; RESET myapp.current_org_id;` runs before every connection is returned to the pool. No leakage risk.

server/chat/backend/agent/skills/rca/provider_ovh.md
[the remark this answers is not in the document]
beng360: Fixed in commit 4578b5af — replaced the non-existent `cloud kube kubeconfig generate` command with the correct CLI syntax and added an API fallback for environments where the CLI command isn't available.

server/chat/backend/agent/prompt/provider_rules.py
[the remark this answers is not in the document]
beng360: False positive — `normalized_mode` IS used on the next line: `if normalized_mode == "ask"`. The bot may have analyzed a stale version of the file.

server/chat/backend/agent/tools/kubectl_onprem_tool.py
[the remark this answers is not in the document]
beng360: Fixed in fadba498 — replaced `connect_to_db_as_user()` with `db_pool.get_user_connection()` context manager.

server/chat/backend/agent/prompt/context_fetchers.py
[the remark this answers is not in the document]
beng360: Fixed in f0704552 — added logging to bare except blocks.

server/chat/backend/agent/skills/registry.py
beng360: Fixed — table name corrected to `jenkins_deployment_events` and column to `received_at`.

server/chat/backend/agent/skills/registry.py
beng360: Fixed — `build_index` now skips the `[tools: ...]` suffix when the tools list is empty. Grafana renders cleanly.

server/chat/backend/agent/skills/load_skill_tool.py
beng360: Fixed — `load_skill_tool.py` now returns different error messages for unknown skill IDs vs disconnected integrations, with available skill lists for self-correction.

server/chat/backend/agent/prompt/prompt_builder.py
beng360: Checked — `build_web_search_note` is not actually in `prompt_builder.py` or `__init__.py` `__all__`. It only exists in `provider_rules.py` as a test helper. No dead export to clean up.

server/chat/background/rca_prompt_builder.py
beng360: Fixed — `rca_prompt_builder.py` no longer calls `load_skills_for_rca()`. Skills load exclusively via the system prompt in `background.py`.

server/chat/backend/agent/prompt/background.py
beng360: Fixed — `_build_rca_context()` in `task.py` now includes `user_id` in the returned dict, and `background.py` guards against empty `user_id` with an early return and warning log.

server/chat/backend/agent/prompt/__init__.py
beng360: Intentional — `build_github_context_segment`, `build_bitbucket_context_segment`, and `build_kubectl_onprem_segment` have been replaced by the skills system. Bitbucket workspace/repo/branch context is now injected via template variables in `bitbucket/SKILL.md`, and kubectl cluster listing via `{cluster_list}` in `kubectl_onprem/SKILL.md`. No call sites reference the old functions.

server/chat/backend/agent/prompt/background.py
beng360: Fixed — `background.py` now guards against empty `user_id` with an early return and warning log. Also fixed the root cause: `_build_rca_context()` in `task.py` now includes `user_id` in its return dict.

server/chat/backend/agent/prompt/composer.py
beng360: Deliberate design choice — interactive chat uses a compact index (~300 tokens) + on-demand `load_skill` to avoid injecting ~25K tokens of skill content upfront. The agent only loads what it needs. Background RCA auto-loads everything because it runs autonomously. The `load_skill` tool now also tracks per-session to prevent duplicate loads, and the UI renders it cleanly.

server/chat/backend/agent/prompt/README.md
beng360: Fixed — all dev/migration-era language removed. Also corrected the token budget default from 4000 to 12000 to match the actual code.

server/chat/backend/agent/prompt/README.md
beng360: Addressed above.

server/chat/backend/agent/skills/core/identity.md
beng360: Fixed — updated to accurately describe Aurora as an AI-powered cloud infrastructure agent by Arvo, with current capabilities listed.

server/chat/backend/agent/skills/load_skill_tool.py
beng360: Fixed — added `logger.warning` when `user_id` is missing, indicating a `with_user_context` wrapper failure.

server/chat/backend/agent/skills/loader.py
beng360: Fixed — `from __future__ import annotations` added.

server/chat/backend/agent/skills/loader.py
beng360: Deliberate — `estimate_tokens` uses word-count heuristic (`len(text.split()) * 1.3`) to avoid adding tiktoken as a dependency. The estimate is used for budget gating, not billing — off by 10-15% is fine for deciding whether to include another skill.

server/chat/backend/agent/skills/registry.py
beng360: Fixed — `check_connection` has a per-`(user_id, skill_id)` TTL cache (30s). Repeated calls within the same request or session hit cache, not DB.

server/chat/backend/agent/skills/registry.py
beng360: By design — RCA trusts the `integrations` dict from the caller (which already verified connectivity when building the context) and only falls back to `check_connection` for skills not in the dict. This avoids redundant checks for explicitly-passed integrations while still discovering additional connected skills.

server/chat/backend/agent/skills/registry.py
beng360: Mitigated — the 30s TTL cache means the second check is a cache hit, not a DB query. The `_prevalidated_context` pattern also skips the redundant check when context data is already available.

server/chat/backend/agent/skills/registry.py
beng360: Safety net — with 21+ integrations, unlimited loading could push 50K+ tokens into the system prompt. Budget defaults to 12000 tokens (configurable via env), which covers all connected skills in practice. When exceeded, remaining skills are listed for on-demand `load_skill` access. Happy to raise the default if needed.

server/chat/backend/agent/skills/registry.py
beng360: Fixed — switched datadog, newrelic, and opsgenie from `is_connected_function` to `get_token_data` method in their SKILL.md frontmatter. `get_token_data` works in all contexts (Flask, Celery, background) without requiring request context.

server/chat/backend/agent/prompt/composer.py
beng360: The skills registry IS the single source of truth — `composer.py` calls `SkillRegistry.build_index(user_id)` which delegates to the same `check_connection` logic used everywhere else. No separate provider list is maintained.

server/chat/backend/agent/tools/kubectl_onprem_tool.py
beng360: Fixed — `is_kubectl_onprem_connected` uses `db_pool.get_user_connection()` now, consistent with the rest of the codebase.

server/chat/backend/agent/tools/kubectl_onprem_tool.py
beng360: Same answer — the skills registry is the single source. The kubectl tool's `is_kubectl_onprem_connected` function is called BY the registry via `is_connected_function`, not as a separate source.

server/chat/backend/agent/skills/rca/background/background_context_update.md
beng360: All items addressed:

**HIGH RISK** — all 6 fixed:
1. Response formatting (result/steps_taken/verification) → added to `behavioral_rules.md`
2. Bitbucket dynamic context → template variables `{workspace_slug}`, `{repo_name}`, `{branch_name}` populated from DB
3. kubectl cluster listing → `{cluster_list}` template variable populated from DB
4. AWS multi-account fan-out → added to `background_provider_tools.md`
5. Jira mode enforcement → tool-level hard guardrail in `cloud_tools.py` (write tools not registered in comment_only mode)
6. JQL escaping → `{escaped_service}` auto-escaped in registry context builder

**MEDIUM RISK** — all fixed:
7. Region mapping → full table added to `cloud_access.md`
8-9. cloud_exec + deletion examples → concrete examples in `tool_selection.md`
10. Self-questioning → added to `behavioral_rules.md`
11. Web search examples → 3 concrete examples added
12. Cloudflare routing → fallback check in `load_skills_for_rca` discovers it regardless of integrations dict
13. Tool alias mapping → documented in `rca/tool_mapping.md`

**LOW RISK** (15-23) — addressed where applicable in the respective skill files.

server/chat/backend/agent/prompt/provider_rules.py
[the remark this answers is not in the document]
beng360: False positive — `normalized_mode` is used on the very next line: `if normalized_mode == "ask"`. The variable assignment and usage are adjacent.

server/chat/backend/agent/tools/kubectl_onprem_tool.py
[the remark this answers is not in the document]
beng360: Fixed — `kubectl_onprem_tool.py` now uses `db_pool.get_user_connection()` with a context manager.

server/chat/backend/agent/prompt/context_fetchers.py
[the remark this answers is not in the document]
beng360: Intentional — these are non-critical context enrichment lookups (managed VMs, knowledge base). A failure here should not block prompt assembly. Added explanatory comments.

server/chat/backend/agent/prompt/context_fetchers.py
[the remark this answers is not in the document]
beng360: Connection pool uses context managers for proper cleanup. The bare excepts are intentional for non-critical context enrichment — prompt assembly continues without these optional segments.

server/chat/backend/agent/skills/integrations/cloudbees/SKILL.md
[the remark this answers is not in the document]
beng360: The CloudBees SKILL.md references tool functions that are registered in `cloud_tools.py` — the tools exist and are tested in background RCA runs.

server/chat/backend/agent/skills/rca/provider_ovh.md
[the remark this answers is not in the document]
beng360: Fixed — replaced non-existent `cloud kube kubeconfig generate` with the correct `ovhcloud kube kubeconfig` syntax plus API fallback.

server/chat/backend/agent/prompt/context_fetchers.py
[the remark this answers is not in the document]
beng360: Noted — the `knowledge_base_entries` query is only called once per prompt assembly, not in a hot loop. Adding an index is a good idea but out of scope for this PR.

server/chat/backend/agent/skills/registry.py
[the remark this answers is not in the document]
beng360: The bare `except Exception: return []` is intentional for the deploy lookup — it's non-critical context enrichment and should not block skill loading. The exception is already caught and a fallback value returned. Adding a log line would be noisy for expected failures (e.g., table doesn't exist in dev).

server/chat/backend/agent/skills/registry.py
```diff
@@ -0,0 +1,556 @@
...
+    if value < 500:
+        logger.warning(
+            f"RCA token budget too low ({value}) — using minimum (500)"
+        )
+        return 500
+    return value
+
+
+# Default token budget for auto-loaded RCA skills
+RCA_TOKEN_BUDGET = _get_rca_token_budget()
```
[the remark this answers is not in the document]
beng360: Fixed in `23448ad5` — both branches now return 12000 as the default.

server/chat/backend/agent/tools/splunk_tool.py
```diff
@@ -100,9 +100,10 @@ def _get_splunk_credentials(user_id: str) -> Optional[Dict[str, Any]]:
         creds = get_token_data(user_id, "splunk")
         if not creds:
             return None
-        api_token = creds.get("api_token")
-        base_url = creds.get("base_url")
+        api_token = creds.get("api_token") or creds.get("token") or creds.get("access_token")
+        base_url = creds.get("base_url") or creds.get("url") or creds.get("instance_url")
         if not api_token or not base_url:
+            logger.warning("[SPLUNK-TOOL] Credentials exist but missing expected fields. Keys: %s", list(creds.keys()))
```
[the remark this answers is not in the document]
beng360: Fixed in `23448ad5` — removed `list(creds.keys())` from the log message.

server/chat/background/rca_prompt_builder.py
```diff
@@ -396,6 +480,7 @@ def build_rca_prompt(
     alert_details: Dict[str, Any],
     providers: Optional[List[str]] = None,
     user_id: Optional[str] = None,
+    integrations: Optional[Dict[str, bool]] = None,
```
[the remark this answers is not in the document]
beng360: Fixed in `23448ad5` — `build_rca_prompt` now auto-derives `integrations` from the skill registry when not passed by the caller. Jira/Confluence gating is no longer dead code.

server/chat/backend/agent/skills/integrations/bitbucket/SKILL.md
```diff
@@ -0,0 +1,79 @@
...
+metadata:
+  author: aurora
+  version: "1.0"
+---
+
+# Bitbucket Integration
+
+## Overview
+Bitbucket code repository integration for managing repositories, branches, pull requests, issues, and CI/CD pipelines.
+Connected account: {display_name}
```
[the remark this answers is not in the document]
beng360: Fixed in `23448ad5` — `_get_bitbucket_workspace_context()` now fetches `display_name` from the bitbucket credentials and includes it in the template context.

server/chat/backend/agent/skills/registry.py
```diff
@@ -0,0 +1,642 @@
...
+                    tokens_used += estimate_tokens(body)
+
+        # 3. Load general RCA skills (k8s, ssh) if any cloud provider is connected
+        if providers_lower:
+            for general_id in ("general_k8s", "ssh_investigation"):
+                if general_id in self._rca_skills and tokens_used < RCA_TOKEN_BUDGET:
+                    body = self._rca_bodies.get(general_id, "")
+                    if body:
+                        parts.append(body)
+                        tokens_used += estimate_tokens(body)
```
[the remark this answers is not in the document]
beng360: Fixed in `23448ad5` — general RCA skills (k8s, ssh) now go through `resolve_template(body, extra_ctx)` and token estimation uses the rendered output.

OlivierTrudeau (approved):

damianloch (approved):

PR #259, opened 2026-04-14
fix: RLS policy coverage and session variable handling

damianloch:
## Summary

- **Create CRUD policies for ALL RLS tables**: Previously only 12 of ~30 tables with `FORCE ROW LEVEL SECURITY` had INSERT/UPDATE/DELETE policies. Tables like `user_connections` and `incidents` had zero write policies, causing unconditional DENY on any write. Removed the hardcoded whitelist so all `rls_tables` get full policy coverage.

- **Permissive fallback when session var is unset**: Widened all RLS policy conditions to allow operations when `myapp.current_org_id` is not set (empty/null). When the session var IS set, strict org-scoped multi-tenant isolation is preserved. This prevents crashes from the many code paths that use `get_admin_connection()` without calling `SET myapp.current_org_id`.

- **Reset session vars on pool return**: Clear `myapp.current_org_id` and `myapp.current_user_id` when connections are returned to the pool, preventing stale org_id leakage between requests.

- **Explicit SET on hot write paths**: Added `SET myapp.current_org_id` before writes in `store_tokens_in_db`, `save_connection_metadata`, `set_connection_status`, and `trigger_rca_tool` for proper org isolation on security-sensitive operations.

## Test plan

- [ ] Deploy to K8s with `FORCE ROW LEVEL SECURITY` enabled (e.g. PSL environment)
- [ ] Verify connector setup works (New Relic, AWS, GitHub) -- stores tokens in `user_tokens` and metadata in `user_connections`
- [ ] Verify connectors show as "connected" on the connector status page
- [ ] Trigger an RCA from chat -- inserts into `incidents` table
- [ ] Verify chat sessions auto-create without errors
- [ ] Check logs for absence of `InsufficientPrivilege` / `row-level security policy` errors

## Summary by CodeRabbit

* **Chores**
  * Bumped application version to 1.2.13
  * Enhanced database session management for improved data isolation and security
  * Improved connection pool cleanup procedures
  * Refined row-level security policies for consistent access control

### DIFF
--- a/server/chat/backend/agent/tools/trigger_rca_tool.py
+++ b/server/chat/backend/agent/tools/trigger_rca_tool.py
@@ -123,6 +123,8 @@ def trigger_rca(
     try:
         with db_pool.get_admin_connection() as conn:
             with conn.cursor() as cursor:
+                cursor.execute("SET myapp.current_user_id = %s;", (user_id,))
+                cursor.execute("SET myapp.current_org_id = %s;", (org_id,))
                 cursor.execute(
                     """INSERT INTO incidents
                        (user_id, org_id, source_type, source_alert_id, alert_title,
--- a/server/utils/db/connection_pool.py
+++ b/server/utils/db/connection_pool.py
@@ -98,10 +98,18 @@ class DatabaseConnectionPool:
         finally:
             if connection:
                 try:
+                    connection.rollback()
+                    with connection.cursor() as cur:
+                        cur.execute(
+                            "RESET myapp.current_user_id; RESET myapp.current_org_id;"
+                        )
+                    connection.commit()
+                except Exception:
+                    pass
+                try:
                     pool.putconn(connection)
-                    logger.debug("Returned connection to pool")
                 except Exception as e:
-                    logger.error(f"Error returning connection to pool: {e}")
+                    logger.error("Error returning connection to pool: %s", e)
 
     # Backward compatibility aliases
     def get_user_connection(self):
--- a/server/utils/db/connection_pool.py
+++ b/server/utils/db/connection_pool.py
@@ -104,8 +104,8 @@ class DatabaseConnectionPool:
                             "RESET myapp.current_user_id; RESET myapp.current_org_id;"
                         )
                     connection.commit()
-                except Exception:
-                    pass
+                except Exception as e:
+                    logger.warning("Failed to reset session vars on pool return: %s", e)
                 try:
                     pool.putconn(connection)
                 except Exception as e:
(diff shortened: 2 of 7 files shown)


server/utils/db/db_utils.py
```diff
@@ -1885,84 +1885,57 @@ def initialize_tables():
...
-                # Create org-based SELECT policy (DROP+CREATE to ensure latest definition)
-                policy_sql = f"""
-                    DO $$
-                    BEGIN
+                # RLS condition: enforce org scoping when session var is set,
+                # allow through when unset (prevents crashes from code paths that
+                # don't call SET myapp.current_org_id before writing)
+                _rls_using = f"""
+                    org_id IS NOT NULL
+                    AND (
```
Zarlanx, line 1893: this makes RLS fail-open when session var is unset, any code path that forgets SET myapp.current_org_id now returns cross-org data instead of empty results, the RESET-on-pool-return helps but worth documenting as a conscious security trade-off

server/utils/db/connection_pool.py
```diff
@@ -98,10 +98,15 @@ def get_connection(self):
         finally:
             if connection:
                 try:
+                    connection.rollback()
```
Zarlanx, line 101: if rollback/RESET/commit throws here, putconn is skipped and the connection leaks from the pool, wrapping putconn in a finally block or catching the RESET separately would prevent pool exhaustion under broken connections

server/utils/db/connection_pool.py
```diff
@@ -98,10 +98,15 @@ def get_connection(self):
...
+                    connection.cursor().execute(
```
Zarlanx: connection.cursor() creates a cursor that is never closed, minor resource leak per pool return, could use a with block

server/utils/db/db_adapters.py
```diff
@@ -27,11 +27,15 @@ def close(self):
         """Override close to return connection to pool instead of actually closing."""
         if not self._closed:
             try:
+                self._connection.rollback()
```
Zarlanx, line 30: same pool leak risk as connection_pool.py, if RESET or commit fails the connection is never returned to the pool

server/chat/backend/agent/tools/trigger_rca_tool.py
```diff
@@ -123,6 +123,9 @@ def trigger_rca(
     try:
         with db_pool.get_admin_connection() as conn:
             with conn.cursor() as cursor:
+                cursor.execute("SET myapp.current_user_id = %s;", (user_id,))
+                cursor.execute("SET myapp.current_org_id = %s;", (org_id,))
+                conn.commit()
```
Zarlanx: SET (not SET LOCAL) is session-scoped so it persists regardless of transaction boundaries, the commit here is unnecessary and could confuse readers into thinking SET needs its own transaction

Zarlanx (approved):

PR #265, opened 2026-04-14
feat(helm): support additional ingress resources

damianloch:
## Summary

- Adds `ingress.additional` list to `values.yaml` enabling users to define extra Kubernetes Ingress objects without modifying the primary ingress
- New `additional-ingress.yaml` template iterates over the list and creates one Ingress per entry with full control over className, annotations, TLS, and routing rules
- Service names are passed through as-is (no auto-resolve) -- users write the full Kubernetes service name, which is simpler and supports both Aurora services and external services without magic
- Bumps Helm chart to 1.2.15

## Use Case

Deployments that need a second ingress -- for example, a public-facing ingress for webhook endpoints (OpsGenie/JSM, New Relic) while keeping the primary ingress internal behind a Cloudflare tunnel or VPN. Users can add it declaratively:

```yaml
ingress:
  additional:
    - name: external-webhooks
      className: "nginx"
      annotations:
        nginx.ingress.kubernetes.io/proxy-body-size: "10m"
      rules:
        - host: webhooks.example.com
          http:
            paths:
              - path: /
                pathType: Prefix
                backend:
                  service: aurora-oss-server
                  port: 5080
```

### DIFF
--- a/deploy/helm/aurora/Chart.yaml
+++ b/deploy/helm/aurora/Chart.yaml
@@ -2,8 +2,8 @@ apiVersion: v2
 name: aurora-oss
 description: Aurora – AI-powered cloud operations platform (on-prem deployment)
 type: application
-version: 1.2.14
-appVersion: "1.2.14"
+version: 1.2.15
+appVersion: "1.2.15"
 home: https://github.com/Arvo-AI/aurora
 sources:
   - https://github.com/Arvo-AI/aurora
--- a/deploy/helm/aurora/templates/additional-ingress.yaml
+++ b/deploy/helm/aurora/templates/additional-ingress.yaml
@@ -3,7 +3,7 @@
 apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
-  name: {{ include "aurora.fullname" $ }}-{{ .name }}
+  name: {{ include "aurora.fullname" $ }}-{{ required ".ingress.additional[].name is required" .name }}
   labels:
     {{- include "aurora.labels" $ | nindent 4 }}
   {{- with .annotations }}
--- a/deploy/helm/aurora/templates/additional-ingress.yaml
+++ b/deploy/helm/aurora/templates/additional-ingress.yaml
@@ -1,0 +1,36 @@
+{{- range .Values.ingress.additional }}
+---
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: {{ include "aurora.fullname" $ }}-{{ .name }}
+  labels:
+    {{- include "aurora.labels" $ | nindent 4 }}
+  {{- with .annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+spec:
+  {{- with .className }}
+  ingressClassName: {{ . }}
+  {{- end }}
+  {{- with .tls }}
+  tls:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+  rules:
+    {{- range .rules }}
+    - host: {{ .host | quote }}
+      http:
+        paths:
+          {{- range .http.paths }}
+          - path: {{ .path }}
+            pathType: {{ .pathType | default "Prefix" }}
+            backend:
+              service:
+                name: {{ include "aurora.fullname" $ }}-{{ .backend.service }}
+                port:
+                  number: {{ .backend.port }}
+          {{- end }}
+    {{- end }}
+{{- end }}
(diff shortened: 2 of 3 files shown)


deploy/helm/aurora/templates/additional-ingress.yaml
```diff
@@ -0,0 +1,36 @@
+{{- range .Values.ingress.additional }}
+---
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: {{ include "aurora.fullname" $ }}-{{ .name }}
```
beng360: **Missing `required` validation on `name` and collision risk with primary ingress.**

Two problems here:

1. If `name` is omitted, this renders as `aurora-oss-` — an invalid Kubernetes resource name that silently fails at apply time.
2. If `name: ingress` is passed, this renders `aurora-oss-ingress`, which is the **same name as the primary ingress** (defined in `ingress.yaml` as `aurora-oss-ingress`). Kubernetes will overwrite the primary ingress on the next apply.

Fix:
```yaml
  name: {{ include "aurora.fullname" $ }}-{{ required ".ingress.additional[].name is required" .name }}
```

And add a note in `values.yaml` that `name: ingress` is reserved.

beng360 (changes requested):
## Review Summary

Tested with `helm template` and `helm lint` against the PR branch. The feature works as intended — additional ingresses render correctly, TLS passes through cleanly, and empty `additional: []` produces no output. However there are a few correctness issues that need addressing before merge.

**Verified working:**
- Multiple additional ingresses render as separate `Ingress` objects
- TLS block passes through correctly via `toYaml | nindent`
- `pathType` default of `Prefix` renders properly
- Empty list renders nothing (no guard needed)

**Needs fixing:**
1. Missing `required` on the `name` field — an entry without `name` renders `aurora-oss-` which is an invalid k8s resource name and silently fails at apply time
2. Name collision with the primary ingress — if someone passes `name: ingress`, it renders as `aurora-oss-ingress`, which is the same name as the primary ingress object, causing a silent overwrite
3. Duplicate names in the list produce two objects with the same k8s name — Kubernetes will reject the second apply; no template-level guard or doc warning
4. `appVersion` should not be bumped for a Helm-only change — `appVersion` tracks the application version, only `version` should change here

deploy/helm/aurora/templates/additional-ingress.yaml
```diff
@@ -0,0 +1,36 @@
+{{- range .Values.ingress.additional }}
```
beng360, line 1: **Duplicate names across entries silently produce colliding Kubernetes objects.**

If two entries share the same `name`, both render as `{{ release }}-{{ name }}` — Kubernetes will fail the second `kubectl apply` with a conflict but `helm upgrade` may silently overwrite the first.

No template-level guard is possible with pure Helm, but add a note in `values.yaml` (and the PR description already mentions it) to make the uniqueness requirement explicit:
```yaml
# IMPORTANT: 'name' must be unique across all entries and must not be 'ingress'
# (which is reserved for the primary ingress object).
```

deploy/helm/aurora/Chart.yaml
```diff
@@ -2,8 +2,8 @@ apiVersion: v2
 name: aurora-oss
 description: Aurora – AI-powered cloud operations platform (on-prem deployment)
 type: application
```
beng360, line 4: **`appVersion` should not change for a Helm-only addition.**

`appVersion` is meant to track the version of the application being packaged (the Aurora app itself), not the Helm chart. Since this PR only adds a new Helm template with no application code changes, only `version` should bump.

```yaml
version: 1.2.15
appVersion: "1.2.14"  # unchanged — no app code changed
```

See [Helm docs on Chart.yaml](https://helm.sh/docs/topics/charts/#the-chartyaml-file): *"appVersion: The version of the app that this contains. This doesn't need to be SemVer."*

damianloch: im going to push the tag as soon as merged but i cant include it in PR so i need to increment beforehand

damianloch: Helm ci only runs on appVersion change btw

deploy/helm/aurora/templates/additional-ingress.yaml
damianloch: added comment

deploy/helm/aurora/templates/additional-ingress.yaml
damianloch: did this

beng360 (approved):

PR #266, opened 2026-04-15
fix: save RCA llm_context_history synchronously for citation extraction

isiddharthsingh:
## Summary
- `ContextManager.save_context_history` used an async-queue fast path that raced with `asyncio.run()` teardown in the Celery background chat task — the save task was cancelled before `_execute_actual_save` ran, leaving `llm_context_history = []` and the citation extractor reading zero tool outputs. Every RCA triggered by any connector was affected: incident summaries rendered without `[n]` citation badges.
- Run `_execute_actual_save` synchronously. We're already off the request path in Celery, so the async queue bought no latency — only the race.
- Added `_ensure_llm_context_history` as defense in depth: if the column is still empty post-stream, rebuild synthetic `AIMessage`/`ToolMessage` pairs from `chat_sessions.messages[*].toolCalls` and persist via the same save path.
- Pass the loaded context through to `_extract_tool_calls_for_viz` so the happy path does one DB read instead of two.

## Test plan
- [x] Trigger a fresh RCA — confirmed `Saving context …` + `Saved complete LLM context history …` logs fire before `[CitationExtractor] Extracted N citations` with N > 0
- [x] DB check: `jsonb_array_length(llm_context_history) > 0` and `incident_citations` populated for the verification incident
- [x] Incident page renders `[n]` citation badges inline in the summary

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of chat context persistence to prevent silent data loss.
  * Enhanced recovery mechanism to rebuild chat history when persistence data is unavailable.
  * Fixed race conditions that could result in dropped context updates during background processing.

### DIFF
--- a/server/chat/background/task.py
+++ b/server/chat/background/task.py
@@ -166,12 +166,24 @@ def _ensure_llm_context_history(
 
         llm_context, ui_messages = row[0], row[1]
         if isinstance(llm_context, str):
-            llm_context = json.loads(llm_context) if llm_context else []
+            try:
+                llm_context = json.loads(llm_context) if llm_context else []
+            except (ValueError, TypeError) as e:
+                logger.error(
+                    f"[BackgroundChat] Malformed llm_context_history JSON for session {session_id}: {e}; treating as empty"
+                )
+                llm_context = []
         if llm_context:
             return llm_context
 
         if isinstance(ui_messages, str):
-            ui_messages = json.loads(ui_messages) if ui_messages else []
+            try:
+                ui_messages = json.loads(ui_messages) if ui_messages else []
+            except (ValueError, TypeError) as e:
+                logger.error(
+                    f"[BackgroundChat] Malformed messages JSON for session {session_id}: {e}; cannot rebuild context"
+                )
+                return []
         if not ui_messages:
             return []
 
--- a/server/chat/background/task.py
+++ b/server/chat/background/task.py
@@ -91,27 +91,36 @@ def cancel_rca_for_incident(incident_id: str) -> bool:
         return False
 
 
-def _extract_tool_calls_for_viz(session_id: str, user_id: str) -> List[Dict]:
-    """Extract infrastructure tool calls from database for final visualization."""
+def _extract_tool_calls_for_viz(
+    session_id: str,
+    user_id: str,
+    llm_context: Optional[List[Dict[str, Any]]] = None,
+) -> List[Dict]:
+    """Extract infrastructure tool calls for visualization.
+
+    Accepts a pre-loaded ``llm_context`` to avoid an extra SELECT on the happy
+    path (``_ensure_llm_context_history`` already fetches the column).
+    """
     try:
-        with db_pool.get_admin_connection() as conn:
-            with conn.cursor() as cursor:
-                cursor.execute("""
-                    SELECT llm_context_history
-                    FROM chat_sessions
-                    WHERE id = %s AND user_id = %s
-                """, (session_id, user_id))
-                
-                row = cursor.fetchone()
-        
-        if not row or not row[0]:
-            logger.warning(f"[Visualization] No llm_context_history for session {session_id}")
-            return []
-        
-        llm_context = row[0]
-        if isinstance(llm_context, str):
-            llm_context = json.loads(llm_context)
-        
+        if llm_context is None:
+            with db_pool.get_admin_connection() as conn:
+                with conn.cursor() as cursor:
+                    cursor.execute(
+                        """
+                        SELECT llm_context_history
+                        FROM chat_sessions
+                        WHERE id = %s AND user_id = %s
+                        """,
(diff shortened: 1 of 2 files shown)


server/chat/background/task.py
```diff
@@ -91,41 +91,153 @@ def cancel_rca_for_incident(incident_id: str) -> bool:
...
+                        name=tool_name,
+                    )
+                )
+
+            if not ai_tool_calls:
+                continue
+
+            ai_content = ui_msg.get("content", "") if ui_msg.get("sender") == "bot" else ""
+            rebuilt_messages.append(AIMessage(content=ai_content, tool_calls=ai_tool_calls))
+            rebuilt_messages.extend(tool_messages)
```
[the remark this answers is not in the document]
isiddharthsingh: bot-sender content now reads text first and falls back to content, non-bot stays empty.

server/chat/background/task.py
```diff
@@ -91,41 +91,153 @@ def cancel_rca_for_incident(incident_id: str) -> bool:
...
+                    (session_id, user_id),
+                )
+                row = cursor.fetchone()
+
+        if not row:
+            return None
+
+        llm_context, ui_messages = row[0], row[1]
+        if isinstance(llm_context, str):
+            llm_context = json.loads(llm_context) if llm_context else []
```
Zarlanx: json.loads without try/except, column is JSONB so unlikely to be malformed but a corrupt or manually edited row would crash the entire ensure path

server/chat/background/task.py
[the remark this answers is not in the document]
Zarlanx: agreed, bot messages in chat_sessions.messages use the text field not content, this will produce empty AIMessage.content even when the bot wrote reasoning

server/chat/background/task.py
```diff
@@ -91,41 +91,153 @@ def cancel_rca_for_incident(incident_id: str) -> bool:
...
+
+        if not rebuilt_messages:
+            return []
+
+        ContextManager.save_context_history(session_id, user_id, rebuilt_messages)
+        logger.info(
+            f"[BackgroundChat] Rebuilt llm_context_history for session {session_id} "
+            f"with {len(rebuilt_messages)} synthetic messages"
+        )
+        return [LLMContextManager.serialize_message(m) for m in rebuilt_messages]
```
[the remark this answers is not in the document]
Zarlanx: agreed, if a previous save cached a hash for this session the rebuild silently no-ops in the DB, the in-memory return still feeds visualization but the column stays empty for citation extraction and future reads

server/chat/background/task.py
```diff
@@ -91,41 +91,153 @@ def cancel_rca_for_incident(incident_id: str) -> bool:
...
+            return None
+
+        llm_context, ui_messages = row[0], row[1]
+        if isinstance(llm_context, str):
+            llm_context = json.loads(llm_context) if llm_context else []
+        if llm_context:
+            return llm_context
+
+        if isinstance(ui_messages, str):
+            ui_messages = json.loads(ui_messages) if ui_messages else []
```
Zarlanx: same json.loads without try/except for ui_messages, if this throws the entire fallback aborts and returns None

server/chat/background/task.py
```diff
@@ -91,41 +91,153 @@ def cancel_rca_for_incident(incident_id: str) -> bool:
...
+        if not ui_messages:
+            return []
+
+        logger.warning(
```
Zarlanx, line 190: f-string in logger.warning evaluates even when log level is above WARNING, using logger.warning('...%s', session_id) would defer formatting, minor but this runs on every empty-context fallback

server/chat/background/task.py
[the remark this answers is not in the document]
isiddharthsingh: the fallback now calls ContextManager._get_instance()._execute_actual_save(...) directly to bypass the dedup and returns None on save failure instead of logging success.

server/chat/background/task.py
isiddharthsingh: Malformed messages JSON now logs and returns [] — nothing to rebuild from in that case — instead of bubbling up to the outer catch.

server/chat/background/task.py
isiddharthsingh: Wrapped in try/except; a malformed row now logs an error and treats the column as empty so the fallback can still try rebuilding from messages, instead of hitting the outer except and aborting the whole path.

server/chat/background/task.py
isiddharthsingh: technically correct that f-strings format unconditionally, but WARNING is always enabled by default so the formatting cost is paid either way. The rest of task.py uses f-strings for logging (100+ occurrences), switching one line to %s would be inconsistent. Can revisit as a separate file-wide change if we want to standardize on lazy formatting.

Zarlanx (approved):

PR #268, opened 2026-04-15
Fix streaming

OlivierTrudeau:
- Fixed the streaming for the side chats to have incremental updates
- Made the style of the midchats same as the thoughts 
- Removed the duplicate "RCA: ..." Tab
- Removed Trimming of user message because if it's small enough to get ingested by LLM it's small enough to fit in DB
- Fix RCA Summary context injection appearing as part of the LLM response during mid chat questions
- Fix tables stretching thoughts tab rather than being scrollable

## Summary by CodeRabbit

* **New Features**
  * Background chats can be started with the current question and now persist streaming assistant output for smoother, continuous replies.
  * Tables in markdown now support horizontal scrolling for better readability.

* **Bug Fixes**
  * Chat sessions now exclude RCA entries from chat tabs and restore the correct active tab on incident change.
  * Deterministic merge of live and server messages preserves local “you” messages while adopting assistant updates.
  * Hidden RCA-summary entries are no longer shown in chat.

* **Style**
  * Updated message layout, labels (“You/Aurora”) and “Thinking…” visuals.


server/main_chatbot.py
```diff
@@ -296,6 +296,72 @@ def save_incident_thought(content: str, force: bool = False):
...
+                            bot_msg = msg
+                            break
+
+                    if bot_msg:
+                        bot_msg["text"] = accumulated_text
+                    else:
+                        messages.append({
+                            "sender": "bot",
+                            "text": accumulated_text,
+                            "_streaming": True,
```
beng360, line 356: The `_streaming` flag gets set here but never gets cleaned up when the workflow finishes. `_update_session_status` just sets the status column — it doesn't touch the messages JSON.

So after the chat completes, you've got messages sitting in the DB with `{"_streaming": true}` permanently. Probably harmless right now since the frontend doesn't read that flag, but it's misleading data. Either strip it out in the final save or at least set it to `false` when the force-flush happens at the end.

server/main_chatbot.py
```diff
@@ -296,6 +296,72 @@ def save_incident_thought(content: str, force: bool = False):
...
+        if not should_flush or (not force and len(accumulated_text) < 30):
+            return
+
+        try:
+            from utils.db.connection_pool import db_pool
+
+            with db_pool.get_admin_connection() as conn:
+                with conn.cursor() as cursor:
+                    cursor.execute(
+                        "SELECT messages FROM chat_sessions WHERE id = %s FOR UPDATE",
```
beng360, line 332: `SELECT ... FOR UPDATE` on every flush (every 1.5s or 200 chars) is a pretty heavy pattern. During an active RCA with streaming, this is holding a row lock on `chat_sessions` while doing JSON parse + modify + write.

If the frontend is also polling that same row every 2s, you could get lock contention. Not a showstopper for now but worth noting — if you see slow chat polling during RCA, this is probably why.

beng360 (dismissed):
Solid improvements overall — streaming for side chats, the RCA tab dedup, and the table overflow fix all look good. A few things need addressing before merge though.

server/main_chatbot.py
beng360, line 356: `_streaming: true` gets written here and never removed. After the chat finishes, every bot message in the DB still has this flag set permanently. You need to strip it when the session completes — otherwise this is just junk data accumulating in every chat session forever.

server/main_chatbot.py
beng360, line 332: You're doing `SELECT ... FOR UPDATE` every 1.5 seconds on the same row the frontend polls every 2 seconds. That's a row lock on `chat_sessions` while you parse JSON, mutate it, and write it back. This will cause visible lag on the frontend during active streaming — the poll query blocks waiting for this lock to release.

server/main_chatbot.py
```diff
@@ -296,6 +296,72 @@ def save_incident_thought(content: str, force: bool = False):
...
+
+    def save_streaming_chat_message(content: str, force: bool = False):
+        """Incrementally save the assistant's streaming response to chat_sessions.messages."""
+        if not is_background or not session_id or session_id == 'unknown':
+            return
+
+        if content:
+            accumulated_chat_msg.append(content)
+
+        accumulated_text = "".join(accumulated_chat_msg)
```
beng360, line 314: `force=True` doesn't actually force anything. When `accumulated_text` is empty (like line 606 where you call `force=True` with empty string before starting a new message), the `len(accumulated_text) < 30` check returns early. So the "force flush" path is dead code in that case.

server/chat/backend/agent/workflow.py
```diff
@@ -1440,6 +1440,9 @@ def _convert_to_ui_messages(self, llm_messages, tool_capture=None):
                 
             elif 'AI' in msg_type:
                 raw_content = getattr(msg, 'content', '')
```
beng360, line 1453: This is going to break. If anyone touches the prefix string in `_compress_rca_context`, the summary starts leaking into chat and nobody will know why. Extract this into a constant that both places import.

client/src/app/incidents/components/ThoughtsPanel.tsx
```diff
@@ -157,21 +160,36 @@ export default function ThoughtsPanel({ thoughts, incident, isVisible, canIntera
...
+          const prevAssistantLast = prevAssistant[prevAssistant.length - 1]?.content || '';
+          const serverAssistantLast = serverAssistantMessages[serverAssistantMessages.length - 1]?.content || '';
+          if (prevAssistant.length === serverAssistantMessages.length && prevAssistantLast === serverAssistantLast) {
+            return prev;
+          }
+
+          // Server has new assistant content but hasn't persisted all our user messages yet.
+          // Keep server order and append only the optimistic user messages the server is missing.
+          const optimisticTail = localUserMessages.slice(serverUserMessages.length);
+          return [...messages, ...optimisticTail];
```
beng360, line 201: What happens when the user sends a second message before the server has persisted the first? `optimisticTail` grabs both unpersisted messages and appends them after `messages`. Then when the server catches up with the first one, you get a duplicated message briefly visible until the next poll cycle reconciles it.

server/chat/background/task.py
```diff
@@ -1934,6 +1935,10 @@ def create_background_chat_session(
                 if trigger_metadata:
                     ui_state["triggerMetadata"] = trigger_metadata
                 
+                initial_messages = []
+                if question:
+                    initial_messages.append({"sender": "user", "text": question})
```
beng360, line 1940: The raw question gets saved here, but by the time the LLM sees it the backend has wrapped it in `<user_message>` tags. So when the session completes and messages get overwritten with the final state, the display changes from the raw question to the wrapped version. The frontend's `extractUserMessage` handles it but it's still a flash of different content on reload.

beng360 (changes requested):
Streaming side chats and the RCA tab dedup are needed fixes. But there are real issues here that will bite us in prod.

client/src/app/incidents/components/ThoughtsPanel.tsx
```diff
@@ -116,8 +119,8 @@ export default function ThoughtsPanel({ thoughts, incident, isVisible, canIntera
     // Don't reset if we're creating a session
     if (creatingSessionIds.current.size > 0) return;
     
-    if (incident.activeTab === 'chat' && incident.chatSessions && incident.chatSessions.length > 0) {
-      setActiveTab(incident.chatSessions[incident.chatSessions.length - 1].id);
+    if (incident.activeTab === 'chat' && chatSessions.length > 0) {
+      setActiveTab(chatSessions[chatSessions.length - 1].id);
```
beng360: You switched from `incident.chatSessions` (prop) to `chatSessions` (state) but the dep array is still `[incident.id]`. Before this change it was fine because reading the prop directly doesn't go stale. Now `chatSessions` is captured in the closure and never updates — so on the first incident load with `activeTab === 'chat'`, this reads whatever `chatSessions` was at mount time, not the current value.

Either add `chatSessions` to the dep array or switch back to reading from the prop here.

client/src/app/incidents/components/ThoughtsPanel.tsx
```diff
@@ -85,7 +85,10 @@ export default function ThoughtsPanel({ thoughts, incident, isVisible, canIntera
   // Preserves optimistic sessions that exist in local state but not yet in parent's data.
   useEffect(() => {
     // Parent's data: from incident prop (polled from backend every 3s)
-    const incidentSessions = incident.chatSessions || [];
+    // Exclude the RCA session — its content is already shown in the Thoughts tab
+    const incidentSessions = (incident.chatSessions || []).filter(
+      (s: ChatSession) => s.id !== incident.chatSessionId
```
beng360, line 92: When `incident.chatSessionId` is `undefined` (which it will be until the backend populates it), this filter becomes `s.id !== undefined` — always true. So the RCA session doesn't get filtered out on initial load. The whole point of this change is to hide the duplicate RCA tab, but it silently does nothing until `chatSessionId` shows up in a later poll cycle.

client/src/app/incidents/components/ThoughtsPanel.tsx
OlivierTrudeau: would require mega speed and would be solved after anyways. This is side chat, good luck sending 2 messages in a row fast enough

client/src/app/incidents/components/ThoughtsPanel.tsx
OlivierTrudeau: done

client/src/app/incidents/components/ThoughtsPanel.tsx
OlivierTrudeau: done

server/chat/background/task.py
OlivierTrudeau: done

server/chat/backend/agent/workflow.py
OlivierTrudeau: used a constant

server/main_chatbot.py
OlivierTrudeau: now returns early if nothing

server/main_chatbot.py
OlivierTrudeau: strip the flag before storing now

server/main_chatbot.py
OlivierTrudeau: frontend poll is just a read so it's non blocking

OlivierTrudeau: it's read only so not blocking I believe

server/main_chatbot.py
OlivierTrudeau: done

client/src/app/incidents/components/ThoughtsPanel.tsx
beng360: okok

client/src/app/incidents/components/ThoughtsPanel.tsx
```diff
@@ -157,21 +165,36 @@ export default function ThoughtsPanel({ thoughts, incident, isVisible, canIntera
...
+          const prevAssistantLast = prevAssistant[prevAssistant.length - 1]?.content || '';
+          const serverAssistantLast = serverAssistantMessages[serverAssistantMessages.length - 1]?.content || '';
+          if (prevAssistant.length === serverAssistantMessages.length && prevAssistantLast === serverAssistantLast) {
+            return prev;
           }
+
+          // Server has new assistant content but hasn't persisted all our user messages yet.
+          // Keep server order and append only the optimistic user messages the server is missing.
+          const optimisticTail = localUserMessages.slice(serverUserMessages.length);
+          return [...messages, ...optimisticTail];
```
beng360, line 201: This causes a React key collision. `optimisticTail` messages keep their original IDs (e.g. `session-4`) and when they get appended after `messages`, the server array already has a message at that same index with the same key. Getting `Encountered two children with the same key` in console.

Need to either dedupe by content before appending, or re-key the optimistic messages with a suffix so they can't collide with the server-indexed ones.

beng360 (approved):

PR #270, opened 2026-04-15
fix: handle missing fingerprint in Grafana alert webhooks

beng360:
## Summary
- Grafana alert webhooks without a `fingerprint` field cause a `NOT NULL` constraint violation on `source_alert_id`, silently dropping the incident
- Synthesize a deterministic fingerprint from the alert's `labels` JSON when the field is absent, so every webhook creates a trackable incident
- Discovered during stress testing: the RCA agent itself found this bug while investigating a PostgreSQLReplicationLag alert

## Test plan
- [ ] Send a Grafana webhook **without** a `fingerprint` field — verify incident is created (previously failed silently)
- [ ] Send a Grafana webhook **with** a `fingerprint` field — verify behavior is unchanged
- [ ] Send duplicate webhooks with the same labels but no fingerprint — verify deduplication via ON CONFLICT works


OlivierTrudeau:
not sure abotu all of this. I tested grafana on the SAAS and it works

beng360:
Closing — Grafana always sends a fingerprint (hash of rule UID + labels, computed server-side). No real-world scenario where this triggers. Not worth the merge noise for a defensive guard against hand-crafted payloads we don't support.

PR #281, opened 2026-04-18
Next.js API proxy layer + WebSocket JWT authentication

OlivierTrudeau:
## Summary

Eliminates direct browser-to-Flask API calls by routing all authenticated requests through Next.js API routes, and adds WebSocket token-based handshake authentication. This is a subset of the larger `more-saas-sec` branch, isolated for easier review.

### Next.js API proxy layer
- **`client/src/lib/backend-proxy.ts`** — single `forwardRequest()` helper that reads Auth.js session via `auth-helper.ts`, injects `X-User-ID`, `X-Org-ID`, and `X-Internal-Secret` headers, then forwards body/method/query to Flask
- **22 proxy route files** under `client/src/app/api/proxy/` using a DRY `handler` + re-export pattern (`export { handler as GET, handler as POST, ... }`)
- All authenticated browser HTTP calls now go through `/api/proxy/...` → Next.js server → Flask
- `NEXT_PUBLIC_BACKEND_URL` kept on client for display-only purposes (MCP config JSON, kubectl Helm commands, Azure/AWS setup scripts)
- Registration (`/api/auth/register`) remains a direct browser-to-Flask call (unauthenticated endpoint)

### WebSocket token-based handshake authentication
- **`client/src/app/api/ws-token/route.ts`** — generates 60-second HS256 JWT containing `userId`/`orgId`, signed with `INTERNAL_API_SECRET`
- **`client/src/hooks/useWebSocket.ts`** — fetches token before connecting, appends `?token=xxx` to WS URL
- **`server/main_chatbot.py`** — `_validate_ws_token()` validates JWT on handshake, sets `current_user_id`; rejects identity mismatch if `init` message claims a different user
- Backwards compatible: if `INTERNAL_API_SECRET` is unset, falls back to legacy init-based auth

### INTERNAL_API_SECRET enforcement (Flask)
- Flask `before_request` hook requires `X-Internal-Secret` on all endpoints except:
  - `_OPEN_PATHS`: `/api/auth/login`, `/api/auth/register`, `/health`
  - `_OPEN_PREFIXES`: webhook endpoints, OAuth callbacks, event endpoints, setup scripts (these have their own auth: HMAC, OAuth state, signing secrets)
- Backwards compatible: skipped entirely when `INTERNAL_API_SECRET` env var is empty

### Client-side changes
- All service files (`rcaEmails.ts`, `graph-discovery.ts`, `bitbucket-integration-service.ts`, etc.) updated to call `/api/proxy/...` instead of `NEXT_PUBLIC_BACKEND_URL` directly
- All hooks (`use-connector-status.ts`, `use-bitbucket-status.ts`, `use-github-status.ts`, etc.) updated similarly
- Page components (`aws/onboarding`, `azure/auth`, `cloudflare/auth`, `vm-config`, etc.) updated
- Dead code removed from `GeneralSettings.tsx`
- `PyJWT>=2.8.0` added to `server/requirements.txt`

## Test plan

- [ ] `make dev` starts cleanly, frontend loads
- [ ] Browser network tab shows API calls going to `/api/proxy/...` (no direct `:5080` calls except registration)
- [ ] WebSocket connects with `?token=xxx` query parameter visible in network tab
- [ ] With `INTERNAL_API_SECRET` set: direct `curl` to Flask returns 403; same request via UI works
- [ ] Webhook endpoints (e.g. Datadog, Slack events) still work without the secret header
- [ ] OAuth flows (GitHub, Bitbucket, Slack) complete successfully
- [ ] MCP client config and kubectl Helm commands show correct backend URL
- [ ] Without `INTERNAL_API_SECRET`: everything works as before (backwards compatible)

## Summary by CodeRabbit

* **New Features**
  * WebSocket JWT token support for authenticated real-time connections
  * Many new frontend proxy endpoints and a short-lived token endpoint for WS auth

* **Bug Fixes**
  * Centralized server-side enforcement of internal-secret for proxied requests
  * Removed client-side user ID propagation to tighten authentication behavior

* **Chores**
  * Migrated client API calls to use the frontend proxy layer
  * Added internal-secret validation and updated server dependencies (JWT support)


client/src/app/api/ws-token/route.ts
```diff
@@ -0,0 +1,33 @@
...
+  const secret = new TextEncoder().encode(INTERNAL_API_SECRET);
+
+  const token = await new SignJWT({
+    userId: session.userId,
+    orgId: session.orgId ?? null,
+  })
+    .setProtectedHeader({ alg: 'HS256' })
+    .setIssuedAt()
+    .setExpirationTime('60s')
+    .sign(secret);
```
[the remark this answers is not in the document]
OlivierTrudeau: Added the jti but the other in't necessary

client/src/lib/services/graph-discovery.ts
```diff
@@ -38,15 +26,9 @@ export interface DiscoveryStatus {
...
+  taskId: string,
 ): Promise<DiscoveryStatus> {
-  const res = await fetch(
-    `${BACKEND_URL}/api/graph/discover/status/${taskId}`,
-    {
-      headers: { "X-User-ID": userId },
-    }
-  );
-  if (!res.ok) throw new Error("Failed to poll discovery status");
+  const res = await fetch(`/api/proxy/graph/discover/status/${taskId}`);
```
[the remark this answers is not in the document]
OlivierTrudeau: fixed

server/main_compute.py
```diff
@@ -212,19 +212,84 @@
...
+        return None
+
+    if any(request.path.startswith(prefix) for prefix in _OPEN_PREFIXES):
+        return None
+
+    provided = request.headers.get("X-Internal-Secret", "")
+    if not provided or provided != _INTERNAL_API_SECRET:
+        return jsonify({"error": "Forbidden: invalid or missing X-Internal-Secret header"}), 403
+
+    return None
```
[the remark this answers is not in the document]
OlivierTrudeau: fixed

client/src/app/api/ws-token/route.ts
```diff
@@ -0,0 +1,34 @@
+import { NextResponse } from 'next/server';
+import { SignJWT } from 'jose';
+import { auth } from '@/auth';
+
+const INTERNAL_API_SECRET = process.env.INTERNAL_API_SECRET || '';
```
Zarlanx, line 5: INTERNAL_API_SECRET is used both as the HS256 signing key for WS tokens AND as the Flask header secret for proxy calls. Compromise of the shared value yields WS impersonation plus backend access. Derive the WS key from the shared secret plus a static label or introduce a separate env var.

client/src/app/api/ws-token/route.ts
```diff
@@ -0,0 +1,34 @@
...
+  const secret = new TextEncoder().encode(INTERNAL_API_SECRET);
+
+  const token = await new SignJWT({
+    userId: session.userId,
+    orgId: session.orgId ?? null,
+    jti: crypto.randomUUID(),
+  })
+    .setProtectedHeader({ alg: 'HS256' })
+    .setIssuedAt()
+    .setExpirationTime('60s')
```
Zarlanx: 60s lifetime is tight when the WS handshake is slow (cold start, TLS resumption miss). Users on flaky networks will see token-expired failures and trigger a reconnect loop. Bump to 120-180s or re-fetch on first WS handshake failure.

client/src/app/api/ws-token/route.ts
```diff
@@ -0,0 +1,34 @@
...
+      { status: 500 },
+    );
+  }
+
+  const secret = new TextEncoder().encode(INTERNAL_API_SECRET);
+
+  const token = await new SignJWT({
+    userId: session.userId,
+    orgId: session.orgId ?? null,
+    jti: crypto.randomUUID(),
```
Zarlanx, line 28: jti is generated but never recorded anywhere. Anyone who captures a token within the 60s window can replay it. For a purely-internal WS this may be acceptable but should be documented; otherwise track jti in a Redis set with TTL to enforce single-use.

client/src/app/api/ws-token/route.ts
```diff
@@ -0,0 +1,34 @@
...
+  if (!INTERNAL_API_SECRET) {
+    return NextResponse.json(
+      { error: 'Server misconfiguration: INTERNAL_API_SECRET not set' },
+      { status: 500 },
+    );
+  }
+
+  const secret = new TextEncoder().encode(INTERNAL_API_SECRET);
+
+  const token = await new SignJWT({
```
Zarlanx, line 24: Payload has no aud/iss claim. If other consumers ever sign HS256 with the same secret (CLI auth, cron jobs, future webhooks), their tokens cross-authenticate against the WS. Add aud: 'chatbot-ws' and verify it on the server.

client/src/lib/server-env.ts
```diff
@@ -0,0 +1,17 @@
...
+  const value = process.env[name];
+  if (!value) {
+    throw new Error(`Missing required environment variable: ${name}`);
+  }
+  return value;
+}
+
+export const env = {
+  BACKEND_URL: required('BACKEND_URL'),
+  INTERNAL_API_SECRET: process.env.INTERNAL_API_SECRET || '',
```
Zarlanx: INTERNAL_API_SECRET || '' silently disables enforcement when the env is empty. A misconfigured prod deploy ships without auth and nothing fails loudly. Use required() in production or expose a health check that flags 'secret not set' so ops sees it.

client/src/lib/backend-proxy.ts
```diff
@@ -1,36 +1,69 @@
...
- */
-export async function forwardAuthenticatedGet(
+export async function forwardRequest(
   request: NextRequest,
+  method: string,
   backendPath: string,
   errorLabel: string,
+  options: { timeoutMs?: number; passBody?: boolean } = {},
 ): Promise<NextResponse> {
+  const { timeoutMs = 10_000, passBody = method !== 'GET' } = options;
```
Zarlanx: Default 10_000ms timeout is too tight for cloud-provider operations (graph discovery, setup-script generation, kubectl helm install commands). Users hit 504s on legitimate long-running calls. Either raise the default or force every long-op caller to pass timeoutMs explicitly.

Zarlanx: passBody = method !== 'GET' is a coarse heuristic. HEAD requests should not forward a body, and Next.js may allow custom methods (PATCH, OPTIONS) where the right behavior depends on the route. Require callers to declare passBody explicitly or match on a known-method allowlist.

client/src/lib/backend-proxy.ts
```diff
@@ -1,36 +1,69 @@
...
+    }
+
+    let body: BodyInit | undefined;
+    if (passBody) {
+      const ct = request.headers.get('content-type') || '';
+      if (ct.includes('application/json')) {
+        headers['Content-Type'] = 'application/json';
+        body = await request.text();
+      } else if (ct.includes('multipart/form-data')) {
+        body = await request.arrayBuffer() as unknown as BodyInit;
```
Zarlanx: multipart body is read fully into an arrayBuffer and forwarded. A 100MB knowledge-base upload is now buffered twice (incoming to Next.js, outgoing to Flask). Stream via ReadableStream or forward the request body directly to avoid doubling memory usage on the proxy hop.

server/main_chatbot.py
```diff
@@ -75,14 +75,50 @@ def is_allowed(self, client_id):
...
+def _validate_ws_token(websocket) -> dict | None:
+    """Extract and validate a JWT from the WebSocket handshake query string.
+
+    Returns the decoded payload dict on success, or None if no token is present or validation fails.
+    """
+    if not _INTERNAL_API_SECRET:
+        return None
+
+    try:
+        qs = parse_qs(urlparse(str(websocket.request.path)).query)
```
Zarlanx: Three fallbacks for finding the token (websocket.request.path, raw path split on ?, getattr path) mean any of them can succeed in isolation. Use the websockets library's documented request_uri or path_with_query API consistently and fail if that one source is absent.

server/main_chatbot.py
```diff
@@ -75,14 +75,50 @@ def is_allowed(self, client_id):
...
+            request_path = getattr(websocket, "path", "")
+            if "?" in request_path:
+                qs = parse_qs(request_path.split("?", 1)[1])
+
+        token_list = qs.get("token")
+        if not token_list:
+            return None
+
+        token = token_list[0]
+        payload = pyjwt.decode(token, _INTERNAL_API_SECRET, algorithms=["HS256"])
```
Zarlanx: pyjwt.decode does not verify aud, iss, or sub. A token signed with the shared secret for any other purpose (future API auth, CLI tool) authenticates against this WS. Add options={'require': ['exp', 'aud']} and set audience='chatbot-ws'.

server/main_chatbot.py
```diff
@@ -93,11 +129,26 @@ def _normalize_mode(mode: Optional[str]) -> str:
...
+    """Handle connection initialization message and return updated state.
+
+    If the connection was already authenticated via a handshake token,
+    current_user_id is pre-set and we skip the DB validation step.
+    """
 
     user_id = data.get('user_id')
 
-    if user_id and current_user_id != user_id:
+    if user_id and current_user_id and current_user_id != user_id:
```
Zarlanx, line 161: Identity mismatch between token and init sends an error message but does NOT close the socket. An attacker with a valid token for user A can keep sending init with different user_ids until one is accepted. Close the WS on first mismatch with code 1008.

server/main_chatbot.py
```diff
@@ -93,11 +129,26 @@ def _normalize_mode(mode: Optional[str]) -> str:
...
+            f"WebSocket init rejected: token user {current_user_id!r} "
+            f"does not match init user_id {user_id!r}"
+        )
+        await websocket.send(json.dumps({
+            "type": "error",
+            "data": {"text": "Authentication failed: user identity mismatch."}
+        }))
+        return current_user_id, deployment_listener_task
+
+    if user_id and not current_user_id:
```
Zarlanx, line 173: Legacy init-based auth runs when current_user_id is unset (token missing/invalid). When INTERNAL_API_SECRET is configured, this path is unreachable by design, but it remains executable if the secret env is cleared at runtime. Guard with 'if not _INTERNAL_API_SECRET' so the legacy path cannot execute while enforcement is configured.

client/src/hooks/useWebSocket.ts
```diff
@@ -230,7 +214,37 @@ export const useWebSocket = (config: WebSocketConfig) => {
     }));
 
     try {
-      const ws = new WebSocket(configRef.current.url);
+      let wsUrl = configRef.current.url;
+
+      // Fetch a short-lived token from the Next.js API route for handshake auth
+      try {
+        const tokenRes = await fetch('/api/ws-token');
```
Zarlanx, line 222: Every reconnect attempt fetches a fresh token. With exponential backoff up to 30s plus 1s jitter plus 10 attempts, a disconnected client makes 10 GET /api/ws-token calls. If the token endpoint is rate-limited or throttled, the reconnect storm compounds the outage. Cache the token until expiry and re-use across attempts.

client/src/hooks/useWebSocket.ts
```diff
@@ -230,7 +214,37 @@ export const useWebSocket = (config: WebSocketConfig) => {
...
+            wsUrl = `${wsUrl}${separator}token=${encodeURIComponent(token)}`;
+          }
+        } else {
+          console.error('Failed to fetch WS token, aborting connection');
+          setState(prev => ({
+            ...prev,
+            isConnecting: false,
+            error: 'Authentication failed: unable to obtain WebSocket token'
+          }));
+          return;
```
Zarlanx, line 242: When token fetch fails, state is updated but shouldReconnectRef stays true. The next close event fires the same reconnect logic and the token fetch fails again. A session that expired client-side loops forever. Flip shouldReconnectRef to false on auth-level failure so the loop halts until the user re-authenticates.

client/src/hooks/useWebSocket.ts
```diff
@@ -230,7 +214,37 @@ export const useWebSocket = (config: WebSocketConfig) => {
...
+      let wsUrl = configRef.current.url;
+
+      // Fetch a short-lived token from the Next.js API route for handshake auth
+      try {
+        const tokenRes = await fetch('/api/ws-token');
+        if (tokenRes.ok) {
+          const { token } = await tokenRes.json();
+          if (token) {
+            const separator = wsUrl.includes('?') ? '&' : '?';
+            wsUrl = `${wsUrl}${separator}token=${encodeURIComponent(token)}`;
```
Zarlanx: wsUrl concatenation checks for a '?' but does not strip an existing token= parameter. If the caller passes a URL that already has a token, this appends a second one. Most servers pick the first, not the fresh one. Parse and rebuild the URL with URL/URLSearchParams.

client/src/lib/backend-proxy.ts
```diff
@@ -55,8 +88,23 @@ export async function forwardAuthenticatedGet(
...
+      return NextResponse.json(data, { status: response.status });
+    }
+
+    const text = await response.text();
+    return new NextResponse(text, {
+      status: response.status,
+      headers: { 'Content-Type': ct || 'text/plain' },
+    });
   } catch (error) {
     console.error(`[api/${errorLabel}] Error:`, error);
```
Zarlanx: console.error logs the full error object including any request metadata. Next.js server logs often ship to centralized log aggregation; auth headers or request paths containing tokens can leak. Scrub headers from the logged object before console.error.

server/main_compute.py
```diff
@@ -212,19 +213,89 @@
...
+
+@app.before_request
+def verify_internal_api_secret():
+    """Reject requests that don't carry a valid INTERNAL_API_SECRET header.
+
+    Skipped when the secret is not configured (backward-compatible default)
+    and for open endpoints (login, register, health) and external-facing
+    webhook/callback/event endpoints called by third-party services.
+    """
+    if not _INTERNAL_API_SECRET:
```
Zarlanx, line 272: if not _INTERNAL_API_SECRET: return None silently disables enforcement when the env is unset. A prod deploy without the secret has no auth on internal endpoints and nothing surfaces. Emit a loud startup warning or require the env when APP_ENV=production.

server/main_compute.py
```diff
@@ -212,19 +213,89 @@
...
+        return None
+
+    if request.method == "OPTIONS":
+        return None
+
+    if request.path in _OPEN_PATHS:
+        return None
+
+    if any(request.path.startswith(prefix) for prefix in _OPEN_PREFIXES):
```
Zarlanx: startswith(prefix) matching allows suffix-manipulation attacks: /datadog/webhook/ matches /datadog/webhook/../something if path normalization is wrong anywhere upstream. Switch to path.startswith(prefix + '/') plus an exact-match check to tighten.

server/main_compute.py
```diff
@@ -212,19 +213,89 @@
...
+# ============================================================================
+# Internal API Secret Verification
+# ============================================================================
+# Ensures requests originate from the Next.js frontend (or another trusted
+# internal service) rather than from an unauthenticated external caller.
+
+_INTERNAL_API_SECRET = os.getenv("INTERNAL_API_SECRET", "")
+_OPEN_PATHS = frozenset(("/api/auth/login", "/api/auth/register"))
+
+_OPEN_PREFIXES = (
```
Zarlanx, line 233: _OPEN_PREFIXES is a hardcoded list. Any new public endpoint (password reset, OAuth start, SAML ACS) must be added here or gets 403. There's no audit log at startup listing what's publicly reachable; consider logging the resolved open paths so ops can review in each release.

server/main_compute.py
```diff
@@ -212,19 +213,89 @@
...
+    "/ovh_api/ovh/oauth2/callback",
+    "/azure/setup-script",
+    "/azure/setup-script-ps1",
+    "/aws/setup-script",
+    "/aws/setup-role",
+    "/aws/setup-script-ps1",
+    "/aws/setup-role-ps1",
+)
+
+@app.before_request
```
Zarlanx, line 264: Flask before_request hook ordering matters: if CORS middleware runs AFTER this check, a 403 response may lack CORS headers and the browser sees a generic fetch error instead of the actual message. Verify CORS runs first or emit CORS headers on the 403 response.

server/main_chatbot.py
```diff
@@ -75,14 +75,50 @@ def is_allowed(self, client_id):
...
 
+_INTERNAL_API_SECRET = os.getenv("INTERNAL_API_SECRET", "")
+
+
+def _validate_ws_token(websocket) -> dict | None:
+    """Extract and validate a JWT from the WebSocket handshake query string.
+
+    Returns the decoded payload dict on success, or None if no token is present or validation fails.
+    """
+    if not _INTERNAL_API_SECRET:
```
Zarlanx, line 95: When INTERNAL_API_SECRET is unset, _validate_ws_token returns None and handle_connection falls through to legacy init-based auth. A prod deploy without the secret has no handshake auth. Fail loud at startup when the secret is missing in non-dev environments.

client/src/lib/backend-proxy.ts
```diff
@@ -55,8 +88,23 @@ export async function forwardAuthenticatedGet(
...
+      return new NextResponse(null, { status: response.status });
+    }
+
+    if (ct.includes('application/json')) {
+      const data = await response.json();
+      return NextResponse.json(data, { status: response.status });
+    }
+
+    const text = await response.text();
+    return new NextResponse(text, {
```
Zarlanx, line 114: Non-JSON responses are passed through verbatim via new NextResponse(text, ...). If Flask returns an HTML page (werkzeug stack trace in dev, a reverse-proxy 502 page, a cloud-provider error page) that HTML flows all the way to the browser with its original content-type. Wrap or redact non-JSON responses so debug pages do not leak through the proxy.

server/main_compute.py
```diff
@@ -212,19 +213,89 @@
...
+_TENANT_OPEN_PREFIXES = ("/api/auth/login", "/api/auth/register", "/health")
 
 @app.before_request
 def enforce_user_org_binding():
     """Reject requests where X-Org-ID doesn't match the user's actual org."""
     if request.method == "OPTIONS":
         return None
 
-    if any(request.path.startswith(p) for p in _OPEN_PREFIXES):
+    if any(request.path.startswith(p) for p in _TENANT_OPEN_PREFIXES):
```
Zarlanx, line 311: _TENANT_OPEN_PREFIXES is a second skip list parallel to _OPEN_PREFIXES on the internal-secret check above. They are maintained separately and will drift: a new public endpoint added to one but not the other creates an asymmetry where a route is exempt from one check but not the other. Unify into one source of truth or document the reason for the split.

client/src/app/api/ws-token/route.ts
OlivierTrudeau: all good

OlivierTrudeau: Fixed — WS key is now derived with a label suffix.

client/src/app/api/ws-token/route.ts
OlivierTrudeau: Bumped to 120s.

client/src/app/api/ws-token/route.ts
OlivierTrudeau: Enforced — jti is now single-use via Redis.

client/src/app/api/ws-token/route.ts
OlivierTrudeau: Added aud: 'chatbot-ws', verified server-side.

client/src/lib/server-env.ts
OlivierTrudeau: Fails at startup in non-dev environments now.

client/src/lib/backend-proxy.ts
OlivierTrudeau: Raised to 30s default.

OlivierTrudeau: Uses a METHODS_WITHOUT_BODY set (GET, HEAD, OPTIONS).

client/src/lib/backend-proxy.ts
OlivierTrudeau: Streams via request.body now, no double-buffer.

server/main_chatbot.py
OlivierTrudeau: Simplified to just websocket.request.path (websockets v15).

server/main_chatbot.py
OlivierTrudeau: Added audience='chatbot-ws' with require=['exp','aud'].

server/main_chatbot.py
OlivierTrudeau: Now closes with code 1008 on mismatch.

server/main_chatbot.py
OlivierTrudeau: Guarded — rejects with close if secret is configured.

client/src/hooks/useWebSocket.ts
OlivierTrudeau: Fresh token per connect; backoff + max attempts caps retries.

client/src/hooks/useWebSocket.ts
OlivierTrudeau: Fixed — stops reconnecting on auth failure (401/403).

client/src/hooks/useWebSocket.ts
OlivierTrudeau: Rebuilt with URL/URLSearchParams now.

client/src/lib/backend-proxy.ts
OlivierTrudeau: Scrubbed — only logs error name and message now.

server/main_compute.py
OlivierTrudeau: Centralized — startup_checks.py fails the container.

server/main_compute.py
OlivierTrudeau: Tightened to exact match or prefix+slash boundary.

server/main_compute.py
OlivierTrudeau: Acknowledged, won't change for now.

server/main_compute.py
OlivierTrudeau: Verified — flask-cors after_request runs on early returns.

server/main_chatbot.py
OlivierTrudeau: Same centralized startup check covers this.

client/src/lib/backend-proxy.ts
OlivierTrudeau: HTML responses are now redacted with a 502.

server/main_compute.py
OlivierTrudeau: Added comment clarifying why they differ (by design).

server/main_chatbot.py
```diff
@@ -75,14 +75,68 @@ def is_allowed(self, client_id):
...
 rate_limiter = RateLimiter(rate=5, per=60)
 
+_INTERNAL_API_SECRET = os.getenv("INTERNAL_API_SECRET", "")
+_AURORA_ENV = os.getenv("AURORA_ENV", "production")
+
+if _AURORA_ENV != "dev" and not _INTERNAL_API_SECRET:
+    raise RuntimeError(
+        "FATAL: INTERNAL_API_SECRET is not set and AURORA_ENV='%s' (non-dev). "
+        "Refusing to start without authentication secrets in production." % _AURORA_ENV
+    )
```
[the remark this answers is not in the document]
OlivierTrudeau: Intentional. I don't want someone to accidentally not configure their vars correctly and be opened to attack. I'd rather have dev fail because they didn't set their env var correctly

server/main_chatbot.py
```diff
@@ -75,14 +75,68 @@ def is_allowed(self, client_id):
...
+        jti = payload.get("jti")
+        if not jti:
+            logger.warning("WebSocket token missing required jti claim")
+            return None
+
+        from utils.cache.redis_client import get_redis_client
+        r = get_redis_client()
+        if r and not r.set(f"ws:jti:{jti}", "1", nx=True, ex=120):
+            logger.warning("WebSocket token replay detected: jti=%s", jti)
+            return None
```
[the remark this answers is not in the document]
OlivierTrudeau: no I fixed it

PR #282, opened 2026-04-18
Add MCP service to airtight, prebuilt, and Kubernetes deployment paths

damianloch:
## Summary
This PR makes surgical, minimal updates so the Aurora MCP server is consistently included across all deployment variants:

- Added `aurora-mcp` service to `docker-compose.airtight.yml`
- Added explicit `image: aurora_mcp:latest` to `docker-compose.prod-local.yml` MCP service
- Updated prebuilt image tagging flows to alias server image as `aurora_mcp:latest`:
  - `Makefile` (`make prod-prebuilt`)
  - `deploy/vm-deploy.sh` (`--prebuilt` path)
- Updated airtight bundling flows to include MCP image alias:
  - `scripts/package-airtight.sh`
  - `.github/workflows/publish-airtight.yml`
- Added minimal Helm resources for Kubernetes MCP runtime:
  - `deploy/helm/aurora/templates/mcp-deployment.yaml`
  - `deploy/helm/aurora/templates/mcp-service.yaml`
  - `deploy/helm/aurora/templates/serviceaccount.yaml` (new MCP SA)
  - `deploy/helm/aurora/values.yaml` (replica + resources for MCP)

## Why
MCP was available in dev/prod-local source builds, but deployment parity was incomplete for airtight/prebuilt/Kubernetes paths. These changes ensure MCP is present and wired the same way across all app variants with minimal surface-area changes.

## Validation
- Reviewed diffs for all touched deployment paths to ensure consistency with existing naming and patterns.
- Attempted runtime/config validation, but this cloud environment does not have `docker` or `helm` installed (`command not found`), so compose/helm render checks could not be executed here.

<div><a href="https://cursor.com/agents/bc-479e3387-90c3-49af-8699-5f9fd329f8d4"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-479e3387-90c3-49af-8699-5f9fd329f8d4"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

## Summary by CodeRabbit

* **New Features**
  * Added MCP server service to all deployment targets (Kubernetes, Docker Compose, and VM deployments)
  * Service runs on port 8811 with integrated health checks and resource management
  * Configured with automatic startup dependencies and restart policies


deploy/helm/aurora/templates/ingress.yaml
```diff
@@ -74,4 +77,17 @@ spec:
...
+    # MCP Server: mcp.aurora.example.com
+    - host: {{ .Values.ingress.hosts.mcp | quote }}
+      http:
+        paths:
+          - path: /
+            pathType: Prefix
+            backend:
+              service:
+                name: {{ include "aurora.fullname" . }}-mcp
+                port:
```
beng360, line 90: This exposes MCP publicly with zero authentication. The MCP server can execute infra commands, read secrets, run background RCAs — basically full platform access. We need auth middleware or at minimum a note that this MUST be paired with an auth proxy (e.g. oauth2-proxy sidecar or nginx auth_request). Defaulting the host to empty string in values.yaml is a good safety net but someone will set it without thinking.

Can we add an annotation guard here or at least a loud comment that says "do not enable without configuring authentication"?

deploy/helm/aurora/templates/mcp-deployment.yaml
```diff
@@ -0,0 +1,74 @@
...
+            timeoutSeconds: 5
+            failureThreshold: 3
+          livenessProbe:
+            tcpSocket:
+              port: 8811
+            periodSeconds: 20
+            timeoutSeconds: 5
+            failureThreshold: 3
+          resources:
+            {{- toYaml .Values.resources.mcp | nindent 12 }}
```
beng360, line 80: Missing the SSL CA cert volume mount. The server deployment has a conditional `ssl-ca-cert` volume + volumeMount block. Since MCP connects to Postgres through the same DB secrets (envFrom), any deployment using TLS for Postgres will break here because the CA cert isn't mounted. Need the same conditional volume/volumeMount block that server-deployment.yaml has.

server/mcp_server.py
```diff
@@ -156,6 +156,7 @@ async def _api(method: str, path: str, *, params: dict | None = None,
         "Use the curated tools for incidents and infrastructure. "
         "For anything else, use aurora_api -- read aurora://api-catalog first to discover endpoints."
     ),
+    host="0.0.0.0",
```
beng360: Moving `host="0.0.0.0"` into the constructor means it binds to all interfaces even when the module is imported for testing or by another process. The original placement inside `if __name__ == "__main__"` was deliberate — it scoped the bind to direct execution only.

Is there a reason this needed to move? If FastMCP needs it at init time for the new version, fine, but if not I'd keep the original placement.

docker-compose.prod-local.yml
```diff
@@ -482,6 +482,7 @@ services:
       context: ./server
       dockerfile: Dockerfile
       target: prod
+    image: aurora_mcp:latest
     container_name: aurora-mcp
     logging: *default-logging
```
beng360, line 487: Having both `build:` and `image: aurora_mcp:latest` on the same service means `docker compose build` tags the result as `aurora_mcp:latest`, which stomps the prebuilt tag from `make prod-prebuilt`. This creates a subtle ordering dependency — if someone runs `docker compose build` after pulling prebuilt images, the prebuilt MCP image gets overwritten with a local build.

The other services (celery-worker, celery-beat, chatbot) don't do this — they just use `build:` without an explicit `image:` tag. What's the rationale for diverging here?

deploy/helm/aurora/templates/mcp-deployment.yaml
```diff
@@ -0,0 +1,74 @@
...
+      {{- end }}
+      {{- include "aurora.scheduling" (dict "service" "mcp" "global" $) | nindent 6 }}
+      containers:
+        - name: aurora-mcp
+          image: "{{ .Values.image.registry }}/aurora-server:{{ .Values.image.tag }}"
+          imagePullPolicy: IfNotPresent
+          command: ["python", "mcp_server.py"]
+          securityContext:
+            {{- include "aurora.containerSecurityContext" (dict "service" "mcp" "global" $ "defaults" (dict)) | nindent 12 }}
+          ports:
```
beng360, line 44: Hardcoding `imagePullPolicy: IfNotPresent` diverges from the other deployments which use `{{ .Values.image.pullPolicy }}`. If someone sets pullPolicy to Always in values (common for staging envs), MCP won't respect it.

docker-compose.airtight.yml
```diff
@@ -425,6 +425,30 @@ services:
...
+    image: aurora_mcp:latest
+    pull_policy: never
+    container_name: aurora-mcp
+    logging: *default-logging
+    ports:
+      - "${MCP_PORT:-8811}:8811"
+    command: ["python", "mcp_server.py"]
+    environment:
+      <<: *common-env
+      OTEL_SERVICE_NAME: aurora-mcp
```
beng360, line 438: Airtight is an airgapped environment with no network access. Does MCP actually function without outbound connectivity? If any of its tools try to reach external APIs (incident.io, GitHub, etc.) they'll silently fail. Not a blocker but worth a comment or env var to disable external-dependent tools in airtight mode.

deploy/helm/aurora/values.yaml
```diff
@@ -642,6 +650,7 @@ ingress:
     frontend: "aurora.example.com"
     api: "api.aurora.example.com"
     ws: "ws.aurora.example.com"
+    mcp: ""  # MCP server (optional) - e.g., mcp.aurora.example.com
```
beng360: The empty string default is good for safety, but the comment should be stronger. Something like:

`mcp: ""  # WARNING: Enabling this exposes MCP externally. Requires authentication proxy.`

Right now "optional" makes it sound like a nice-to-have toggle, not "this gives unauthenticated access to your entire platform if you set it without auth".

beng360 (changes requested):
A few things need addressing before this is mergeable — mainly the ingress security concern and the missing volume mounts on the Helm deployment.

deploy/helm/aurora/values.yaml
damianloch: Fixed. Updated to: `mcp: ""  # WARNING: Requires auth proxy (e.g. oauth2-proxy). Do not enable without authentication.`

deploy/helm/aurora/templates/ingress.yaml
damianloch: Good catch. Added a warning comment in the ingress template: "WARNING: pair with an auth proxy; unauthenticated access grants full platform control". Also strengthened the values.yaml comment as noted in the other thread.

deploy/helm/aurora/templates/mcp-deployment.yaml
damianloch: Good catch — added the conditional ssl-ca-cert volume and volumeMount blocks matching the server-deployment.yaml pattern. Any deployment using TLS for Postgres will now have the CA cert available.

server/mcp_server.py
damianloch: This is intentional — binding 0.0.0.0 is required for container networking (both Docker and K8s). DNS rebinding protection with localhost-only would break all non-local access. Auth is enforced via Bearer token (MCP_AUTH_TOKEN). Added an inline comment clarifying this.

deploy/helm/aurora/templates/mcp-deployment.yaml
damianloch: Every deployment in the chart hardcodes `imagePullPolicy: IfNotPresent` — there is no `.Values.image.pullPolicy` field. MCP is consistent with the existing pattern (checked all 13 deployments/statefulsets). If we want to templatize this it should be a separate PR across all services.

docker-compose.airtight.yml
damianloch: MCP proxies to the local Aurora API server, so core functionality works without outbound access. Tools that call external APIs (incident.io, GitHub, etc.) will gracefully fail. Added `AURORA_AIRGAPPED: "true"` env var to the airtight service definition with a comment noting this.

docker-compose.prod-local.yml
damianloch: This is actually consistent with every other built service in the file. All of them have both `build:` and `image:`:

- `aurora-server`: `build:` + `image: aurora_server:latest`
- `celery_worker`: `build:` + `image: aurora_celery-worker:latest`
- `celery_beat`: `build:` + `image: aurora_celery-beat:latest`
- `chatbot`: `build:` + `image: aurora_chatbot:latest`
- `frontend`: `build:` + `image: aurora_frontend:latest`
- `aurora-mcp`: `build:` + `image: aurora_mcp:latest`

The `image:` tag is what `make prod-prebuilt` targets with `docker tag`, and `docker compose up` will use the pre-existing image if it finds one matching the tag (skipping build). This is the established pattern across all services.

server/mcp_server.py
beng360: ok

docker-compose.prod-local.yml
beng360: sounds good

deploy/helm/aurora/templates/mcp-deployment.yaml
beng360: fair

docker-compose.airtight.yml
beng360: perfect

beng360:
## Documentation gaps

This PR adds the MCP server + multi-platform deployment but the docs don't cover it yet:

### Missing entirely
- **MCP client setup guide** (`/website/docs/integrations/mcp.md`) — how to connect Cursor, Claude Desktop, Windsurf to Aurora's MCP server. Token generation, endpoint config, available tools, example usage.
- **Environment variable reference** — MCP_PORT, replica count, auth token config not documented in `/website/docs/configuration/environment.md`.
- **README mention** — MCP isn't referenced in the main README as an integration point.

### Existing docs need updates
- `/website/docs/deployment/kubernetes.md` — no mention of `mcp-deployment.yaml`, `mcp-service.yaml`, port 8811, or the optional MCP ingress host.
- `/website/docs/deployment/docker-compose.md` — MCP service added to `docker-compose.airtight.yml` but not documented.
- Sidebar (`/website/sidebars.ts`) needs an entry for the new MCP page.

### Security/architecture
- No documentation on the MCP auth model (bearer token resolved via direct DB access, not Flask API). Should be explicitly called out for security review and operator understanding.
- No guidance on when to expose MCP externally (ingress) vs keep it cluster-internal only.

These should ship with the PR or as a fast-follow — an undocumented MCP endpoint with token auth is a support burden waiting to happen.

beng360 (approved):

PR #295, opened 2026-04-22
fix(DEV-1064): SonarQube medium reliability findings

isiddharthsingh:
## Summary
- Fixes 67 medium-severity reliability findings across Python, TypeScript, and k8s Helm templates, grouped one commit per rule.
- 31 additional findings audited and marked `accepted` in Sonar (token-compliance / active-regression / deterministic-return cases); no code change.
- Notable: removed an unreachable duplicate `elif event_type == "values"` in `main_chatbot.py` that was shadowing the real state-update handler.

## Rules addressed
- `python:S3923` — collapsed identical conditional branches
- `python:S2583` — dropped always-true fallback on terraform placeholder check
- `python:S1862` — removed unreachable duplicate elif
- `typescript:S6440` — moved early-return guards past hook declarations (Rules of Hooks)
- `typescript:S7773` — `Number.isNaN` / `Number.parseInt` over globals
- `kubernetes:S6897` — ephemeral-storage requests/limits on services with real disk usage (server, celeryWorker, frontend, kubectl-agent)
- `typescript:S6853 + S6848` — label `htmlFor`+`id` and `role="button"`+`tabIndex`+`onKeyDown` on non-native interactive elements

## Test plan
- [ ] `pytest tests/services/correlation/` — 59/59 pass
- [ ] `tsc --noEmit` — baseline error count unchanged
- [ ] `helm template deploy/helm/aurora` renders; ephemeral-storage only on server/celeryWorker/frontend/mcp/memgraph
- [ ] Send a chat message and verify streaming works end-to-end (covers the `main_chatbot.py` dead-elif removal)
- [ ] Keyboard-tab through flagged UI elements; Enter/Space activates them
- [ ] Open an incident in merged/running/complete states; ThoughtsPanel and RecentAlertsSection render correctly

### DIFF
--- a/server/main_chatbot.py
+++ b/server/main_chatbot.py
@@ -573,12 +573,6 @@ async def process_workflow_async(wf, state, websocket, user_id, incident_id=None
                                 save_incident_thought(token_text, force=False)
                                 save_streaming_chat_message(token_text, force=False)
                     
-                    elif event_type == "values":
-                        # Final state update
-                        logger.debug(f"[STREAM DEBUG] Received values event")
-                        # Handle final state if needed
-                        continue
-                    
                     elif event_type == "messages":
                         try:
                             msg_chunk, _ = event_data
--- a/server/main_chatbot.py
+++ b/server/main_chatbot.py
@@ -572,7 +572,15 @@ async def process_workflow_async(wf, state, websocket, user_id, incident_id=None
                                 # Save tokens incrementally to incident thoughts
                                 save_incident_thought(token_text, force=False)
                                 save_streaming_chat_message(token_text, force=False)
-                    
+
+                    elif event_type == "values":
+                        # Stub: skip complete-state "values" events.
+                        # The real handler below would re-send content already streamed
+                        # via the "messages" event path, causing duplicates in the UI / DB.
+                        # Keep this branch first so the duplicate-send path is unreachable.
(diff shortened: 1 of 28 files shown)


isiddharthsingh:
@claude review this

server/main_chatbot.py
```diff
@@ -573,12 +573,6 @@ async def process_stream():
                                 save_incident_thought(token_text, force=False)
                                 save_streaming_chat_message(token_text, force=False)
                     
-                    elif event_type == "values":
-                        # Final state update
-                        logger.debug(f"[STREAM DEBUG] Received values event")
-                        # Handle final state if needed
-                        continue
```
OlivierTrudeau, line 580: This activates the elif values branch that was being skipped with this one. Will need to be careful as to how this influences chats. This might duplicate the messages in db or on user's ui. Values comes in when a block finishes streaming with the whole content i think so this is why it might have been skipped in the past because we were building dynamically

OlivierTrudeau (commented):
just check the comment I amde + fix the sonarqube issues

server/main_chatbot.py
isiddharthsingh: good catch, you're right. reverted the removal in c91e0fac — put the continue stub back with a comment explaining why it needs to stay (otherwise the real handler below re-sends content that was already streamed chunk-by-chunk via the messages event, which would double up in the ui and db). also marked the sonarqube s1862 issue as accepted so it stops flagging.

PR #298, opened 2026-04-23
fix(bitbucket): migrate to /2.0/user/workspaces (CHANGE-2770)

isiddharthsingh:
## Summary
- Bitbucket removed `/2.0/workspaces` on 2026-04-14 (CHANGE-2770, returns HTTP 410), which broke the workspace dropdown in the connectors UI with `workspaces.map is not a function`.
- Switched `BitbucketAPIClient.get_workspaces()` to the CHANGE-3022 replacement `/2.0/user/workspaces` and unwrapped `values[].workspace` so the existing list-of-workspaces contract is preserved.
- `/2.0/user/permissions/workspaces` was also killed by the same change, so it's explicitly ruled out in the docstring.

## Test plan
- [x] Verified live against Bitbucket: `/2.0/user/workspaces` returns 200 with the expected shape
- [x] Reloaded `/connectors` — workspace dropdown populates, repo + branch selection saves end-to-end (`Saved Bitbucket workspace selection … arvoai / test / main`)
- [x] Probed all other workspace- and user-scoped endpoints the client uses (`/workspaces/{ws}`, `/repositories/{ws}/...`, `/workspaces/{ws}/search/code`, etc.) — all 200, none affected by CHANGE-2770

## Summary by CodeRabbit

* **Bug Fixes**
  * Bitbucket workspace retrieval updated to use the correct integration endpoint.
  * Improved handling of paginated responses and error payloads for more reliable syncs.
  * Workspace results are now flattened and cleaned of invalid entries, yielding a more consistent list for display.

### DIFF
--- a/server/connectors/bitbucket_connector/api_client.py
+++ b/server/connectors/bitbucket_connector/api_client.py
@@ -205,8 +205,16 @@ class BitbucketAPIClient:
     # ------------------------------------------------------------------
 
     def get_workspaces(self):
-        """List all workspaces the authenticated user has access to."""
-        return self._paginated_get(f"{BITBUCKET_API_BASE}/workspaces")
+        """List all workspaces the authenticated user has access to.
+
+        Uses ``/2.0/user/workspaces``. The legacy ``/2.0/workspaces`` and
+        ``/2.0/user/permissions/workspaces`` endpoints were removed in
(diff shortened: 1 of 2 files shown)


OlivierTrudeau (approved):

Zarlanx (approved):
lgtm

PR #302, opened 2026-04-23
fix(security): resolve all 202 open SonarQube security findings

isiddharthsingh:
## Summary

- **187 × S5145 (log injection)**: wrap user-controlled log arguments with `sanitize()` from `utils.log_sanitizer` across 56 files
- **2 × S7044 (SSRF path traversal)**: URL-encode every path segment in `bitbucket_connector/api_client.py` via `quote(..., safe='')` and reject standalone `..` segments in `_validate_bitbucket_url`
- **13 × S8233 / S8264 (GH Actions perms)**: move `permissions:` blocks from workflow level to the specific jobs that need them in 5 workflow files

## Test plan

- [x] All 62 modified Python files `py_compile` clean
- [x] All 5 modified YAML workflows parse clean
- [x] Live Bitbucket smoke against arvoai workspace: `get_workspaces` / `get_workspace` / `get_repositories` / `get_diff base..head`
- [x] Path-traversal validator: bare `../` rejected; `quote()`-encoded slug stays as one safe segment
- [x] Per-job GH Actions perm audit against every action in every workflow — no missing / no excess
- [x] Log-scan after live calls: zero format-string / sanitize exceptions across server + celery
- [x] Direct invocation of heavily-modified logger lines in `connection_utils`, `limiter_ext`, `ssh_utils`, `token_management` — all render cleanly
- [ ] Sonar re-scan: confirm 202 findings close (mark any re-raised S7044 false-positive with `_validate_bitbucket_url` as justification)
- [ ] Post-merge: first run of `publish-images.yml`, `publish-helm.yml`, `publish-airtight.yml`, `docs.yml` succeeds (watch for `Resource not accessible by integration`)

## Summary by CodeRabbit

* **Security**
  * Reduced CI token privileges by moving to least-privilege, job-scoped permissions.
  * Strengthened logging to redact or hash identifiers and sensitive values before they are emitted.
  * Hardened Bitbucket integration to better validate and encode repository paths and diff specs, preventing unsafe path usage.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,13 +6,12 @@ on:
   # pull_request:
   #   branches: [ main, release ]
 
-permissions:
-  contents: read
-
 jobs:
   build-dev:
     name: Dev - ${{ matrix.service }}
     runs-on: ubuntu-24.04
+    permissions:
+      contents: read
     # Only run builds on the main repository, not on forks
     if: github.event.pull_request.head.repo.full_name == github.repository || github.event_name != 'pull_request'
     environment: ${{ github.ref == 'refs/heads/main' && 'staging' || (startsWith(github.ref, 'refs/heads/release/') && 'production' || 'development') }}
@@ -95,6 +94,8 @@ jobs:
     # Only run builds on the main repository, not on forks
     if: github.event.pull_request.head.repo.full_name == github.repository || github.event_name != 'pull_request'
     environment: ${{ github.ref == 'refs/heads/main' && 'staging' || (startsWith(github.ref, 'refs/heads/release/') && 'production' || 'development') }}
+    permissions:
+      contents: read
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -174,6 +174,7 @@ jobs:
     runs-on: ubuntu-24.04
     needs: [build-dev, build-prod]
     if: always()
+    permissions: {}
     steps:
       - name: Check build results
         run: |
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -10,11 +10,6 @@ on:
       - '.github/workflows/docs.yml'
   workflow_dispatch:
 
-permissions:
-  contents: read
-  pages: write
-  id-token: write
-
 concurrency:
   group: pages
   cancel-in-progress: false
@@ -22,6 +17,8 @@ concurrency:
 jobs:
   build:
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
     steps:
       - name: Checkout
         uses: actions/checkout@v4
@@ -53,6 +50,9 @@ jobs:
       url: ${{ steps.deployment.outputs.page_url }}
     runs-on: ubuntu-latest
     needs: build
+    permissions:
+      pages: write
+      id-token: write
     steps:
       - name: Deploy to GitHub Pages
         id: deployment
--- a/.github/workflows/publish-airtight.yml
+++ b/.github/workflows/publish-airtight.yml
@@ -182,6 +182,7 @@ jobs:
     runs-on: ubuntu-24.04
     needs: [airtight]
     if: always()
+    permissions: {}
     steps:
       - name: Summary
         run: |
--- a/.github/workflows/publish-airtight.yml
+++ b/.github/workflows/publish-airtight.yml
@@ -14,17 +14,16 @@ concurrency:
   group: publish-airtight-${{ github.ref }}
   cancel-in-progress: true
 
-permissions:
-  contents: read
-  id-token: write
-  attestations: write
-  artifact-metadata: write
-
 jobs:
   airtight:
     name: Airtight ${{ matrix.arch }}
     runs-on: ${{ matrix.runner }}
     if: github.repository == 'Arvo-AI/aurora'
+    permissions:
+      contents: read
+      id-token: write
+      attestations: write
+      artifact-metadata: write
 
     strategy:
       fail-fast: false
--- a/.github/workflows/publish-helm.yml
+++ b/.github/workflows/publish-helm.yml
@@ -10,11 +10,6 @@ concurrency:
   group: publish-helm-${{ github.ref }}
   cancel-in-progress: true
 
-permissions:
-  contents: write   # GitHub Pages release assets
-  packages: write   # GHCR OCI push
-  pages: write      # gh-pages branch
-
 env:
   CHART_DIR: deploy/helm
   OCI_REGISTRY: ghcr.io
@@ -55,7 +50,9 @@ jobs:
     runs-on: ubuntu-24.04
     needs: [validate-version]
     if: github.repository == 'Arvo-AI/aurora' && needs.validate-version.result == 'success'
-
+    permissions:
+      contents: write   # GitHub Pages release assets + gh-pages branch
+      pages: write
     steps:
       - name: Checkout repository
         uses: actions/checkout@v5
@@ -87,6 +84,9 @@ jobs:
     runs-on: ubuntu-24.04
     needs: [validate-version]
     if: github.repository == 'Arvo-AI/aurora' && needs.validate-version.result == 'success'
+    permissions:
+      contents: read
+      packages: write   # GHCR OCI push
     outputs:
       chart_version: ${{ steps.package.outputs.chart_version }}
 
--- a/.github/workflows/publish-helm.yml
+++ b/.github/workflows/publish-helm.yml
@@ -127,6 +127,7 @@ jobs:
     runs-on: ubuntu-24.04
     needs: [publish-ghpages, publish-oci]
     if: always()
+    permissions: {}
     steps:
       - name: Summary
         run: |
--- a/.github/workflows/publish-helm.yml
+++ b/.github/workflows/publish-helm.yml
@@ -23,6 +23,8 @@ jobs:
     name: Validate chart version matches tag
     runs-on: ubuntu-24.04
     if: github.repository == 'Arvo-AI/aurora' && startsWith(github.ref, 'refs/tags/v')
+    permissions:
+      contents: read
     steps:
       - name: Checkout repository
         uses: actions/checkout@v5
--- a/.github/workflows/publish-images.yml
+++ b/.github/workflows/publish-images.yml
@@ -225,6 +225,7 @@ jobs:
     runs-on: ubuntu-24.04
     needs: [build, merge]
     if: always()
+    permissions: {}
     steps:
       - name: Summary
         run: |
--- a/.github/workflows/publish-images.yml
+++ b/.github/workflows/publish-images.yml
@@ -17,15 +17,14 @@ concurrency:
 env:
   REGISTRY: ghcr.io
 
-permissions:
-  contents: read
-  packages: write
-
 jobs:
   build:
     name: Build ${{ matrix.image }} (${{ matrix.platform }})
     if: github.repository == 'Arvo-AI/aurora'
     runs-on: ${{ matrix.runner }}
+    permissions:
+      contents: read
+      packages: write
     strategy:
       fail-fast: false
       matrix:
@@ -139,6 +138,9 @@ jobs:
     needs: build
     if: github.repository == 'Arvo-AI/aurora'
     runs-on: ubuntu-24.04
+    permissions:
+      contents: read
+      packages: write
     strategy:
       fail-fast: false
       matrix:
--- a/server/connectors/bitbucket_connector/api_client.py
+++ b/server/connectors/bitbucket_connector/api_client.py
@@ -22,9 +22,11 @@ def _validate_bitbucket_url(url: str) -> None:
         raise ValueError(f"URL scheme '{parts.scheme}' is not allowed; only HTTPS is permitted")
     if parts.hostname not in _BITBUCKET_ALLOWED_HOSTS:
         raise ValueError(f"URL host '{parts.hostname}' is not a known Bitbucket domain")
-    # Reject path traversal. Decode each segment to a fixed point so that
-    # doubly-encoded forms like ``%252e%252e`` (which downstream clients may
-    # re-decode into ``..``) are also rejected.
+    # Reject path traversal. Decode each segment to a fixed point (to catch
+    # doubly-encoded forms like ``%252e%252e``) AND re-split on "/" afterwards,
+    # because ``%2f`` inside a single segment decodes to a slash that would
+    # otherwise smuggle a ".." component past a naive whole-segment equality
+    # check (e.g. ``%2e%2e%2fsecret`` is one segment but decodes to ``../secret``).
     for segment in parts.path.split("/"):
         decoded = segment
         for _ in range(3):
@@ -32,7 +34,7 @@ def _validate_bitbucket_url(url: str) -> None:
             decoded = unquote(decoded)
             if decoded == previous:
                 break
-        if decoded == "..":
+        if any(sub == ".." for sub in decoded.split("/")):
             raise ValueError("URL path contains a traversal segment")
 
 
@@ -395,7 +397,13 @@ class BitbucketAPIClient:
         if ".." in spec:
             parts = spec.split("..")
             if len(parts) != 2:
-                raise ValueError("diff spec must contain exactly one '..' separator")
+                # Match the rest of this client's error contract (callers use
+                # forward_if_error on the result rather than try/except).
+                return {
+                    "error": True,
+                    "status": None,
+                    "message": "diff spec must contain exactly one '..' separator",
+                }
             base, head = parts
             encoded_spec = f"{quote(base, safe='')}..{quote(head, safe='')}"
         else:
--- a/server/connectors/bitbucket_connector/api_client.py
+++ b/server/connectors/bitbucket_connector/api_client.py
@@ -22,11 +22,17 @@ def _validate_bitbucket_url(url: str) -> None:
         raise ValueError(f"URL scheme '{parts.scheme}' is not allowed; only HTTPS is permitted")
     if parts.hostname not in _BITBUCKET_ALLOWED_HOSTS:
         raise ValueError(f"URL host '{parts.hostname}' is not a known Bitbucket domain")
-    # Reject path traversal: inspect each decoded path segment so that `..`
-    # embedded inside a legitimate segment (e.g. a diff spec like `abc..def`)
-    # is allowed, while a standalone `..` segment is rejected.
+    # Reject path traversal. Decode each segment to a fixed point so that
+    # doubly-encoded forms like ``%252e%252e`` (which downstream clients may
+    # re-decode into ``..``) are also rejected.
     for segment in parts.path.split("/"):
-        if unquote(segment) == "..":
+        decoded = segment
+        for _ in range(3):
+            previous = decoded
+            decoded = unquote(decoded)
+            if decoded == previous:
+                break
+        if decoded == "..":
             raise ValueError("URL path contains a traversal segment")
 
 
@@ -387,7 +393,10 @@ class BitbucketAPIClient:
         ``..`` separator is preserved as a single path segment.
         """
         if ".." in spec:
-            base, _, head = spec.partition("..")
+            parts = spec.split("..")
+            if len(parts) != 2:
+                raise ValueError("diff spec must contain exactly one '..' separator")
+            base, head = parts
             encoded_spec = f"{quote(base, safe='')}..{quote(head, safe='')}"
         else:
             encoded_spec = quote(spec, safe='')
--- a/server/connectors/bitbucket_connector/api_client.py
+++ b/server/connectors/bitbucket_connector/api_client.py
@@ -4,7 +4,7 @@ Wraps the Bitbucket 2.0 REST API with authentication and pagination support.
 """
 import base64
 import logging
-from urllib.parse import quote, urlsplit
+from urllib.parse import quote, unquote, urlsplit
 
 import requests
 
@@ -15,12 +15,19 @@ _BITBUCKET_ALLOWED_HOSTS = frozenset({"api.bitbucket.org", "bitbucket.org"})
 
 
 def _validate_bitbucket_url(url: str) -> None:
-    """Raise ValueError if url does not point to a known Bitbucket host over HTTPS."""
+    """Raise ValueError if url does not point to a known Bitbucket host over HTTPS,
+    or if the path contains traversal segments (``..``)."""
     parts = urlsplit(url)
     if parts.scheme != "https":
         raise ValueError(f"URL scheme '{parts.scheme}' is not allowed; only HTTPS is permitted")
     if parts.hostname not in _BITBUCKET_ALLOWED_HOSTS:
         raise ValueError(f"URL host '{parts.hostname}' is not a known Bitbucket domain")
+    # Reject path traversal: inspect each decoded path segment so that `..`
+    # embedded inside a legitimate segment (e.g. a diff spec like `abc..def`)
+    # is allowed, while a standalone `..` segment is rejected.
+    for segment in parts.path.split("/"):
+        if unquote(segment) == "..":
+            raise ValueError("URL path contains a traversal segment")
 
 
 def _sanitize_url(url: str) -> str:
@@ -210,19 +217,26 @@ class BitbucketAPIClient:
 
     def get_workspace(self, workspace):
         """Get a single workspace by slug."""
-        return self._get(f"{BITBUCKET_API_BASE}/workspaces/{workspace}")
+        return self._get(f"{BITBUCKET_API_BASE}/workspaces/{quote(workspace, safe='')}")
 
     def get_projects(self, workspace):
         """List projects in a workspace."""
-        return self._paginated_get(f"{BITBUCKET_API_BASE}/workspaces/{workspace}/projects")
+        return self._paginated_get(
+            f"{BITBUCKET_API_BASE}/workspaces/{quote(workspace, safe='')}/projects"
+        )
 
     def get_repositories(self, workspace):
         """List repositories in a workspace."""
-        return self._paginated_get(f"{BITBUCKET_API_BASE}/repositories/{workspace}")
+        return self._paginated_get(
+            f"{BITBUCKET_API_BASE}/repositories/{quote(workspace, safe='')}"
+        )
 
     def get_repository(self, workspace, repo_slug):
         """Get a single repository."""
-        return self._get(f"{BITBUCKET_API_BASE}/repositories/{workspace}/{repo_slug}")
+        return self._get(
+            f"{BITBUCKET_API_BASE}/repositories/"
+            f"{quote(workspace, safe='')}/{quote(repo_slug, safe='')}"
+        )
 
     # ------------------------------------------------------------------
     # File / Directory / Code Search
@@ -264,7 +278,10 @@ class BitbucketAPIClient:
 
     def create_or_update_file(self, workspace, repo_slug, path, content, message, branch, author=None):
         """Create or update a file via multipart form POST to /src."""
-        url = f"{BITBUCKET_API_BASE}/repositories/{workspace}/{repo_slug}/src"
+        url = (
+            f"{BITBUCKET_API_BASE}/repositories/"
+            f"{quote(workspace, safe='')}/{quote(repo_slug, safe='')}/src"
+        )
         form_data = {
             "message": message,
             "branch": branch,
@@ -276,7 +293,10 @@ class BitbucketAPIClient:
 
     def delete_file(self, workspace, repo_slug, path, message, branch):
         """Delete a file via POST to /src with files param."""
-        url = f"{BITBUCKET_API_BASE}/repositories/{workspace}/{repo_slug}/src"
+        url = (
+            f"{BITBUCKET_API_BASE}/repositories/"
+            f"{quote(workspace, safe='')}/{quote(repo_slug, safe='')}/src"
+        )
         form_data = {
             "message": message,
             "branch": branch,
@@ -297,7 +317,7 @@ class BitbucketAPIClient:
 
     def search_code(self, workspace, query):
         """Search code across a workspace."""
-        url = f"{BITBUCKET_API_BASE}/workspaces/{workspace}/search/code"
+        url = f"{BITBUCKET_API_BASE}/workspaces/{quote(workspace, safe='')}/search/code"
         return self._get(url, params={"search_query": query})
 
     # ------------------------------------------------------------------
@@ -307,12 +327,16 @@ class BitbucketAPIClient:
     def get_branches(self, workspace, repo_slug):
         """List branches for a repository."""
         return self._paginated_get(
-            f"{BITBUCKET_API_BASE}/repositories/{workspace}/{repo_slug}/refs/branches"
+            f"{BITBUCKET_API_BASE}/repositories/"
+            f"{quote(workspace, safe='')}/{quote(repo_slug, safe='')}/refs/branches"
         )
 
     def create_branch(self, workspace, repo_slug, name, target_hash):
         """Create a new branch from a target commit hash."""
-        url = f"{BITBUCKET_API_BASE}/repositories/{workspace}/{repo_slug}/refs/branches"
+        url = (
+            f"{BITBUCKET_API_BASE}/repositories/"
+            f"{quote(workspace, safe='')}/{quote(repo_slug, safe='')}/refs/branches"
+        )
         return self._post(url, json_data={
             "name": name,
             "target": {"hash": target_hash},
@@ -320,7 +344,11 @@ class BitbucketAPIClient:
 
     def delete_branch(self, workspace, repo_slug, name):
         """Delete a branch."""
-        url = f"{BITBUCKET_API_BASE}/repositories/{workspace}/{repo_slug}/refs/branches/{name}"
+        url = (
+            f"{BITBUCKET_API_BASE}/repositories/"
+            f"{quote(workspace, safe='')}/{quote(repo_slug, safe='')}"
+            f"/refs/branches/{quote(name, safe='')}"
+        )
         return self._delete(url)
 
     # ------------------------------------------------------------------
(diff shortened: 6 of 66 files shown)


server/routes/aws/onboarding.py
```diff
@@ -104,11 +105,11 @@ def get_aws_onboarding_links(user_id, workspace_id):
         if aws_conn and aws_conn.get('role_arn'):
             response_data["roleArn"] = aws_conn['role_arn']
         
-        logger.info(f"Retrieved AWS onboarding info for workspace {workspace_id}")
+        logger.info(f"Retrieved AWS onboarding info for workspace {sanitize(workspace_id)}")
         return jsonify(response_data)
         
     except Exception as e:
-        logger.error(f"Failed to get AWS onboarding info for workspace {workspace_id}: {e}")
+        logger.error(f"Failed to get AWS onboarding info for workspace {sanitize(workspace_id)}: {sanitize(e)}")
```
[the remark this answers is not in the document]
isiddharthsingh: Not fixed — false positive. `sanitize(e)` is safe because `utils.log_sanitizer.sanitize` internally calls `str(value)` first, so it handles Exception objects correctly. Switching to `exc_info=True` would change the log format and isn't needed.

server/routes/aws/onboarding.py
```diff
@@ -197,8 +198,8 @@ def set_aws_role(user_id, workspace_id):
             except Exception as read_only_error:
                 logger.warning(
                     "Read-only role validation failed for workspace %s using %s: %s",
-                    workspace_id,
-                    read_only_role_arn,
+                    sanitize(workspace_id),
+                    sanitize(read_only_role_arn),
                     read_only_error,
                 )
```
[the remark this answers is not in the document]
isiddharthsingh: Fixed in d00b9925 — wrapped `read_only_error` with `sanitize()` on line 204 since AWS API error messages can echo attacker-controlled content.

server/routes/bitbucket/bitbucket.py
```diff
@@ -122,7 +123,7 @@ def bitbucket_callback():
                 frontend_url=FRONTEND_URL,
             )
 
-        logger.info(f"Received Bitbucket code: {code[:5]}...")
+        logger.info(f"Received Bitbucket code: {sanitize(code)[:5]}...")
```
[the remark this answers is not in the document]
isiddharthsingh: Not fixed — kept `sanitize(code)[:5]` intentionally. The OAuth code is single-use and short-lived; logging a 5-char sanitized prefix is a defensive audit trail that's harmless even if control-chars were in the input.

server/routes/datadog/datadog_routes.py
```diff
@@ -303,7 +304,7 @@ def connect(user_id):
 
     masked_api = mask_credential_value(api_key)
     masked_app = mask_credential_value(app_key)
-    logger.info("[DATADOG] Connecting user %s to site=%s api=%s app=%s", user_id, site, masked_api, masked_app)
+    logger.info("[DATADOG] Connecting user %s to site=%s api=%s app=%s", sanitize(user_id), sanitize(site), masked_api, masked_app)
```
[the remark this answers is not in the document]
isiddharthsingh: Not fixed — `mask_credential_value()` replaces all but ~4 chars with `*`; logging that last fragment is an industry-standard audit pattern (Stripe, Datadog, Auth0 all do this). Removing it entirely would hurt post-hoc incident investigation.

server/routes/grafana/grafana_routes.py
```diff
@@ -115,20 +116,20 @@ def alert_webhook(user_id: str):
     if not row_exists or (row_exists and not is_active):
         skip_rca = True
         if not row_exists:
-            logger.info("[GRAFANA] Auto-connecting user %s via webhook", user_id)
+            logger.info("[GRAFANA] Auto-connecting user %s via webhook", sanitize(user_id))
             try:
                 store_tokens_in_db(user_id, {}, "grafana")
             except Exception as exc:
-                logger.exception("[GRAFANA] Failed to auto-connect user %s: %s", user_id, exc)
+                logger.exception("[GRAFANA] Failed to auto-connect user %s: %s", sanitize(user_id), exc)
```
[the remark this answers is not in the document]
isiddharthsingh: Fixed in d00b9925 — removed the duplicate `exc` argument from `logger.exception`; `exc_info` is now implicit and the log no longer logs the exception twice.

server/routes/incident_feedback/routes.py
```diff
@@ -295,7 +296,7 @@ def set_aurora_learn_setting(user_id):
 
     try:
         store_user_preference(user_id, AURORA_LEARN_PREFERENCE_KEY, enabled)
-        logger.info("[AURORA LEARN] User %s set Aurora Learn to %s", user_id, enabled)
+        logger.info("[AURORA LEARN] User %s set Aurora Learn to %s", sanitize(user_id), enabled)
         return jsonify({"success": True, "enabled": enabled}), 200
     except Exception as exc:
         logger.exception("[AURORA LEARN] Failed to update setting for user %s", user_id)
```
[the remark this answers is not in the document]
isiddharthsingh: Fixed in 17376e9d — wrapped `user_id` in the `logger.exception` call at line 302.

server/routes/newrelic/newrelic_routes.py
```diff
@@ -387,7 +388,7 @@ def webhook(user_id: str):
 
     logger.info(
         "[NEWRELIC][WEBHOOK] Received alert for user %s: %s (issue=%s)",
-        user_id, title, issue_id,
+        sanitize(user_id), sanitize(title), sanitize(issue_id),
     )
```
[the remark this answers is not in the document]
isiddharthsingh: Not fixed — false positive. Our `sanitize()` helper (in `utils/log_sanitizer.py`) calls `str(value)` internally, so `sanitize(None)` returns `"None"` and `sanitize(42)` returns `"42"` — it can't raise on non-string input.

server/routes/opsgenie/opsgenie_routes.py
```diff
@@ -642,15 +643,15 @@ def webhook(user_id: str):
...
+        logger.warning("[OPSGENIE] Webhook received for user %s with no OpsGenie connection", sanitize(user_id))
         return jsonify({"error": "OpsGenie not connected for this user"}), 404
 
     payload = request.get_json(silent=True) or {}
     metadata = {
         "headers": dict(request.headers),
         "remote_addr": request.remote_addr,
     }
-    logger.info("[OPSGENIE] Received webhook for user %s action=%s", user_id, payload.get("action"))
+    logger.info("[OPSGENIE] Received webhook for user %s action=%s", sanitize(user_id), sanitize(payload.get("action")))
```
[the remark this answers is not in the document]
isiddharthsingh: Not fixed — same as newrelic/opsgenie pattern above: our `sanitize()` helper coerces any type via `str()`, so `sanitize(None)` is safe and returns `"None"`.

server/routes/org_routes.py
```diff
@@ -115,7 +116,7 @@ def _delete_user_org_data(cursor, user_id: str, org_id: str):
         try:
             cursor.execute(f'DELETE FROM "{tbl}" WHERE user_id = %s AND org_id = %s', (user_id, org_id))
         except Exception as e:
-            logger.warning("Failed to delete %s data for user %s in org %s: %s", tbl, user_id, org_id, e)
+            logger.warning("Failed to delete %s data for user %s in org %s: %s", tbl, sanitize(user_id), sanitize(org_id), e)
```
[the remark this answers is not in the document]
isiddharthsingh: Not fixed — kept `sanitize()` on `user_id` / `org_id` / `old_org_id` / `new_org_id`. These IDs originate from a user-supplied JWT and (for `new_org_id`) from `request.get_json()`, so they're not fully server-controlled. Keeping the wrappers consistent with the rest of the 202-finding sweep.

server/routes/pagerduty/pagerduty_routes.py
```diff
@@ -251,18 +252,18 @@ def webhook(user_id: str):
...
     if not creds:
-        logger.warning("[PAGERDUTY] Webhook received for user %s with no PagerDuty connection", user_id)
+        logger.warning("[PAGERDUTY] Webhook received for user %s with no PagerDuty connection", sanitize(user_id))
         return jsonify({"error": "PagerDuty not connected for this user"}), 404
     
     payload = request.get_json(silent=True) or {}
     
     # Log raw payload
-    logger.info("[PAGERDUTY] Raw webhook payload for user %s: %s", user_id, json.dumps(payload))
+    logger.info("[PAGERDUTY] Raw webhook payload for user %s: %s", sanitize(user_id), json.dumps(payload))
```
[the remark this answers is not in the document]
isiddharthsingh: Fixed in `8b25b1e0` — wrapped `json.dumps(payload)` with `sanitize()`. Keeping the log at `info` level (rather than demoting to `debug` or removing) because the PagerDuty raw-payload log is useful for audit/correlation when investigating routing bugs.

server/utils/auth/oauth2_state_cache.py
```diff
@@ -161,7 +162,7 @@ def retrieve_oauth2_state(state: str) -> Optional[Dict[str, Any]]:
         value = results[0]
 
     if not value:
-        logger.warning(f"State token not found in Redis: {state[:10]}... (expired or invalid)")
+        logger.warning(f"State token not found in Redis: {sanitize(state)[:10]}... (expired or invalid)")
```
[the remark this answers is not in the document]
isiddharthsingh: Not fixed — the other `state` log sites in this file (store / success-path debug logs) were not part of the original 202-finding scope. Only the flagged line 165 was touched. Happy to open a separate cleanup PR if a codebase-wide consistency sweep is wanted.

server/utils/auth/stateless_auth.py
```diff
@@ -159,7 +159,7 @@ def get_org_id_from_request() -> Optional[str]:
                         g._org_id_resolved = row[0]
                         return row[0]
         except Exception as e:
-            logger.warning(f"Error looking up org_id for user {user_id}: {e}")
+            logger.warning(f"Error looking up org_id for user {sanitize(user_id)}: {e}")
```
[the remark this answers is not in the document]
isiddharthsingh: Not fixed — this is stylistic (f-string vs. %s lazy substitution). `sanitize(user_id)` is already applied, so the security posture is equivalent either way. Keeping the existing f-string style to match the rest of this file's log calls.

server/utils/aws/aws_sts_client.py
```diff
@@ -123,8 +125,8 @@ def assume_workspace_role(
         
         if external_id != workspace_external_id:
             logger.error(
-                f"SECURITY: External ID mismatch for workspace {workspace_id}. "
-                f"Provided: '{external_id}', Expected: '{workspace_external_id}'. Rejecting role assumption."
+                f"SECURITY: External ID mismatch for workspace {sanitize(workspace_id)}. "
+                f"Provided: '{sanitize(external_id)}', Expected: '{sanitize(workspace_external_id)}'. Rejecting role assumption."
```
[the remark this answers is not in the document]
isiddharthsingh: Not fixed — kept `sanitize()` wrappers. `workspace_id` comes from route params / request body, `role_arn` and `external_id` are stored after user-supplied onboarding values, and `error_code` originates from AWS API responses (attacker-influenceable via crafted inputs to STS). The `workspace_id` check at the top of `assume_workspace_role` validates format but the values still traversed a user-input path, so consistent sanitization is the safer choice and keeps the S5145 sweep uniform.

server/utils/log_sanitizer.py
```diff
@@ -1,10 +1,63 @@
...
+    """Lazily resolve the HMAC key from ``FLASK_SECRET_KEY``.
+
+    Evaluated on first call (not at import) so test collection and non-web
+    entry points that import this module without Flask env vars don't die at
+    import time.
+    """
+    key = os.environ.get("FLASK_SECRET_KEY")
+    if not key:
+        raise RuntimeError("FLASK_SECRET_KEY is required for log-hash correlation")
+    return key.encode("utf-8")
```
Zarlanx, line 45: raising inside a log-hashing helper means a missing FLASK_SECRET_KEY crashes any request whose only sin was logging a user_id; return a sentinel so logging never takes down the request

server/utils/log_sanitizer.py
```diff
@@ -1,10 +1,63 @@
...
 
 _CONTROL_CHARS = re.compile(r"[\x00-\x1f\x7f]")
 
+# Dict-based allowlist so static analyzers can see the return value is either a
+# frozen literal from the set or the sentinel ``"unknown"`` — never the caller's
+# input. This breaks taint flow more unambiguously than a conditional return.
+_SAFE_PROVIDER_LABELS = {name: name for name in KNOWN_PROVIDERS}
+
+
+@lru_cache(maxsize=1)
```
Zarlanx, line 34: lru_cache(maxsize=1) freezes the salt at first call, so rotating FLASK_SECRET_KEY has no effect until every worker restarts

server/utils/log_sanitizer.py
```diff
@@ -1,10 +1,63 @@
...
 
+import hmac
+import os
 import re
+from functools import lru_cache
+from hashlib import sha256
+
+from utils.providers import KNOWN_PROVIDERS
 
 _CONTROL_CHARS = re.compile(r"[\x00-\x1f\x7f]")
```
Zarlanx: this class only covers ASCII control chars; U+2028, U+2029 and zero-width separators still forge new log lines in most parsers

server/utils/log_sanitizer.py
```diff
@@ -1,10 +1,63 @@
...
+
+    The lookup goes through a dict whose keys are the literal set of known
+    providers, so the function either returns one of those literals or the
+    sentinel ``"unknown"`` — never the caller's raw input.
+    """
+    if not value:
+        return "unknown"
+    normalized = _CONTROL_CHARS.sub("", str(value)).strip().lower()
+    return _SAFE_PROVIDER_LABELS.get(normalized, "unknown")
+
```
Zarlanx, line 64: silently lowercasing and stripping means upstream bugs that pass malformed provider strings quietly collapse to unknown instead of being flagged

server/utils/log_sanitizer.py
```diff
@@ -1,10 +1,63 @@
...
+    The default ``length=12`` hex chars (~48 bits) is chosen for correlation
+    only — it's NOT collision-resistant enough for de-duplication or any
+    security-sensitive uniqueness check. Use for identifiers (user_id,
+    account_id, credential prefixes) we want to correlate across logs without
+    exposing raw values.
+    """
+    if value is None or value == "":
+        return "-"
+    text = _CONTROL_CHARS.sub("", str(value)).encode("utf-8")
+    return hmac.new(_get_log_hash_salt(), text, sha256).hexdigest()[:length]
```
Zarlanx: 48 bits of HMAC output is birthday-collision-prone past ~16M values; safe for correlation only as the docstring says, but nothing enforces callers honor that

server/utils/providers.py
```diff
@@ -0,0 +1,54 @@
+"""Canonical provider identifiers used across the codebase.
+
+Keep this in sync with:
+- connector blueprint directories under ``server/routes/`` (CONNECTOR_DIRS)
```
Zarlanx, line 4: the keep-in-sync note is enforced one direction only by test_connector_rbac; a stray provider value written to user_tokens silently collapses to unknown with no test to catch it

server/connectors/bitbucket_connector/api_client.py
```diff
@@ -15,12 +15,19 @@
...
     if parts.scheme != "https":
         raise ValueError(f"URL scheme '{parts.scheme}' is not allowed; only HTTPS is permitted")
     if parts.hostname not in _BITBUCKET_ALLOWED_HOSTS:
         raise ValueError(f"URL host '{parts.hostname}' is not a known Bitbucket domain")
+    # Reject path traversal: inspect each decoded path segment so that `..`
+    # embedded inside a legitimate segment (e.g. a diff spec like `abc..def`)
+    # is allowed, while a standalone `..` segment is rejected.
+    for segment in parts.path.split("/"):
+        if unquote(segment) == "..":
+            raise ValueError("URL path contains a traversal segment")
```
Zarlanx, line 38: unquote(segment) is a single-decode check, so double-encoded %252e%252e slips past and can decode to .. in clients that re-decode

server/connectors/bitbucket_connector/api_client.py
```diff
@@ -15,12 +15,19 @@
...
+    or if the path contains traversal segments (``..``)."""
     parts = urlsplit(url)
     if parts.scheme != "https":
         raise ValueError(f"URL scheme '{parts.scheme}' is not allowed; only HTTPS is permitted")
     if parts.hostname not in _BITBUCKET_ALLOWED_HOSTS:
         raise ValueError(f"URL host '{parts.hostname}' is not a known Bitbucket domain")
+    # Reject path traversal: inspect each decoded path segment so that `..`
+    # embedded inside a legitimate segment (e.g. a diff spec like `abc..def`)
+    # is allowed, while a standalone `..` segment is rejected.
+    for segment in parts.path.split("/"):
```
Zarlanx, line 30: splits on forward slash only, missing backslash-separated paths on any host that later normalizes backslash to slash

server/connectors/bitbucket_connector/api_client.py
```diff
@@ -332,26 +360,57 @@ def delete_branch(self, workspace, repo_slug, name):
...
-        """Get diff for a spec (commit hash, branch, or base..head range)."""
-        url = f"{BITBUCKET_API_BASE}/repositories/{workspace}/{repo_slug}/diff/{spec}"
+        """Get diff for a spec (commit hash, branch, or base..head range).
+
+        For a ``base..head`` range, each side is URL-encoded individually so
+        that refs containing reserved characters remain safe while the literal
+        ``..`` separator is preserved as a single path segment.
+        """
+        if ".." in spec:
+            base, _, head = spec.partition("..")
```
Zarlanx: spec.partition discards everything after the first ..; a..b..c becomes an incorrectly encoded spec rather than a clear ValueError

server/connectors/bitbucket_connector/api_client.py
```diff
@@ -332,26 +360,57 @@ def delete_branch(self, workspace, repo_slug, name):
...
     # Pull Requests
     # ------------------------------------------------------------------
 
+    def _pr_base(self, workspace, repo_slug, pr_id=None):
+        base = (
+            f"{BITBUCKET_API_BASE}/repositories/"
+            f"{quote(workspace, safe='')}/{quote(repo_slug, safe='')}/pullrequests"
+        )
+        if pr_id is not None:
+            return f"{base}/{quote(str(pr_id), safe='')}"
```
Zarlanx, line 428: quote(str(pr_id)) now coerces ints, floats or None where the old f-string path raised TypeError, hiding caller bugs behind a weird URL

server/utils/logging/secure_logging.py
```diff
@@ -15,21 +15,27 @@
...
         Masked string like "sk_te***MASKED***"
     """
     if not value or len(value) <= show_prefix:
         return "***MASKED***"
-    
-    return f"{value[:show_prefix]}***MASKED***"
+
+    # Strip control chars from the prefix so a crafted credential that starts
+    # with CR/LF cannot forge log entries via the masked output (S5145).
+    safe_prefix = _LOG_CONTROL_CHARS.sub("", value[:show_prefix])
```
Zarlanx, line 37: stripping control chars from the prefix only leaves the rest of the credential untouched; if any other caller later logs raw value the mask is bypassed

server/routes/datadog/datadog_routes.py
```diff
@@ -66,7 +66,7 @@ def _normalize_site(site: Optional[str]) -> Tuple[str, str]:
     if host_candidate in _SITE_BASE_URLS:
         return host_candidate, _SITE_BASE_URLS[host_candidate]
```
Zarlanx, line 68: dropping the rejected site value removes the only operator signal for why a user keeps landing on the default; log a hash or first N chars instead

server/utils/auth/token_management.py
```diff
@@ -541,6 +542,6 @@ def _schedule_prediscovery(user_id: str) -> None:
             kwargs={"user_id": user_id, "trigger": "new_connector"},
             countdown=600,
         )
-        logger.info("[STORE-TOKENS] Scheduled prediscovery for user %s (10min delay)", user_id)
+        logger.info("[STORE-TOKENS] Scheduled prediscovery for user_hash=%s (10min delay)", hash_for_log(user_id))
```
Zarlanx, line 545: hashing user_id in this one line breaks grep correlation with every other log in the same flow that still prints the raw id

server/utils/db/connection_utils.py
```diff
@@ -80,7 +81,7 @@ def save_connection_metadata(
                 ),
             )
         conn.commit()
-        logger.info("[CONN-META] Upsert successful for %s/%s/%s", user_id, provider, account_id)
+        logger.info("[CONN-META] Upsert successful user=%s provider=%s account=%s", hash_for_log(user_id), safe_provider(provider), hash_for_log(account_id))
```
Zarlanx, line 84: routing this upsert log through hash_for_log means a missing FLASK_SECRET_KEY converts a successful DB write into an unhandled 500

.github/workflows/build.yml
```diff
@@ -95,6 +94,8 @@ jobs:
     # Only run builds on the main repository, not on forks
     if: github.event.pull_request.head.repo.full_name == github.repository || github.event_name != 'pull_request'
     environment: ${{ github.ref == 'refs/heads/main' && 'staging' || (startsWith(github.ref, 'refs/heads/release/') && 'production' || 'development') }}
+    permissions:
+      contents: read
```
Zarlanx, line 98: build-summary further down still has no permissions block, so removing the workflow-level default lets it inherit the repo default and potentially widen scope

.github/workflows/publish-images.yml
```diff
@@ -139,6 +138,9 @@ jobs:
     needs: build
     if: github.repository == 'Arvo-AI/aurora'
     runs-on: ubuntu-24.04
+    permissions:
+      contents: read
+      packages: write
```
Zarlanx, line 143: publish-summary further down has no permissions block; after removing the workflow-level one it inherits the repo default, defeating the hardening

.github/workflows/publish-airtight.yml
```diff
@@ -14,17 +14,16 @@ concurrency:
...
-
 jobs:
   airtight:
     name: Airtight ${{ matrix.arch }}
     runs-on: ${{ matrix.runner }}
     if: github.repository == 'Arvo-AI/aurora'
+    permissions:
+      contents: read
+      id-token: write
+      attestations: write
```
Zarlanx, line 25: publish-summary job further down has no permissions block; removing the workflow-level default means it now inherits the repo default

.github/workflows/publish-helm.yml
```diff
@@ -87,6 +86,9 @@ jobs:
     runs-on: ubuntu-24.04
     needs: [validate-version]
     if: github.repository == 'Arvo-AI/aurora' && needs.validate-version.result == 'success'
+    permissions:
+      contents: read
+      packages: write   # GHCR OCI push
```
Zarlanx, line 91: publish-summary further down has no permissions block; it now inherits the repo default which may include write scopes

.github/workflows/docs.yml
```diff
@@ -53,6 +50,9 @@ jobs:
       url: ${{ steps.deployment.outputs.page_url }}
     runs-on: ubuntu-latest
     needs: build
+    permissions:
+      pages: write
+      id-token: write
```
Zarlanx, line 55: deploy job declares pages and id-token but no contents: read, so any future step that calls actions/checkout will silently fail

.github/workflows/publish-helm.yml
```diff
@@ -55,7 +52,9 @@ jobs:
     runs-on: ubuntu-24.04
     needs: [validate-version]
     if: github.repository == 'Arvo-AI/aurora' && needs.validate-version.result == 'success'
-
+    permissions:
+      contents: write   # GitHub Pages release assets + gh-pages branch
```
Zarlanx, line 56: contents: write here grants full repo-content and release-asset write; chart-releaser only needs the gh-pages branch but yaml cannot express that, so at minimum explain the scope in the comment

server/routes/grafana/grafana_routes.py
```diff
@@ -115,20 +116,20 @@ def alert_webhook(user_id: str):
     if not row_exists or (row_exists and not is_active):
         skip_rca = True
         if not row_exists:
-            logger.info("[GRAFANA] Auto-connecting user %s via webhook", user_id)
+            logger.info("[GRAFANA] Auto-connecting user %s via webhook", sanitize(user_id))
             try:
                 store_tokens_in_db(user_id, {}, "grafana")
-            except Exception as exc:
-                logger.exception("[GRAFANA] Failed to auto-connect user %s: %s", user_id, exc)
+            except Exception:
```
Zarlanx, line 122: dropping exc from logger.exception keeps the traceback via exc_info but loses the exception class from the message line; inconsistent with how the rest of this PR rewrites these lines

server/routes/datadog/datadog_routes.py
```diff
@@ -301,26 +301,24 @@ def connect(user_id):
 
     site, base_url = _normalize_site(raw_site)
 
-    masked_api = mask_credential_value(api_key)
-    masked_app = mask_credential_value(app_key)
-    logger.info("[DATADOG] Connecting user %s to site=%s api=%s app=%s", user_id, site, masked_api, masked_app)
+    logger.info("[DATADOG] Connecting user %s to site=%s api_hash=%s app_hash=%s", sanitize(user_id), sanitize(site), hash_for_log(api_key), hash_for_log(app_key))
```
Zarlanx, line 304: hash_for_log(api_key) in the hot path of /connect turns a missing FLASK_SECRET_KEY into a 500 on every credential submission, before the key is even validated

server/connectors/spinnaker_connector/client.py
```diff
@@ -266,7 +268,7 @@ def _request(self, method: str, path: str, **kwargs: Any) -> Any:
         try:
             resp = self._session.request(method, url, **kwargs)
         except requests.RequestException as exc:
-            logger.error("[SPINNAKER] %s %s network error: %s", method, path, exc)
+            logger.error("[SPINNAKER] %s %s network error: %s", sanitize(method), sanitize(path), exc)
```
Zarlanx: method here is a hardcoded HTTP verb passed from internal call sites and cannot be tainted; wrapping it in sanitize is pure noise

server/routes/admin_routes.py
```diff
@@ -134,7 +135,8 @@ def create_user(user_id):
                 conn.commit()
 
                 logger.info("Admin %s created invitation %s for existing user %s (%s) to join org %s",
-                            user_id, invitation_id, target_id, target_email, org_id)
+                            sanitize(user_id), invitation_id, sanitize(target_id),
```
Zarlanx, line 138: invitation_id is a UUID generated one line up so it is safe, but it was left unwrapped while the neighboring fields all got sanitize; either wrap everything or add a comment explaining the exemption

server/routes/bigpanda/bigpanda_routes.py
```diff
@@ -124,32 +125,32 @@ def webhook(user_id: str):
         return jsonify({"error": "user_id is required"}), 400
 
     if not _verify_webhook_user(user_id):
-        logger.warning("[BIGPANDA] Webhook rejected: invalid or unconfigured user_id %s", user_id[:50])
+        logger.warning("[BIGPANDA] Webhook rejected: invalid or unconfigured user_id %s", sanitize(user_id)[:50])
```
Zarlanx, line 128: sanitize(user_id)[:50] slices after sanitize; a tainted user_id that used to show 50 chars now shows a different prefix than existing operator runbooks expect

server/utils/aws/aws_sts_client.py
```diff
@@ -123,8 +125,8 @@ def assume_workspace_role(
         
         if external_id != workspace_external_id:
             logger.error(
-                f"SECURITY: External ID mismatch for workspace {workspace_id}. "
-                f"Provided: '{external_id}', Expected: '{workspace_external_id}'. Rejecting role assumption."
+                f"SECURITY: External ID mismatch for workspace {sanitize(workspace_id)}. "
```
Zarlanx: the expected workspace_external_id is the stored trust-policy secret; logging it even sanitized turns log read access into external-id exposure and defeats the point of treating it as a shared secret

server/utils/log_sanitizer.py
isiddharthsingh: fixed — _get_log_hash_salt now returns None and hash_for_log falls back to a '?' sentinel, so a missing FLASK_SECRET_KEY can't take down any request whose only sin was logging a user_id

server/utils/log_sanitizer.py
isiddharthsingh: leaving as-is — we treat FLASK_SECRET_KEY rotation as a deploy event (flask's own session signing has the same restart constraint), so reading it once on first log call is the behaviour we want

server/utils/log_sanitizer.py
isiddharthsingh: fixed — the regex now covers U+2028/2029, the zero-width joiners (200b-200f), word joiner and BOM, and it's written with \uXXXX escapes so the invisible chars can't silently disappear on edit

server/utils/log_sanitizer.py
isiddharthsingh: leaving as-is — the collapse to 'unknown' is the whole point of an allowlist; malformed providers showing up as 'unknown' in the log is itself the signal, a warning log here would just create noise

server/utils/log_sanitizer.py
isiddharthsingh: leaving as-is — the docstring already says correlation-only and we never use hash_for_log as a dedup/lookup key, no caller needs collision-resistance

server/utils/providers.py
isiddharthsingh: leaving as-is — a stray provider value collapses to 'unknown' in logs which is itself the signal; a real enforcement test would have to enumerate every insert site and isn't worth the upkeep

server/connectors/bitbucket_connector/api_client.py
isiddharthsingh: fixed — _validate_bitbucket_url now decodes each segment to a fixed point (up to 3 rounds) so %252e%252e and deeper encodings get rejected

server/connectors/bitbucket_connector/api_client.py
isiddharthsingh: leaving as-is — urlsplit only treats / as a separator and our two allowed hosts don't normalise \ to / at the path level, so backslashes stay opaque inside the segment and can't sneak past the allowlist

server/connectors/bitbucket_connector/api_client.py
isiddharthsingh: fixed — get_diff now splits on '..' and raises ValueError if there's more than one separator instead of silently mangling a..b..c

server/connectors/bitbucket_connector/api_client.py
isiddharthsingh: leaving as-is — pr_id/issue_id are int from every internal caller and the str() coercion existed in the pre-PR f-string path too, so this isn't a regression; strict isinstance would be a behaviour change, not a security fix

server/utils/logging/secure_logging.py
isiddharthsingh: leaving as-is — mask_credential_value is the masker itself; a caller that logs the raw value is a separate problem and out of scope for this sweep

server/routes/datadog/datadog_routes.py
isiddharthsingh: leaving as-is — a rejected site is almost always a typo and we allowlist legal sites in the enum; a hash or prefix here adds log noise without a clear debugging win

server/utils/auth/token_management.py
isiddharthsingh: leaving as-is — every new call site in this PR uses the same HMAC so they correlate with each other; correlating with the old raw-user_id logs elsewhere is what a codebase-wide sweep would fix and that's scoped out of this PR

server/utils/db/connection_utils.py
isiddharthsingh: fixed via the sentinel change in log_sanitizer — hash_for_log no longer raises when FLASK_SECRET_KEY is missing, so this upsert log path can't 500 a db write anymore

.github/workflows/build.yml
isiddharthsingh: fixed — build-summary now has permissions: {} so it doesn't inherit the repo default

.github/workflows/publish-images.yml
isiddharthsingh: fixed — publish-summary now has permissions: {} so it can't inherit the repo default

.github/workflows/publish-airtight.yml
isiddharthsingh: fixed — publish-summary now has permissions: {} so it can't inherit the repo default

.github/workflows/publish-helm.yml
isiddharthsingh: fixed — publish-summary now has permissions: {} so it can't inherit the repo default

.github/workflows/docs.yml
isiddharthsingh: leaving as-is — the deploy job is a single actions/deploy-pages@v4 call with no checkout, so contents: read here would be speculative hardening for a step that doesn't exist; happy to add it the day someone wires checkout in

.github/workflows/publish-helm.yml
isiddharthsingh: leaving as-is — github tokens can't scope contents: write by branch in yaml so this is the tightest expressible scope for chart-releaser's gh-pages push; can add an inline comment explaining the scope if you want

server/routes/grafana/grafana_routes.py
isiddharthsingh: leaving as-is — logger.exception still writes the exception class via exc_info on the next line, so we're not actually losing the class, just dropping one duplicate mention from the message line

server/routes/datadog/datadog_routes.py
isiddharthsingh: fixed via the sentinel change in log_sanitizer — hash_for_log falls back to '?' when FLASK_SECRET_KEY is missing, so /connect can no longer 500 on credential submission

server/connectors/spinnaker_connector/client.py
isiddharthsingh: fixed — dropped the sanitize() wrapper around method on both log lines since it's a hardcoded http verb from internal call sites

server/routes/admin_routes.py
isiddharthsingh: leaving as-is — invitation_id is generated by uuid.uuid4() one line above so it's not tainted; wrapping it would imply taint that isn't there

server/routes/bigpanda/bigpanda_routes.py
isiddharthsingh: leaving as-is — sanitize is a no-op on a valid user_id so the first 50 chars are identical to before for every legitimate input; slice order only mutates output for already-malicious input, which is fine

server/utils/aws/aws_sts_client.py
isiddharthsingh: fixed — the expected external_id is no longer logged at all; the log now only carries a hash_for_log fingerprint of the attacker-provided value, which we can correlate across repeat probes without leaking the trust-policy secret

server/connectors/bitbucket_connector/api_client.py
```diff
@@ -15,12 +15,25 @@
...
+    # re-decode into ``..``) are also rejected.
+    for segment in parts.path.split("/"):
+        decoded = segment
+        for _ in range(3):
+            previous = decoded
+            decoded = unquote(decoded)
+            if decoded == previous:
+                break
+        if decoded == "..":
+            raise ValueError("URL path contains a traversal segment")
```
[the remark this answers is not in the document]
isiddharthsingh: good catch — fixed in b9186827. now decoding each segment to a fixed point AND re-splitting on '/' afterwards, so %2e%2e%2fsecret (and the double-encoded variant) can't smuggle a '..' component past the equality check anymore

server/connectors/bitbucket_connector/api_client.py
```diff
@@ -332,26 +366,60 @@ def delete_branch(self, workspace, repo_slug, name):
...
+            base, head = parts
+            encoded_spec = f"{quote(base, safe='')}..{quote(head, safe='')}"
+        else:
+            encoded_spec = quote(spec, safe='')
+        url = (
+            f"{BITBUCKET_API_BASE}/repositories/"
+            f"{quote(workspace, safe='')}/{quote(repo_slug, safe='')}"
+            f"/diff/{encoded_spec}"
+        )
         return self._get_raw(url)
```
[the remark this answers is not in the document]
isiddharthsingh: also valid — fixed in b9186827. get_diff now returns a structured {'error': True, 'status': None, 'message': ...} dict instead of raising, so branches_tool.py's forward_if_error handles it gracefully like every other path in this client

Zarlanx (approved):

PR #305, opened 2026-04-24
feat(prediscovery): add auto-discovery toggle with cancel

isiddharthsingh:
## Summary
- Adds an Automatic Discovery switch in Knowledge Base settings. When off, the hourly scheduler skips the user via a new `prediscovery_enabled` preference.
- Toggling off also cancels any in-flight run: new `POST /api/prediscovery/cancel` revokes the Celery task (`task_id` now stored in session metadata at dispatch) and marks the session `cancelled`.
- Promotes `cancelled` to a first-class terminal session status — `TERMINAL_SESSION_STATUSES` covers it in `_update_session_status` suggestion propagation and in the RCA context-update gates. UI renders "Last run cancelled".
- Cancel scope is per-user (SELECT filters by `user_id`), so users can't cancel sessions owned by other editors in the same org.

## Test plan
- [x] Toggle off with no in-flight run: preference saves, no session changes, scheduler skips user
- [x] Toggle off during in-flight run: Celery task revoked, session flips to `cancelled`, UI shows "Last run cancelled"
- [x] Toggle on: preference saves, scheduler resumes on next hourly tick
- [x] Manual "Run Now" still works when toggle is off

## Summary by CodeRabbit

* **New Features**
  * Added "Automatic Discovery" toggle in settings with persisted preference.
  * Added ability to cancel in-progress discovery runs from the UI.
  * Status display now shows "Last run cancelled" when a run is cancelled.

* **Improvements**
  * Cancellation and run status handling made more reliable and reflect cancellations promptly.

### DIFF
--- a/client/src/components/DiscoverySettings.tsx
+++ b/client/src/components/DiscoverySettings.tsx
@@ -5,6 +5,7 @@ import { Card, CardContent, CardDescription, CardHeader, CardTitle } from "@/com
 import { Button } from "@/components/ui/button";
 import { Input } from "@/components/ui/input";
 import { Label } from "@/components/ui/label";
+import { Switch } from "@/components/ui/switch";
 import { useToast } from "@/hooks/use-toast";
 import { Radar, Check } from "lucide-react";
 import { useAuth } from "@/hooks/useAuthHooks";
@@ -15,6 +16,8 @@ export function DiscoverySettings() {
   const [discovering, setDiscovering] = useState(false);
   const [intervalHours, setIntervalHours] = useState<number>(24);
   const [savingInterval, setSavingInterval] = useState(false);
+  const [autoEnabled, setAutoEnabled] = useState<boolean>(true);
+  const [savingEnabled, setSavingEnabled] = useState(false);
   const { toast } = useToast();
   const { userId } = useAuth();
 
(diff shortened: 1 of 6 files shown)


server/routes/prediscovery/routes.py
```diff
@@ -36,6 +36,50 @@ def trigger_prediscovery(user_id):
         return jsonify({"error": "Failed to start discovery"}), 500
 
 
+@prediscovery_bp.route("/cancel", methods=["POST"])
+@require_permission("connectors", "write")
+def cancel_prediscovery(user_id):
+    """Cancel any in-progress prediscovery runs for this user."""
+    try:
+        org_id = get_org_id_from_request()
```
Zarlanx, line 44: get_org_id_from_request can return None; the SQL then runs with org_id = NULL which matches no rows, so a missing X-Org-ID header silently returns no_active_sessions instead of a 403

server/routes/prediscovery/routes.py
```diff
@@ -36,6 +36,50 @@ def trigger_prediscovery(user_id):
...
+                """, (org_id, user_id))
+                rows = cur.fetchall()
+
+        if rows:
+            from celery_config import celery_app
+            from chat.background.task import _update_session_status
+            for session_id, task_id in rows:
+                if task_id:
+                    try:
+                        celery_app.control.revoke(task_id, terminate=True, signal="SIGTERM")
```
Zarlanx: SIGTERM via celery revoke does not interrupt a Python thread blocked in a C-level socket call like requests or boto3; the task can keep running and produce side effects after the session is marked cancelled

server/chat/background/context_updates.py
```diff
@@ -11,6 +11,7 @@
 from utils.cache.redis_client import get_redis_client
 from utils.db.connection_pool import db_pool
 from utils.auth.stateless_auth import set_rls_context
+from chat.background.task import TERMINAL_SESSION_STATUSES
```
Zarlanx: new import from chat.background.task creates a cross-file constant dependency; task is already widely imported so verify no circular-import path is created at cold start by a worker that loads context_updates first

client/src/components/DiscoverySettings.tsx
```diff
@@ -36,8 +39,63 @@ export function DiscoverySettings() {
...
+        if (data.value != null) setAutoEnabled(Boolean(data.value));
+      })
+      .catch(() => {});
   }, [userId]);
 
+  const cancelAndRefreshStatus = async () => {
+    try {
+      const cancelRes = await fetch("/api/prediscovery/cancel", { method: "POST", credentials: "include" });
+      if (!cancelRes.ok) {
+        console.error("Failed to cancel prediscovery:", cancelRes.status);
```
Zarlanx, line 57: cancel failure is logged and swallowed here; the outer toggle then shows a success toast claiming the running scan was cancelled even when the cancel request failed

server/routes/prediscovery/routes.py
isiddharthsingh: Added an explicit check — endpoint now returns 400 {"error": "Missing org context"} when get_org_id_from_request() returns None instead of running WHERE org_id = NULL and silently reporting no_active_sessions. Fixed in 2478c974.

server/routes/prediscovery/routes.py
isiddharthsingh: SIGTERM gets queued behind the GIL while the worker is in a blocking requests/boto3 call, so the task can keep running after we mark the session cancelled. Switched the revoke signal to SIGKILL in 2478c974. Aurora's prefork pool runs each task in a child process; SIGKILL terminates that child immediately and Celery spawns a fresh one, so a stuck worker can't keep producing side effects. Verified in worker logs: Terminating <task_id> (9).

client/src/components/DiscoverySettings.tsx
isiddharthsingh: Fixed in dd73d6f3. cancelAndRefreshStatus now returns a boolean; the toggle handler reads it and shows a destructive "Scheduled scans paused, but cancelling the running scan failed." toast when the cancel call returned non-OK or threw. The new dedicated Cancel button uses the same return value to surface a "Failed to cancel" toast on failure instead of claiming success.

OlivierTrudeau (approved):

PR #312, opened 2026-04-25
feat(security): add SigmaHQ-backed signature corpus for L2 guardrail

damianloch:
## Summary

Vendors 37 SigmaHQ rules (release `r2026-01-01`, scope: `linux/process_creation`, severity `high`/`critical`) and transpiles them to regex at import time, extending the L2 signature matcher from 29 hand-written rules to 60 total. Adds per-rule suppression, env-var toggle, and L1/L2 dedupe documentation.

This PR replaces #307 with the content properly rebased on the current tip of `feat/nemo-input-rail-audit`. The original PR was built before the config simplification, fail-closed input rail, audit-log fingerprinting, and other base-branch hardening landed; this branch drops the stale NeMo duplicates and keeps only the three SigmaHQ commits, squashed.

## What changed

- **`sigma_loader.py`** — narrow Sigma-to-regex transpiler. Supports `CommandLine`, `Image`, `OriginalFileName` fields with `contains` / `endswith` / `startswith` / `re` / `contains|all` modifiers, and conditions `selection`, `all of selection_*`, `1 of selection_*`, and simple `X and Y`. Selections referencing unsupported fields are dropped whole to avoid false positives from silently weakened AND conditions.
- **`signature_match.py`** — `_load_sigma()` hook appends Sigma rules after the hand-written rules at import time; failure degrades gracefully to the hand-written subset. Existing rule IDs and audit-log fields are stable.
- **`config.py`** — adds `sigma_enabled` field + `GUARDRAILS_SIGMA_ENABLED` env var (default `true`, gated by `GUARDRAILS_ENABLED`). No per-layer toggles re-introduced.
- **`sigma_rules/`** — vendored corpus with `README`, `VERSION` pin, and `suppressions.txt` for per-rule opt-out. Runtime suppression also available via `GUARDRAILS_SIGMA_SUPPRESS` env var (comma-separated UUIDs).
- **`command_policy.py`** — documents intentional L1 (org-configurable operational policy) vs L2 (security detection signatures) overlap as defense-in-depth. No behavior change.
- **`test_sigma_canary.py`** — 48 benign ops commands (kubectl, aws, docker, terraform, git, curl, ssh, `rm /tmp/*`, etc.) that must not trigger any Sigma rule, plus 12 malicious commands (GTFOBin shell spawns, rootkit install, crypto mining, history deletion) that must match.

## Supported Sigma subset

| Feature | Supported |
|---|---|
| Fields | `CommandLine`, `Image`, `OriginalFileName` |
| Modifiers | `contains`, `endswith`, `startswith`, `re`, `contains\|all` |
| Conditions | `selection`, `all of selection_*`, `1 of selection_*`, simple `X and Y` |
| Severity gate | `high` and `critical` only |
| Unsupported fields | Selection dropped whole (not silently weakened) |

## Test plan

- [x] 60/60 `test_sigma_canary.py` pass (48 benign canaries + 12 malicious detections).
- [x] Verified live in the running `aurora-server` container: 60 rules load (29 hand-written + 31 Sigma), audit events emit well-formed `layer` / `rule_id` (e.g. `sigma-8c1a5675`) / `technique` (e.g. `T1059`) / `command_fp` (sha256) fields.
- [x] Toggle works: `GUARDRAILS_SIGMA_ENABLED=false` loads 29 rules, `true` loads 60.
- [x] Sample coverage verified end-to-end via `terminal_run`:
  - `cat /etc/shadow` -> `cred-shadow` (hand-written)
  - `/usr/bin/awk 'BEGIN {system("/bin/bash")}'` -> `sigma-8c1a5675` ("Suspicious Invocation of Shell via AWK - Linux")
  - `/usr/bin/vim -c ':!/bin/bash'` -> `sigma-7ab8f73a` ("Vim GTFOBin Abuse - Linux")
  - `/usr/bin/find . -exec /bin/sh \;` -> `sigma-6adfbf8f`
  - `xmrig --cpu-priority=5 --donate-level=0` -> `mining-binary`
- [x] No linter errors.

## Notes for reviewers

- Base branch is `feat/nemo-input-rail-audit`, not `main`. Merge #307-sized concerns (input rail, audit event hardening, config simplification) are already on the base and do not appear here.
- Docs in `website/docs/configuration/` are not updated by this PR to avoid churn against the base-branch copy; consider a follow-up to mention `GUARDRAILS_SIGMA_ENABLED` and `GUARDRAILS_SIGMA_SUPPRESS` in the env-var table.

Made-with: Cursor

## Release Notes

* **New Features**
  * Added comprehensive threat detection rules targeting suspicious Linux system behaviors, including shell execution exploits, reverse shells, crypto mining, log tampering, rootkit indicators, and web shell activity.
  * Integrated enhanced detection system with expanded rule coverage for defense-in-depth security.

* **Tests**
  * Added test suite validating detection rule accuracy and rule coverage enforcement.


server/utils/security/signature_match.py
```diff
@@ -102,6 +106,24 @@ def _r(pattern: str, technique: str, rule_id: str, desc: str) -> None:
...
+    try:
+        from utils.security.config import config
+        if not config.sigma_enabled:
+            logger.debug("Sigma rules disabled via config")
+            return
+        from utils.security.sigma_loader import load_sigma_rules
+        sigma_rules = load_sigma_rules()
+        _RULES.extend(sigma_rules)
+    except Exception:
+        logger.warning("Failed to load Sigma rules, continuing with hand-written rules only", exc_info=True)
```
damianloch, line 123: Log the post-load rule count so you can grep logs to confirm Sigma actually loaded

server/utils/security/sigma_rules/proc_creation_lnx_auditctl_clear_rules.yml
damianloch, line 1: this goes for all rules yml files, make sure you properly cite them even tho they have MIT license i think you need to put a comment at the top or stmg

server/utils/security/sigma_rules/README
damianloch, line 1: Bronem forgot the .md

damianloch, line 1: I would be extremely specific here about what exact files at what SHA were taken for the repo so that when we want to update these we have the clear diff instead of needing to search

server/utils/security/sigma_loader.py
```diff
@@ -0,0 +1,260 @@
...
+        if not referenced:
+            referenced = list(selections.keys())
+        parts = [selections[k] for k in referenced if k in selections]
+        if parts:
+            return _and_all(parts)
+
+    if len(selections) == 1:
+        return next(iter(selections.values()))
+
+    return None
```
damianloch, line 238: the condition resolver uses in substring matches on cond_lower, which is brittle. Specifically, a condition like "1 of selection_* and not filter_xyz" will not match the exact string at L167 and instead falls through to the "all of selection" branch at L170, silently flipping OR semantics to AND (under-detection). None of the 37 vendored rules hit this today, but it's a trap for any future rule.

server/utils/security/sigma_loader.py
```diff
@@ -0,0 +1,260 @@
...
+def _extract_mitre_technique(tags: list) -> str:
+    """Extract the first MITRE ATT&CK technique ID from Sigma tags."""
+    for tag in (tags or []):
+        tag_str = str(tag)
+        if tag_str.startswith("attack.t") and len(tag_str) > 8:
+            return tag_str.replace("attack.", "").upper()
+    return ""
+
+
+_SUPPORTED_FIELDS = {"commandline", "image", "originalfilename"}
```
damianloch: OriginalFileName is declared supported but uses _cmdline_pattern fallback, which anchors endswith to the end of the entire command line. OriginalFileName in Sigma refers to the PE/ELF binary's embedded original-name metadat, but in Aurora's shell-string matching context, this field has no meaningful equivalent and we only ever see the command line

See line 106

server/utils/security/sigma_loader.py
```diff
@@ -0,0 +1,260 @@
...
+
+def _escape_literal(s: str) -> str:
+    return re.escape(s)
+
+
+def _image_pattern(lit: str, modifiers: List[str]) -> str:
+    if "endswith" in modifiers:
+        return f"^\\S*{lit}(?:\\s|$)"
+    if "startswith" in modifiers:
+        return f"^{lit}"
```
damianloch: _image_pattern applies a right-side boundary ((?:\s|$)) for endswith and contains but not for startswith. That means Image|startswith: '/bin/sh' would also match /bin/shadow-dump -foo

server/utils/security/sigma_loader.py
```diff
@@ -0,0 +1,260 @@
...
+        if sel_parts:
+            return _and_all(sel_parts)
+
+    if " and " in cond_lower and "not" not in cond_lower:
+        referenced = [k for k in selections if k in cond_lower.replace("_", " ").replace("*", "")]
+        if not referenced:
+            referenced = list(selections.keys())
+        parts = [selections[k] for k in referenced if k in selections]
+        if parts:
+            return _and_all(parts)
```
damianloch: We are doing substring tricks here instead of parsing.There are two silent failure modes: " not " not in cond_lower false-rejects any rule with "not" inside a selection name, and the replace("_", " ") trick at L176 silently flattens complex conditions like selection_a and (selection_b or selection_c) into a pure AND.

server/utils/security/sigma_loader.py
damianloch, line 1: _SUPPRESSIONS is reassigned via global in load_sigma_rules() and read implicitly by _process_sigma_file() as a free variable. It works today because there's only one caller, but any test that calls _process_sigma_file() directly inherits stale state from whatever ran before it.

server/utils/security/config.py
```diff
@@ -11,12 +11,15 @@
...
+    sigma_enabled: bool
     llm_model: str
 
 
 def _load() -> GuardrailsConfig:
+    on = os.getenv("GUARDRAILS_ENABLED", "false").lower() == "true"
     return GuardrailsConfig(
-        enabled=os.getenv("GUARDRAILS_ENABLED", "false").lower() == "true",
+        enabled=on,
+        sigma_enabled=on and os.getenv("GUARDRAILS_SIGMA_ENABLED", "true").lower() != "false",
```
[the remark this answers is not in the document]
beng360: Fixed in 291aae03. `GUARDRAILS_SIGMA_ENABLED` now recognizes `0`, `no`, `off`, and `disabled` as falsy values via a `_FALSY` set, matching operator expectations during incidents.

server/utils/security/sigma_loader.py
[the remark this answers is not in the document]
beng360: Fixed in 291aae03. Rewrote compound condition parsing — extracted `_resolve_token()` and `_resolve_compound()` helpers. Each AND token is now resolved independently: bare names do direct lookup, `1 of X*`/`all of X*` do prefix-matched glob resolution. If any token resolves to None (e.g. all `selection_*` entries dropped due to unsupported fields), the entire condition returns None. This fixes the webshell false-match scenario. SonarQube re-scan: 0 issues (complexity stayed under 15).

server/utils/security/sigma_rules/proc_creation_lnx_auditctl_clear_rules.yml
```diff
@@ -0,0 +1,26 @@
+title: Audit Rules Deleted Via Auditctl
+id: bed26dea-4525-47f4-b24a-76e30e44ffb0
+status: experimental
+description: |
+    Detects the execution of 'auditctl' with the '-D' command line parameter, which deletes all configured audit rules and watches on Linux systems.
+    This technique is commonly used by attackers to disable audit logging and cover their tracks by removing monitoring capabilities.
+    Removal of audit rules can significantly impair detection of malicious activities on the affected system.
+references:
+    - https://www.atomicredteam.io/atomic-red-team/atomics/T1562.012
+    - https://linux.die.net/man/8/auditct
```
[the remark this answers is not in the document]
beng360: Fixed in 291aae03 — `auditct` → `auditctl`.

server/utils/security/sigma_rules/proc_creation_lnx_auditctl_clear_rules.yml
```diff
@@ -0,0 +1,26 @@
...
+    - attack.defense-evasion
+    - attack.t1562.012
+logsource:
+    product: linux
+    category: process_creation
+detection:
+    selection:
+        Image|endswith: '/auditctl'
+        CommandLine|re: '-D'
+    condition: selection
```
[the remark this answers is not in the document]
beng360: Fixed in 291aae03. Changed `CommandLine|re: '-D'` to `CommandLine|contains: ' -D'` (with leading space), which matches the upstream SigmaHQ convention and avoids false matches on `-Debug`, `--Dryrun`, etc.

server/utils/security/sigma_rules/proc_creation_lnx_crypto_mining.yml
```diff
@@ -0,0 +1,42 @@
...
+        CommandLine|contains:
+            - ' --cpu-priority='
+            - '--donate-level=0'
+            - ' -o pool.'
+            - ' --nicehash'
+            - ' --algo=rx/0 '
+            - 'stratum+tcp://'
+            - 'stratum+udp://'
+            # Sub process started by xmrig - the most popular Monero crypto miner - unknown if this causes any false positives
+            - 'sh -c /sbin/modprobe msr allow_writes=on'
```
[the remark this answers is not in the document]
beng360: Acknowledged. The `-o pool.` token is part of the upstream SigmaHQ rule (contains|all with other mining-specific tokens like `--donate-level`, `--cpu-priority`, `stratum+`). The `contains|all` modifier requires ALL tokens to be present simultaneously, so `-o pool.` alone won't fire — it needs the full constellation of mining flags. Tested: `kubectl get pods -o pool.yaml` does not match. Keeping as-is.

server/utils/security/sigma_rules/proc_creation_lnx_curl_wget_exec_tmp.yml
```diff
@@ -0,0 +1,36 @@
+title: Suspicious Download and Execute Pattern via Curl/Wget
+id: a2d9e2f3-0f43-4c7a-bcd9-9acfc0d723aa
+status: experimental
```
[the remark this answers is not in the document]
beng360: Acknowledged. All 37 vendored rules are `test` or `experimental` — the SigmaHQ linux/process_creation corpus has zero `stable` rules at the r2026-01-01 release. Filtering by status would drop the entire corpus. The status reflects SigmaHQ's community validation process, not FP rate — we validate independently via the 48-command canary suite. The `level: high/critical` filter already gates severity, and the suppressions mechanism handles any per-rule opt-out needed in production.

server/utils/security/sigma_rules/proc_creation_lnx_python_shell_os_system.yml
```diff
@@ -0,0 +1,38 @@
+title: Inline Python Execution - Spawn Shell Via OS System Library
+id: 2d2f44ff-4611-4778-a8fc-323a0e9850cc
+status: test
```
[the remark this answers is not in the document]
beng360: Same as the curl/wget comment — all 37 vendored rules are `test` or `experimental` (zero `stable` in the SigmaHQ linux corpus at r2026-01-01). We validate FP rates independently via the canary test suite rather than relying on SigmaHQ's status field. Suppressions mechanism available if this specific rule causes issues in production.

server/utils/security/sigma_rules/proc_creation_lnx_susp_history_delete.yml
```diff
@@ -0,0 +1,34 @@
...
+            - '/shred'
+    selection_history:
+        - CommandLine|contains:
+              - '/.bash_history'
+              - '/.zsh_history'
+        - CommandLine|endswith:
+              - '_history'
+              - '.history'
+              - 'zhistory'
+    condition: all of selection*
```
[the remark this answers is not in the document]
beng360: Verified — the transpiler correctly handles the list-of-dicts in `selection_history`. When a selection value is a list of dicts, `_translate_selection()` iterates each dict and combines them with OR semantics. Tested: `rm /tmp/myfile.txt` → no match; `/usr/bin/rm /home/user/.bash_history` → match. The `all of selection*` condition requires both `selection` (Image|endswith of /rm, /unlink, /shred) AND `selection_history` (history file paths) — neither alone is sufficient.

server/utils/security/sigma_rules/proc_creation_lnx_susp_java_children.yml
```diff
@@ -0,0 +1,32 @@
...
+            - 'bash'
+            - 'dash'
+            - 'ksh'
+            - 'zsh'
+            - 'csh'
+            - 'fish'
+            - 'curl'
+            - 'wget'
+            - 'python'
+    condition: selection
```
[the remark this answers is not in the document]
beng360: Confirmed safe. The java_children rule's only selection uses `ParentImage|endswith` (unsupported field), so `_translate_selection_dict()` returns None, `_extract_selections()` produces an empty dict, and `_translate_rule()` returns None. The rule is silently skipped — no regex is generated, no false positives possible.

server/utils/security/sigma_rules/proc_creation_lnx_susp_shell_child_process_from_parent_tmp_folder.yml
```diff
@@ -0,0 +1,31 @@
...
+        ParentImage|startswith: '/tmp/'
+        Image|endswith:
+            - '/bash'
+            - '/csh'
+            - '/dash'
+            - '/fish'
+            - '/ksh'
+            - '/sh'
+            - '/zsh'
+    condition: selection
```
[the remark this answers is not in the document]
beng360: Confirmed safe. Same as java_children — the rule's only selection uses `ParentImage|startswith: '/tmp'` (unsupported), so the entire rule is skipped during loading. No regex generated, no match possible.

server/utils/security/sigma_rules/proc_creation_lnx_webshell_detection.yml
```diff
@@ -0,0 +1,50 @@
...
+            - '/bin/crontab'
+            - '/hostname'
+            - '/iptables'
+            - '/netstat'
+            - '/pwd'
+            - '/route'
+    condition: 1 of selection_* and sub_processes
+falsepositives:
+    - Web applications that invoke Linux command line tools
+level: high
```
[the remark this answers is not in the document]
beng360: Fixed in 291aae03. The compound condition parser now properly handles `1 of selection_* and sub_processes`. When all `selection_*` entries are dropped (ParentImage/ParentCommandLine unsupported), `_resolve_token('1 of selection_*')` returns None, which causes `_resolve_compound()` to return None for the entire condition. The rule is now correctly skipped. Verified: `whoami`, `hostname`, `uname -a`, `cat /etc/passwd` all produce no match.

server/utils/security/signature_match.py
```diff
@@ -102,6 +106,24 @@ def _r(pattern: str, technique: str, rule_id: str, desc: str) -> None:
...
+
+_load_sigma()
```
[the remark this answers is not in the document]
beng360: Fixed in 291aae03. Added `-> None` return type hint to `_load_sigma()` and moved `logger = logging.getLogger(__name__)` to the imports block next to the other top-level statements.

server/utils/security/signature_match.py
beng360: Fixed in 7e9b7db4. Added `logger.info("Signature matcher ready: %d rules (%d hand-written + %d Sigma)", ...)` after successful Sigma load.

server/utils/security/sigma_rules/proc_creation_lnx_auditctl_clear_rules.yml
beng360: Fixed in 7e9b7db4. Renamed README → README.md and added proper DRL 1.1 license attribution. Original authorship preserved in each YAML file via `author`, `date`, `id` fields — no modifications to the vendored files beyond the auditctl regex tightening.

server/utils/security/sigma_rules/README
beng360: Fixed in 7e9b7db4 — renamed to README.md.

beng360: Fixed in 7e9b7db4. README.md now includes the release tag, source path, and full manifest of all 37 vendored filenames so future diffs are unambiguous.

server/utils/security/sigma_loader.py
beng360: Fixed in 291aae03 + 7e9b7db4. Rewrote the condition resolver — compound conditions are now tokenized on ` and `, each token resolved independently via `_resolve_token()`. Negative tokens (`not X`) are properly stripped (filters unsupported, ignoring errs toward over-detection). The `replace("_", " ")` substring trick is gone. `rsync_shell_spawn.yml` (`selection and not 1 of filter_main_*`) now correctly resolves by parsing the positive `selection` token and stripping the `not 1 of filter_main_*` token.

server/utils/security/sigma_loader.py
beng360: Fixed in 7e9b7db4. Removed `originalfilename` from `_SUPPORTED_FIELDS`. No vendored rule uses it, and it has no meaningful equivalent in our shell-string matching context (Sigma's OriginalFileName refers to PE/ELF embedded metadata).

server/utils/security/sigma_loader.py
beng360: Fixed in 7e9b7db4. `_image_pattern` now applies `\S*(?:\s|$)` right-side boundary for `startswith` as well. `Image|startswith: '/bin/sh'` will match `/bin/sh -c 'echo test'` but not `/bin/shadow-dump -foo`.

server/utils/security/sigma_loader.py
beng360: Fixed in 291aae03 + 7e9b7db4. The substring-based condition parsing is replaced with proper tokenization: split on ` and `, resolve each token via `_resolve_token()`, strip `not X` tokens. No more `replace("_", " ")` flattening, no more `'not' not in cond_lower` false-rejecting selection names containing 'not'.

server/utils/security/sigma_loader.py
beng360: Fixed in 7e9b7db4. `_SUPPRESSIONS` global removed. `_process_sigma_file()` now takes `suppressions: set` as an explicit parameter, passed from `load_sigma_rules()`. No stale state possible in direct callers or tests.

server/utils/security/sigma_rules/proc_creation_lnx_crypto_mining.yml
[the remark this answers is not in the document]
beng360: You were right — I was wrong about this being `contains|all`. It's `CommandLine|contains` with a list, which is OR semantics. `kubectl get pods -o pool.yaml` was indeed a false positive.

Fixed in 9efac35d — removed the ` -o pool.` token from the rule. All other mining indicators (`--donate-level`, `stratum+tcp://`, `--cpu-priority`, `--nicehash`, base64 variants) remain and still detect correctly.

server/utils/security/sigma_loader.py
```diff
@@ -0,0 +1,285 @@
...
+def _escape_literal(s: str) -> str:
+    return re.escape(s)
+
+
+def _image_pattern(lit: str, modifiers: List[str]) -> str:
+    if "endswith" in modifiers:
+        return f"^\\S*{lit}(?:\\s|$)"
+    if "startswith" in modifiers:
+        return f"^{lit}\\S*(?:\\s|$)"
+    if "contains" in modifiers:
```
Zarlanx, line 92: the Image patterns built here require the literal (e.g. /awk) to appear inside the first whitespace-delimited token, so a real agent invocation like `awk 'BEGIN {system("/bin/bash")}'` (no leading slash) silently bypasses every Sigma rule that keys off Image|endswith; the canary tests use absolute paths and pass, giving false confidence that `awk`-style invocations are caught

server/utils/security/sigma_loader.py
```diff
@@ -0,0 +1,285 @@
...
+                       if k.startswith(prefix) or k.replace("_", "").startswith(prefix)]
+            return combiner(matched) if matched else None
+
+    return None
+
+
+def _resolve_compound(cond_lower: str, selections: Dict[str, str]) -> Optional[str]:
+    """Resolve compound 'A and B' conditions by parsing each positive token.
+
+    Negative tokens ('not X') are stripped — we don't support filter
```
Zarlanx, line 204: stripping `not filter_X` tokens silently turns rules with intentional exclusions into broader matchers; the in-code comment acknowledges erring toward over-detection but neither the PR description nor the README warns operators that filters were dropped, so when a rule fires they will not know that its exemption was discarded at transpile time

server/utils/security/sigma_loader.py
```diff
@@ -0,0 +1,285 @@
...
+        logger.debug("Suppressed Sigma rule: %s", sigma_id)
+        return None
+
+    level = (rule.get("level") or "").lower()
+    if level not in ("high", "critical"):
+        return None
+
+    regex_str = _translate_rule(rule)
+    if not regex_str:
+        logger.debug("Could not translate Sigma rule: %s", yml_path.name)
```
Zarlanx: rules that fail `_translate_rule` are skipped at DEBUG level; when bumping the vendored corpus, operators upgrading the Sigma release will not notice that coverage dropped because their default log level hides this signal, raise to WARNING when a rule is dropped due to unsupported condition or field

server/utils/security/config.py
```diff
@@ -11,12 +11,18 @@
...
 
+_FALSY = {"false", "0", "no", "off", "disabled"}
+
+
 def _load() -> GuardrailsConfig:
+    on = os.getenv("GUARDRAILS_ENABLED", "false").lower() == "true"
     return GuardrailsConfig(
-        enabled=os.getenv("GUARDRAILS_ENABLED", "false").lower() == "true",
+        enabled=on,
+        sigma_enabled=on and os.getenv("GUARDRAILS_SIGMA_ENABLED", "true").lower() not in _FALSY,
```
Zarlanx, line 25: GUARDRAILS_ENABLED accepts only the literal string `true` (line 23) while GUARDRAILS_SIGMA_ENABLED accepts anything not in the falsy set; setting GUARDRAILS_ENABLED=on silently disables guardrails entirely while GUARDRAILS_SIGMA_ENABLED=on enables Sigma; both env vars should use the same parser

server/tests/test_sigma_canary.py
```diff
@@ -0,0 +1,123 @@
...
+
+from utils.security.sigma_loader import load_sigma_rules
+from utils.security.signature_match import SignatureVerdict
+
+
+@pytest.fixture(scope="module")
+def sigma_rules():
+    """Load Sigma rules directly (bypasses config toggle for testing)."""
+    rules = load_sigma_rules()
+    assert len(rules) >= 20, f"Expected >=20 Sigma rules, got {len(rules)}"
```
Zarlanx: `>= 20` is too loose given the PR description says 31 of 37 rules currently load; 11 more rules could silently drop on a corpus bump and this assertion still passes; assert each expected sigma rule_id is present, or at minimum bump the threshold to the actual count and add a guard for >= 30

server/utils/security/signature_match.py
```diff
@@ -102,6 +108,24 @@ def _r(pattern: str, technique: str, rule_id: str, desc: str) -> None:
...
+    try:
+        from utils.security.config import config
+        if not config.sigma_enabled:
+            logger.debug("Sigma rules disabled via config")
+            return
+        from utils.security.sigma_loader import load_sigma_rules
+        sigma_rules = load_sigma_rules()
+        _RULES.extend(sigma_rules)
+        logger.info("Signature matcher ready: %d rules (%d hand-written + %d Sigma)", len(_RULES), len(_RULES) - len(sigma_rules), len(sigma_rules))
+    except Exception:
```
Zarlanx, line 122: the broad except swallows every loader error (corrupted yaml, missing dependency, regex compile bug) and falls back to hand-written rules with a WARNING log; an operator who ships GUARDRAILS_SIGMA_ENABLED=true and expects 31 extra rules will silently get only the 29 hand-written ones if anything goes wrong

server/utils/security/sigma_loader.py
```diff
@@ -0,0 +1,285 @@
...
+        return None
+
+    try:
+        compiled = re.compile(regex_str, re.IGNORECASE)
```
Zarlanx, line 281: blanket re.IGNORECASE on every Sigma rule causes FPs on rules that detect attacker tool names like `cobaltstrike`, `Sliver`, `merlin`, `havoc` (see proc_creation_lnx_susp_hktl_execution.yml); a user asking the assistant about these tools or the agent reading docs that mention them in benign contexts trips the hktl rule

server/utils/security/sigma_loader.py
```diff
@@ -0,0 +1,285 @@
...
+    except re.error:
+        logger.warning("Invalid regex from Sigma rule %s: %s", yml_path.name, regex_str[:100].replace("\n", "\\n"))
+        return None
+
+    technique = _extract_mitre_technique(rule.get("tags"))
+    title = rule.get("title", yml_path.stem)
+    rule_id = f"sigma-{sigma_id[:8]}" if sigma_id else f"sigma-{yml_path.stem}"
```
Zarlanx, line 288: rule_id uses the first 8 hex chars of the Sigma UUID (32 bits); 31 rules today is fine, but as the corpus grows toward hundreds of vendored rules, birthday collisions become likely and a colliding rule would silently shadow another in audit logs and SIEM correlation; either keep the full UUID or shorten only to the next collision-free length

server/utils/security/sigma_loader.py
beng360: yeah fair point on the bare awk case — though in practice the agent always sends commands with the full binary path (that's how process creation context works), so `/usr/bin/awk` is what we actually match against. the canary test covers that and it does match.

for bare `awk` without a leading path — we don't really see that in practice. but I'll keep an eye on it if that changes.

server/utils/security/sigma_loader.py
beng360: this is intentional — we don't support Sigma filter exclusions because implementing negation properly would need a full condition parser, and the subset we support doesn't warrant that complexity yet. the tradeoff is explicitly toward over-detection (blocking something benign) rather than under-detection (missing something malicious). the comment in the code says as much but you're right that operators should know — I'll add a note in the README about which Sigma features we don't support so it's clear when triaging a block.

server/utils/security/sigma_loader.py
beng360: good catch — bumped to WARNING so it's visible at default log levels. operators will now see which rules got dropped when they bump the corpus.

server/utils/security/config.py
beng360: yep you're right, that's a bug. fixed — added a _TRUTHY set so GUARDRAILS_ENABLED now accepts `true`, `1`, `yes`, `on`, `enabled`, same family as the sigma toggle. both env vars use consistent parsing now.

server/tests/test_sigma_canary.py
beng360: fair — bumped to >= 27 (27 is the current count after suppressing 2 FP-prone rules). also added canary cases for the FP scenarios you flagged on the rule files.

server/utils/security/signature_match.py
beng360: this is intentional — if the Sigma loader blows up for any reason (bad yaml dep, file permission, whatever), we still want the 29 hand-written rules active. the WARNING log is there so you can see it happened. making this a hard failure would mean a single corrupt yaml file takes down the entire signature matcher, which is worse imo. the hand-written rules are the baseline, Sigma is additive.

server/utils/security/sigma_loader.py
beng360: looked into this — the hktl rule only matches the Image field (binary path / first token), not CommandLine. so `cat cobaltstrike-docs.md` won't trigger because the Image is `cat`, not `cobaltstrike`. you'd have to actually execute `/path/to/cobaltstrike` as a binary for it to match, and case-insensitive on a binary path is fine. if we ever add CommandLine matching to that rule it'd be a different story, but right now it's safe.

server/utils/security/sigma_loader.py
beng360: technically true but at 27 rules the collision probability is negligible. birthday bound for 8 hex chars (32 bits) doesn't get interesting until ~hundreds of rules, and we'd hit other scaling problems before that. if we ever vendor that many we can switch to full UUIDs, but for now the short IDs are nicer in logs and audit trails.

beng360 (approved):

PR #323, opened 2026-04-27
fix(command-policies): persist allowlist/denylist toggle

damianloch:
## Summary
Toggling the Denylist/Allowlist switch on the Command Policies page was a no-op (most noticeable while the Observability Only template was active). `store_user_preference` silently returned when it couldn't resolve an org for the synthetic `__org__<uuid>` pseudo-user, so the write never landed and the UI snapped back.

- Added `store_org_preference` / `get_org_preference` helpers for org-scoped prefs so routes don't have to fake a user id.
- `store_user_preference` now logs an error instead of silently dropping the write when RLS can't be set.
- Command policy routes use the new helpers; also clears the active template on rule edit/delete (previously intended but silently broken).

## Test plan
- [ ] Apply Observability Only template, toggle Denylist off -> persists after refresh
- [ ] Toggle Allowlist on/off with no template active -> persists
- [ ] Edit/delete a rule -> active template badge clears

## Summary by CodeRabbit

* **Refactor**
  * Improved organization-level command policy preference storage through streamlined implementation.
  * Enhanced preference persistence with consolidated database transactions for better consistency of organizational settings.

### DIFF
--- a/server/utils/auth/stateless_auth.py
+++ b/server/utils/auth/stateless_auth.py
@@ -408,6 +408,9 @@ def store_org_preference(org_id: str, key: str, value: Any, *, cursor=None) -> N
     caller is responsible for RLS context and commit; otherwise an admin
     connection is opened here and RLS is configured from org_id directly.
     """
+    if not org_id:
+        raise ValueError("store_org_preference requires a non-empty org_id")
+
     params = (_org_pseudo_user_id(org_id), org_id, key, json.dumps(value))
 
     if cursor is not None:
@@ -432,6 +435,9 @@ def get_org_preference(org_id: str, key: str, default=None):
     works outside Flask request context (e.g. Celery) where connection-pool
     RLS vars aren't auto-populated.
     """
+    if not org_id:
+        raise ValueError("get_org_preference requires a non-empty org_id")
+
     from utils.db.connection_pool import db_pool
     try:
         with db_pool.get_admin_connection() as conn:
--- a/server/utils/auth/stateless_auth.py
+++ b/server/utils/auth/stateless_auth.py
@@ -340,12 +340,22 @@ def get_deployment_task(user_id: str, task_id: str = None) -> Optional[Dict]:
             conn.close()
 
 def store_user_preference(user_id: str, key: str, value: Any):
-    """Store user preference in database."""
+    """Store a user-scoped preference.
+
+    For org-scoped preferences (shared across an organization), use
+    store_org_preference() instead — passing an "__org__<uuid>" pseudo-user id
+    here will fail because that id does not exist in the users table.
+    """
     try:
         conn = connect_to_db_as_user()
         cursor = conn.cursor()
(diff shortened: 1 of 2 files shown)


server/utils/auth/stateless_auth.py
```diff
@@ -366,6 +376,48 @@ def store_user_preference(user_id: str, key: str, value: Any):
...
+            with conn.cursor() as cur:
+                cur.execute(_ORG_PREF_UPSERT_SQL, params)
+            conn.commit()
+    except Exception:
+        logger.exception("Error storing org preference %s for org %s", key, sanitize(org_id))
+
+
+def get_org_preference(org_id: str, key: str, default=None):
+    """Read an org-scoped preference written via store_org_preference()."""
+    return get_user_preference(_org_pseudo_user_id(org_id), key, default)
```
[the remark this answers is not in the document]
damianloch: Good catch, confirmed and fixed in f7692b23.

Traced through it:
- `get_user_preference` calls `set_rls_context(cursor, conn, "__org__<uuid>")`
- `get_org_id_for_user` returns `None` (no such user row), so `set_rls_context` returns `None` without setting RLS vars
- Fallback path queries `WHERE user_id = "__org__<uuid>"`, but `user_preferences` has FORCE RLS requiring `myapp.current_org_id`

This only "worked" in Flask requests because the connection pool pre-populates RLS from X-Org-ID headers. In Celery/background it would silently return default — exactly the failure class we're trying to eliminate.

Fix: `get_org_preference` now uses an admin connection and sets RLS directly from `org_id` (same pattern as `store_org_preference`). Extracted `_set_org_rls` helper so both share it.

server/utils/auth/stateless_auth.py
```diff
@@ -366,6 +376,80 @@ def store_user_preference(user_id: str, key: str, value: Any):
...
+
+def store_org_preference(org_id: str, key: str, value: Any, *, cursor=None) -> None:
+    """Upsert an org-scoped preference row.
+
+    Org-scoped preferences are stored with a synthetic user_id of
+    "__org__<org_id>" so they share the user_preferences table without
+    colliding with real user-scoped keys. When `cursor` is provided, the
+    caller is responsible for RLS context and commit; otherwise an admin
+    connection is opened here and RLS is configured from org_id directly.
+    """
```
beng360, line 410: if org_id is ever None, the `ON CONFLICT` clause (`WHERE org_id IS NOT NULL`) will silently insert duplicates instead of upserting. make sure you guard this

damianloch: Good call. Added explicit `if not org_id: raise ValueError(...)` guards at the top of both `store_org_preference` and `get_org_preference` in 060ed53d.

All current callers pass a non-None `org_id` (routes return 403 before calling when there's no org context), so nothing was actually reachable — but the helper shouldn't trust that, and failing loud here is better than the partial-index silently accepting duplicate NULL-org rows.

beng360 (approved):

PR #325, opened 2026-04-27
Adding security unit testing

OlivierTrudeau:
Summary
- Add 119 negative security tests across 5 files covering: allow-list bypass, prompt injection, LD_PRELOAD escapes, credential creation (key gen, passwd), and lateral movement (reverse shells, bind shells, firewall evasion)
- Each test passes when the security control (signature matcher or org denylist) blocks the malicious command
- Tests run in CI via a new "Run security tests" step in the pre-checks job

Out-of-scope fix included
- Widened the socat signature rule in signature_match.py from TCP: to TCP[-:] to also catch bind shells (socat TCP-LISTEN:4444 EXEC:/bin/sh). Previously only outbound reverse shells were matched at the static layer. Verified against the existing Sigma canary suite (no false positives).

## Summary by CodeRabbit

* **Tests**
  * Added a comprehensive security test suite covering credential controls, lateral movement, LD_PRELOAD/evasion, prompt-injection, mining, and allowlist/denylist layered checks; expanded test helpers, package stubbing, and test discovery support.

* **Bug Fixes**
  * Broadened detection for netcat/socat reverse-shell variants and improved image/process matching and LD_PRELOAD/evasion detections.

* **Chores**
  * CI now installs additional test dependencies and runs the new security test suites.


.github/workflows/linters.yml
```diff
@@ -53,3 +53,11 @@ jobs:
             echo "::warning::tests/architectural/test_rls_coverage.py not present yet; skipping."
           fi
           python -m pytest "${tests[@]}" -v
+
+      - name: Run unit tests
+        working-directory: server
+        run: python -m pytest tests/utils/test_log_sanitizer.py tests/utils/test_payload_timestamp.py -v
```
Harrio-6: Can move them to run right above as they are both unit tests

Harrio-6 (changes requested):

.github/workflows/linters.yml
OlivierTrudeau: good catch. I moved it

damianloch (approved):

Harrio-6 (approved):

PR #328, opened 2026-04-28
Interactive command gate: Yes / No / Yes-Always for foreground chats

damianloch:
## Problem

Our security layers (NeMo signature match, org allow/deny policy, LLM safety judge) block destructive commands outright. That's the right call for background RCAs and scheduled runs, but in a foreground chat the user is typing the request themselves -- a hard block feels broken, and they have no way to say "I know, run it anyway" or "stop asking me about this command."

## What this PR does

Adds a single command gate that runs the existing security layers behind one entry point and, for foreground chats only, prompts the user before a blocked command runs:

- **Yes** -- approve this invocation. Silent to the agent (looks like a normal success, we don't teach the LLM to reason about the gate).
- **No** -- abort with a distinct `USER_DECLINED` code so the agent sees an explicit user rejection rather than a static policy failure.
- **Yes, Always** -- only shown when the block came from the org allow/deny policy (i.e. there's a slot to flip). Applies an editable policy mutation so future runs -- including background RCAs -- inherit the decision.

Background chats and RCAs keep the existing hard-block behavior; no prompt, no bypass.

No new feature flag. The gate is just the interactive surface of the existing guardrails, so `GUARDRAILS_ENABLED` and the per-org list toggles continue to govern whether anything blocks (and therefore prompts) at all.

## How it's wired

**Backend**
- `server/utils/auth/command_gate.py` (new) -- unified gate. One call per tool (`terminal_exec`, `cloud_exec`, `kubectl_onprem`, `tailscale_ssh`) replaces the scattered per-layer checks.
- `server/utils/auth/command_policy.py` -- extended `CommandVerdict` with `deny_rule_id` and `allowlist_exhausted`; added `plan_yes_always` / `apply_yes_always` and `derive_pattern_from_command` so the gate can propose conservative regex mutations (falls back to a fully escaped literal for shell compound statements).
- Two `contextvars` short-circuit duplicate prompts and duplicate guardrail LLM calls when one shell tool routes into another (`terminal_exec` -> `cloud_exec`) or when `terminal_run._check_guardrails` would re-evaluate a command the gate just approved.
- `server/utils/cloud/infrastructure_confirmation.py` -- `wait_for_user_confirmation_ex` returns the full decision payload (`execute` / `execute_always` / `cancel` + edited patterns). The WS protocol is extended with `block_layer`, `yes_always_effect`, the `execute_always` option, and `edited_patterns`.

**Frontend**
- `ToolExecutionWidget.tsx` -- new `ConfirmationPanel` row with three visually distinct buttons (muted Deny, plain Allow, filled-blue Always). Always opens a popover with the proposed pattern editable before commit. Shows the real command in the header pre-approval, including for `cloud_exec`.
- `useMessageHandler.ts` threads the new fields through; `types.ts` gets `PolicyChange` / `YesAlwaysEffect` / richer `ToolCall`.
- `SecuritySettings.tsx` + `routes/command_policies.py` -- rules now require a non-empty description on both sides.

**Reload state model**
The gate introduces a live HITL state that needs to survive a page reload. We keep `chat_sessions.messages` append-only (history, never rewritten) and add a separate `pending_turn` JSONB column for the in-flight confirmation. The gate sets it before sending the prompt, clears it on any response or timeout, and the frontend renders it as a synthetic awaiting-confirmation tail card on load. This is what fixed the duplicate-tool-call / stuck-pending-card / missing-user-prompt issues that appeared during testing.

## Test plan

- Run a command blocked only by allow list -> Deny / Allow / Always all shown; Always adds an editable allow pattern.
- Run a command hitting a deny rule -> Always proposes disabling the rule.
- Run a command blocked by both (deny rule hit and allowlist exhausted) -> Always proposes both mutations; after apply, command actually runs.
- Run a command blocked by NeMo or the LLM safety judge -> Yes / No only, no Always (no policy slot to flip).
- Click Deny -> agent receives `USER_DECLINED`, not a policy code.

## Summary by CodeRabbit

* **New Features**
  * “Always” option for command approvals that can persist suggested policy changes from the approval UI.
  * Awaiting-confirmation UI now shows a summarized command, Deny/Allow actions, and an editable Yes‑Always workflow.

* **Improvements**
  * Pending command confirmations survive reloads and session recovery.
  * Command display prefers the gate-provided authoritative command while awaiting approval.
  * Create/update policy endpoints reject empty descriptions.
  * Stricter confirmation gating with session taint tracking for subsequent executions.

* **Bug Fixes**
  * Add‑rule form now disables Add unless pattern and description are non-empty.


client/src/components/tool-calls/ToolExecutionWidget.tsx
```diff
@@ -586,3 +551,151 @@ const ToolExecutionWidget = ({ tool, className, sendMessage, sendRaw, onToolUpda
...
+      session_id: sessionId,
+    }
+    if (decision === 'execute_always') payload.edited_patterns = edited
+    sendRaw(JSON.stringify(payload))
+    if (decision === 'cancel') {
+      onToolUpdate?.({ status: 'completed', output: 'Operation cancelled by user' })
+    } else {
+      onToolUpdate?.({ status: 'running' })
+    }
+  }
```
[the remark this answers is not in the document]
damianloch: Valid. Fixed in 0910cc23 — respond() now checks the `sendRaw` boolean and keeps the card in pending state if the socket send fails, so the UI no longer diverges from the server.

server/chat/backend/agent/tools/cloud_exec_tool.py
```diff
@@ -1409,25 +1409,22 @@ def cloud_exec(provider: str, command: str, user_id: Optional[str] = None, sessi
...
-        policy_cmd = f"{prefix} {command}" if prefix and not command.strip().startswith(prefix) else command
-        verdict = evaluate_compound_command(org_id, policy_cmd)
-        if not verdict.allowed:
-            reason = (verdict.rule_description or "Matched organization policy")[:200]
-            logger.warning("Policy denied cloud command for user %s (%s)",
-                            user_id, reason)
+        gated_cmd = f"{prefix} {command}" if prefix and not command.strip().startswith(prefix) else command
+        from utils.auth.command_gate import gate_command
+        gate = gate_command(user_id=user_id, tool_name="cloud_exec", command=gated_cmd)
+        if not gate.allowed:
```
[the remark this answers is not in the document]
damianloch: Not applying. The gate runs after `normalized_provider` is computed and prepends the correct CLI prefix (`aws`/`gcloud`/`az`/`scw`/`ovhcloud`) so patterns like `^aws\s+ec2\b` match the exact string the tool executes. `cloud_exec` runs subprocess directly rather than routing through `terminal_run._check_guardrails`, so the approved-hash fast path is irrelevant here. Non-default sub-CLIs (`kubectl`/`gsutil`/`bq`) reach this tool only after `terminal_exec` has already gated the full user-entered string, so they don't become fabricated `aws kubectl ...` entries in practice.

server/chat/backend/agent/tools/terminal_exec_tool.py
```diff
@@ -256,19 +256,16 @@ def terminal_exec(
...
+    if not gate.allowed:
+        logger.warning("terminal_exec blocked for user %s (%s): %s",
+                       user_id, gate.code, gate.block_reason[:200])
         return json.dumps({
             "success": False,
-            "error": f"Command blocked by organization policy: {reason}",
-            "code": "POLICY_DENIED",
+            "error": gate.block_reason,
+            "code": gate.code,
         })
```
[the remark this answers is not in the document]
damianloch: Not applying. Gating before routing is intentional: the user-typed string is what policy patterns like `^aws\s+ec2\b` must match. Moving the gate post-route would re-evaluate against a rewritten string and break existing allow rules. Yes-Always mutations are stored as org-wide command-pattern rows (not tool-scoped), so which tool name recorded the change is not a meaningful distinction here.

server/chat/backend/agent/workflow.py
```diff
@@ -979,8 +979,19 @@ async def stream(self, input_state: State):
...
+                # gets no special UI for this block -- it surfaces only when
+                # the agent next tries to run a tool.
+                if getattr(input_state, "is_background", False):
+                    yield ("token", "Your message was blocked by our safety system. Please rephrase your request.")
+                    return
+                from utils.auth.command_gate import mark_session_tainted
+                mark_session_tainted(
+                    getattr(input_state, "session_id", None),
+                    getattr(input_state, "user_id", None),
+                )
```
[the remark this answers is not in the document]
damianloch: Not applying — this is the intended behavior for this PR. The product requirement is that if NeMo's input rail fires in a foreground chat, we don't hard-block the session; instead we taint it so every subsequent tool call in that session is forced through the user approval gate (Yes/No, plus Always if policy also fires). Background chats still hard-block. The foreground model exposure is an accepted trade-off for keeping the conversational UX unblocked; the command gate is the enforcement point.

server/utils/auth/command_gate.py
```diff
@@ -0,0 +1,365 @@
...
+            cursor = conn.cursor()
+            cursor.execute(
+                "SELECT security_tainted FROM chat_sessions WHERE id = %s",
+                (session_id,),
+            )
+            row = cursor.fetchone()
+            return bool(row and row[0])
+    except Exception as e:
+        logger.warning(f"[CommandGate] taint lookup failed for {session_id}: {e}")
+        return False
```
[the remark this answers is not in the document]
damianloch: Valid. Fixed in 0910cc23. `is_session_tainted` now takes `user_id`, calls `set_rls_context` on the connection, and scopes the query with `AND user_id = %s` so it mirrors `mark_session_tainted`'s RLS behavior. Call site updated.

server/utils/auth/command_policy.py
```diff
@@ -286,6 +315,134 @@ def validate_pattern(pattern: str) -> Optional[str]:
...
+    parts = [tokens[0]]
+    for tok in tokens[1:]:
+        if tok.startswith("-"):
+            break
+        parts.append(tok)
+        # Stop after CLI + first subcommand to keep the pattern reasonably
+        # narrow. Callers (UI) can edit before applying.
+        if len(parts) >= 2:
+            break
+    return r"^" + r"\s+".join(re.escape(p) for p in parts) + r"\b"
```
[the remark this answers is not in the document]
damianloch: Partially valid — fixed the subcommand-length bug in 0910cc23. Raised the parts cap from 2 to 3 so `aws ec2 terminate-instances` matches the docstring example instead of collapsing to `^aws\s+ec2\b`. Not changing `_ENV_ASSIGN_RE`: `shlex.split(..., posix=True)` already unwraps quotes before we pattern-match tokens, so `KUBECONFIG="/tmp/my cfg" kubectl ...` becomes `['KUBECONFIG=/tmp/my', 'cfg', 'kubectl', ...]`. That's a corner case with a split env value, but the user can fix it in the editable regex field before applying — which is the whole point of the editable popover.

server/utils/auth/command_policy.py
```diff
@@ -286,6 +315,134 @@ def validate_pattern(pattern: str) -> Optional[str]:
...
+                    cur.execute(
+                        "INSERT INTO org_command_policies "
+                        "(org_id, mode, pattern, description, priority, updated_by, source) "
+                        "VALUES (%s, 'allow', %s, %s, %s, %s, 'custom') "
+                        "ON CONFLICT (org_id, mode, pattern, source) DO NOTHING",
+                        (org_id, ch.pattern, ch.description or "Auto-approved from chat",
+                         50, user_id),
+                    )
+        conn.commit()
+    invalidate_cache(org_id)
```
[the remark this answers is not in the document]
damianloch: Valid and important — fixed in 0910cc23. Added `_user_is_org_admin` in `command_gate.py` that enforces Casbin `(admin, access)`, the same check the HTTP policy routes use. The Yes-Always option is now hidden for non-admins (they see Yes/No only), and the server-side apply path also gates on the admin check as defense-in-depth so a crafted WS message can't bypass it. Non-admin approvals remain session-scoped and don't touch `org_command_policies`.

server/utils/cloud/infrastructure_confirmation.py
```diff
@@ -29,56 +43,110 @@ def _send_ws(payload: dict, tool_name: str):
...
             logger.debug(f"WEBSOCKET: Confirmation {confirmation_id} resolved with decision: {decision}")
         else:
             logger.warning(f"WEBSOCKET: No pending confirmation found for ID: {confirmation_id}")
-            
+
+        # Clear the durable live-state slot immediately on user response so a
+        # concurrent reload in another tab sees the resolved state rather
+        # than the stale prompt. The waiter's finally block also clears it,
+        # which makes this a safe no-op if it ran first.
+        _clear_pending_turn(session_id, user_id)
```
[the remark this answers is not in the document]
damianloch: Valid. Fixed in 0910cc23. Moved `_clear_pending_turn` inside the `if confirmation_id in _pending_confirmations` branch so a stale/late response for a superseded prompt no longer wipes the active slot.

server/utils/cloud/infrastructure_confirmation.py
```diff
@@ -168,98 +205,118 @@ def wait_for_user_confirmation(
...
+            "created_at": datetime.now().isoformat(),
+        })
+
+    _send_ws(payload, tool_name)
+
+    _pending_confirmations[confirmation_id] = {
+        'result': None,
+        'user_id': user_id,
+        'timestamp': time.time(),
+    }
```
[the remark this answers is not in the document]
damianloch: Valid. Fixed in 0910cc23. Swapped order in both `wait_for_user_confirmation_ex` and the legacy `wait_for_user_confirmation`: the waiter is now inserted into `_pending_confirmations` before `_send_ws`, which dispatches on a daemon thread. Also left a comment so future edits don't regress it.

server/utils/db/db_utils.py
```diff
@@ -1461,6 +1461,40 @@ def initialize_tables():
...
+                    ALTER TABLE chat_sessions
+                    ADD COLUMN IF NOT EXISTS security_tainted BOOLEAN NOT NULL DEFAULT false;
+                """)
+                logging.info(
+                    "Added security_tainted column to chat_sessions table (if not exists)."
+                )
+                conn.commit()
+            except Exception as e:
+                logging.warning(f"Error adding security_tainted column: {e}")
+                conn.rollback()
```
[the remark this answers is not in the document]
damianloch: Not applying. The migration uses `ADD COLUMN IF NOT EXISTS` which is idempotent and succeeds on re-runs, and the `log + rollback-on-exception` pattern is consistent with every other migration in this file (`pending_turn`, `llm_usage_tracking` surcharges, etc). Treating this one column differently would make the migration logic inconsistent. If an `ALTER TABLE IF NOT EXISTS` genuinely fails for non-idempotent reasons the DB is in a much bigger state where fail-fast would be appropriate across the board — which is a separate refactor, not scoped to this PR.

OlivierTrudeau (commented):
test

server/utils/cloud/infrastructure_confirmation.py
```diff
@@ -29,237 +43,206 @@ def _send_ws(payload: dict, tool_name: str):
...
+    """Cancel all pending confirmations for a given session."""
     if not session_id:
         return 0
-    
+
     cancelled_count = 0
     for confirmation_id, confirmation_data in _pending_confirmations.items():
-        # Only cancel if not already resolved
         if confirmation_data.get('result') is None:
             confirmation_data['result'] = 'cancel'
```
OlivierTrudeau, line 148: **Bug: `cancel_pending_confirmations_for_session` doesn't actually filter by session.**

This iterates _all_ entries in `_pending_confirmations` and cancels every unresolved one, regardless of which session it belongs to. If two users have concurrent HITL prompts on the same worker process, cancelling one session cancels the other's confirmation too.

The `session_id` parameter is accepted but never used for filtering — there's no `session_id` stored in the confirmation data dict to compare against.

**Fix:** Store `session_id` when registering the waiter (line 217-220), then filter here:

```python
# In wait_for_user_confirmation_ex, line ~217:
_pending_confirmations[confirmation_id] = {
    'result': None,
    'user_id': user_id,
    'session_id': session_id,   # <-- add this
    'timestamp': time.time(),
}

# Then here:
if confirmation_data.get('result') is None and confirmation_data.get('session_id') == session_id:
```

server/utils/auth/command_gate.py
```diff
@@ -0,0 +1,448 @@
...
+    if decision == "execute_always" and offer_always and org_id:
+        edited = result.get("edited_patterns") or {}
+        applied = []
+        for idx, ch in enumerate(yes_always_changes):
+            if ch.action == "add_allow_rule" and ch.editable:
+                # Users may tighten/loosen the pattern via the editable input.
+                # Indices are sent as strings by JSON.
+                override = edited.get(str(idx)) or edited.get(idx)
+                pattern = (override or ch.pattern or "").strip()
+                err = validate_pattern(pattern) if pattern else "empty pattern"
```
OlivierTrudeau, line 415: **Security: Yes-Always regex injection / ReDoS risk from client input.**

`validate_pattern` only checks that the string is syntactically valid regex. It does **not** limit regex complexity. A malicious user (or compromised frontend) can submit a catastrophic backtracking regex like `(a+)+$` as an allow-rule pattern via the `edited_patterns` payload, which would ReDoS every future `evaluate_command` call for the entire org.

**Suggested fix:** Add a max-length check and ideally a timeout-bounded test match in `validate_pattern`:

```python
def validate_pattern(pattern: str) -> Optional[str]:
    if len(pattern) > 500:
        return "pattern too long (max 500 chars)"
    try:
        compiled = re.compile(pattern)
        compiled.search("a" * 100)  # quick sanity check
        return None
    except re.error as exc:
        return str(exc)
```

Also worth forbidding patterns with nested quantifiers (e.g. reject if `(.*+)` or `(a+)+` is detected).

server/utils/cloud/infrastructure_confirmation.py
```diff
@@ -29,237 +43,206 @@ def _send_ws(payload: dict, tool_name: str):
...
-    Called when user cancels from chat input to unblock waiting confirmation threads.
-    Returns the number of confirmations cancelled.
-    """
+    """Cancel all pending confirmations for a given session."""
     if not session_id:
         return 0
-    
+
     cancelled_count = 0
     for confirmation_id, confirmation_data in _pending_confirmations.items():
```
OlivierTrudeau: **Thread safety: `_pending_confirmations` dict iteration can race.**

`_pending_confirmations` is a module-global dict mutated from multiple threads (Flask workers, Celery threads, WebSocket handler). Python's GIL makes single dict operations atomic on CPython, but iterating here while `wait_for_user_confirmation_ex` or `handle_websocket_confirmation_response` adds/removes entries concurrently can raise `RuntimeError: dictionary changed size during iteration`.

**Quick fix:** Copy before iterating:

```python
for confirmation_id, confirmation_data in list(_pending_confirmations.items()):
```

Longer term, consider a `threading.Lock` around all mutations to this dict.

server/utils/auth/command_gate.py
```diff
@@ -0,0 +1,448 @@
...
+
+    Foreground: prompts the user Yes / No with *summary* as the rendered
+    action. Background: denies (no interactive user). There is no
+    Yes-Always here -- these actions are not regex-addressable, so we
+    cannot persist an allow rule; org policy does not apply.
+
+    Returns the same :class:`GateDecision` shape as :func:`gate_command`
+    so callers can treat both gates uniformly.
+    """
+    if not user_id:
```
OlivierTrudeau, line 161: **Inconsistency: `gate_action` blocks when `user_id` is missing, but `gate_command` (line 132) allows through.**

The docstring says this preserves prior `wait_for_user_confirmation` behavior, which is correct, but the asymmetry is a footgun. If a tool is ever migrated from `gate_command` to `gate_action` (or vice versa), the no-user-id behavior silently flips.

**Suggestion:** At minimum, add a comment in the `gate_command` docstring cross-referencing this difference. Ideally, document in one place _why_ the two gates diverge on the no-user case so future contributors don't accidentally unify them.

server/utils/cloud/infrastructure_confirmation.py
```diff
@@ -29,237 +43,206 @@ def _send_ws(payload: dict, tool_name: str):
...
+    # the WS payload's ``data`` field so the frontend can build the synthetic
+    # tool card with identical semantics to the live WS path.
+    if session_id:
+        _set_pending_turn(session_id, user_id, {
+            "confirmation_id": confirmation_id,
+            "tool_name": tool_name,
+            "message": message,
+            "options": options,
+            **(extra or {}),
+            "created_at": datetime.now().isoformat(),
```
OlivierTrudeau, line 214: **Stale `pending_turn` on backend crash / Celery timeout.**

If the worker crashes or the Celery task times out while blocked in the polling loop (lines 231-239), the `finally` block at line 244 never runs. `pending_turn` stays set in the DB forever. On next session load, the frontend synthesizes a stale confirmation card that can never be resolved (the in-memory waiter is gone).

**Suggested fixes (pick one or both):**
1. Add a TTL check in `useChatHistory.ts` when synthesizing the pending card — if `pending.created_at` is older than e.g. 10 minutes, skip rendering it.
2. Add a periodic cleanup query: `UPDATE chat_sessions SET pending_turn = NULL WHERE pending_turn IS NOT NULL AND (pending_turn->>'created_at')::timestamp < NOW() - INTERVAL '10 minutes'`.

client/src/hooks/useChatHistory.ts
```diff
@@ -223,8 +228,36 @@ export function useChatHistory(): UseChatHistoryReturn {
       const rawMessages = data.messages || [];
       const cleanedMessages = cleanupStaleToolCalls(rawMessages, data.updated_at);
       const cleanupTime = performance.now() - cleanupStart;
+      // If the server has a live HITL confirmation pending for this session,
+      // append a synthetic bot message with an awaiting_confirmation tool
+      // call so the user sees the prompt immediately on reload. The card is
+      // driven entirely by chat_sessions.pending_turn -- history remains
+      // append-only and contains no mid-turn snapshot.
+      const pending = data.pending_turn;
+      const messagesWithPending: ChatMessage[] = pending && pending.confirmation_id
```
OlivierTrudeau, line 237: **Non-owner viewers see stale confirmation prompt.**

The `pending_turn` payload comes back from the backend for any user with access to the session (org RLS). The chat route returns `is_own`, but this code doesn't check ownership before rendering the synthetic confirmation card.

A non-owner viewer would see clickable Allow/Deny buttons. Clicking them would fail server-side (the WS handler validates `user_id`), but it's confusing UX.

**Suggestion:** Only render the pending card when the session belongs to the current user:

```typescript
const pending = data.is_own ? data.pending_turn : null;
```

server/utils/auth/command_gate.py
```diff
@@ -0,0 +1,448 @@
...
+
+
+_ALLOWED = GateDecision(allowed=True)
+
+
+def _block(code: str, reason: str) -> GateDecision:
+    return GateDecision(allowed=False, code=code, block_reason=reason)
+
+
+def _is_foreground() -> bool:
```
OlivierTrudeau: **Nit: `_is_foreground` and `_session_id` do redundant state lookups.**

Both call `get_state_context()` independently. In `_gate_impl` (lines 249-250) they're invoked back-to-back, meaning two separate thread-local lookups for the same object.

**Suggestion:** Refactor to a single helper:

```python
def _get_context() -> tuple[bool, Optional[str]]:
    try:
        from utils.cloud.cloud_utils import get_state_context
        state = get_state_context()
        if state is None:
            return False, None
        return not bool(getattr(state, "is_background", False)), getattr(state, "session_id", None)
    except Exception:
        return False, None
```

OlivierTrudeau (commented):
Review comments on security, correctness, and scalability. Skipping cosmetic nits.

OlivierTrudeau:
**Minor: Policy cache in `server/utils/auth/command_policy.py` is unbounded.**

`_cache` (line 37) is a plain dict with no size limit. With many orgs, entries accumulate indefinitely. The 30s TTL prevents serving stale data, but expired entries are never evicted — they're only replaced on next read.

**Suggestion:** Use `functools.lru_cache` or add periodic eviction. Even a simple size check in `_get_cached` would help:

```python
if len(_cache) > 500:
    _cache.clear()
```

server/utils/auth/command_gate.py
```diff
@@ -0,0 +1,448 @@
...
+        # Re-entry for the same command (e.g. terminal_exec -> cloud_exec).
+        return _ALLOWED
+
+    token = _gate_inflight_command.set(cmd_hash)
+    try:
+        return _gate_impl(user_id=user_id, tool_name=tool_name, command=command,
+                          cmd_hash=cmd_hash)
+    finally:
+        _gate_inflight_command.reset(token)
+
```
OlivierTrudeau, line 143: `_gate_inflight_command` is properly reset here, but `_guardrails_approved_command` (set on lines 268 and 322) is never reset. The hash lingers for the lifetime of the thread/context, so a retried command on the same thread skips `_check_guardrails` using a stale approval from a previous invocation — even if org policy changed in between.

Should mirror the same pattern:

```python
approved_token = _guardrails_approved_command.set(None)
try:
    return _gate_impl(...)
finally:
    _gate_inflight_command.reset(token)
    _guardrails_approved_command.reset(approved_token)
```

The dedup still works because `terminal_run` is called synchronously before `gate_command` returns to its caller.

client/src/components/tool-calls/ToolExecutionWidget.tsx
```diff
@@ -586,3 +551,152 @@ const ToolExecutionWidget = ({ tool, className, sendMessage, sendRaw, onToolUpda
...
+
+  const respond = (decision: 'execute' | 'cancel' | 'execute_always') => {
+    const confirmationId = tool.confirmation_id
+    if (!confirmationId || !sendRaw || !userId) return
+    const payload: Record<string, unknown> = {
+      type: 'confirmation_response',
+      confirmation_id: confirmationId,
+      decision,
+      user_id: userId,
+      session_id: sessionId,
```
OlivierTrudeau, line 604: user_id should be handled server side to avoid injection and impersonation

server/utils/cloud/infrastructure_confirmation.py
```diff
@@ -29,237 +43,206 @@ def _send_ws(payload: dict, tool_name: str):
...
+    except Exception as e:
+        logger.warning(f"Failed to clear pending_turn for session {session_id}: {e}")
+
+
 async def handle_websocket_confirmation_response(data: dict):
     """Handle incoming WebSocket confirmation responses from the frontend."""
     try:
         confirmation_id = data.get('confirmation_id')
         decision = data.get('decision')
         user_id = data.get('user_id')
```
OlivierTrudeau, line 102: Here, shouldn't trust the user_id coming from the data. That can be impersonated

OlivierTrudeau: I think this is also true in main_chatbot line 986. It might be out of the scope of the PR, so if it is, maybe just make sure we make a ticket and we hit this bc it's not safe

OlivierTrudeau (changes requested):

server/chat/backend/agent/tools/cloud_exec_tool.py
```diff
@@ -1228,25 +1156,21 @@ def _cloud_exec_aws_multi_account(
...
+    _gate = gate_command(user_id=user_id, tool_name="cloud_exec", command=_gated)
+    if not _gate.allowed:
+        return json.dumps({
+            "success": False,
+            "error": _gate.block_reason,
+            "code": _gate.code,
+            "multi_account": True,
+            "command": command,
+            "provider": "aws",
+        })
```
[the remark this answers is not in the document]
damianloch: Fixed — removed the second `gate_command` call from `_cloud_exec_aws_multi_account`. The outer `cloud_exec()` already gates before fanning out so this was a duplicate prompt.

server/chat/backend/agent/tools/cloud_exec_tool.py
```diff
@@ -1409,25 +1333,22 @@ def cloud_exec(provider: str, command: str, user_id: Optional[str] = None, sessi
...
         prefix = _CLI_PREFIX.get(provider.lower(), "")
-        policy_cmd = f"{prefix} {command}" if prefix and not command.strip().startswith(prefix) else command
-        verdict = evaluate_compound_command(org_id, policy_cmd)
-        if not verdict.allowed:
-            reason = (verdict.rule_description or "Matched organization policy")[:200]
-            logger.warning("Policy denied cloud command for user %s (%s)",
-                            user_id, reason)
+        gated_cmd = f"{prefix} {command}" if prefix and not command.strip().startswith(prefix) else command
+        from utils.auth.command_gate import gate_command
+        gate = gate_command(user_id=user_id, tool_name="cloud_exec", command=gated_cmd)
```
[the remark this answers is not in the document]
damianloch: Fixed — added `"tailscale": "tailscale"` to `_CLI_PREFIX` so Tailscale subcommands are prefixed before policy matching.

server/utils/auth/command_gate.py
```diff
@@ -0,0 +1,448 @@
...
+    Returns the same :class:`GateDecision` shape as :func:`gate_command`
+    so callers can treat both gates uniformly.
+    """
+    if not user_id:
+        # Preserve prior behavior of wait_for_user_confirmation helpers,
+        # which required a user and otherwise denied.
+        return _block("USER_DECLINED", "Tool call not allowed by user")
+
+    if not _is_foreground():
+        return _block("USER_DECLINED", "Tool call not allowed by user")
```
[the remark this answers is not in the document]
damianloch: Fixed — `gate_action` now returns `TOOL_NOT_ALLOWED` for missing user context and `BACKGROUND_DENIED` for background path. `USER_DECLINED` is reserved for explicit user rejections only.

server/utils/auth/command_gate.py
```diff
@@ -0,0 +1,448 @@
...
+    if safety_blocked:
+        reasons.append(f"safety guardrail: {safety_decision.reason}")
+    if policy_blocked:
+        reasons.append(
+            "organization policy: "
+            + (policy_verdict.rule_description or "matched organization policy")[:200]
+        )
+    if not reasons and tainted:
+        reasons.append("session flagged by input safety check; approval required")
+    block_reason = "Command blocked by " + "; ".join(reasons)
```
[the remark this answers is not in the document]
damianloch: Fixed — `session_taint` is now unconditionally appended to `layers` and `reasons` whenever `tainted=True`, not only when it is the sole blocker.

server/utils/auth/command_policy.py
```diff
@@ -286,6 +315,136 @@ def validate_pattern(pattern: str) -> Optional[str]:
...
+        if tok.startswith("-"):
+            break
+        parts.append(tok)
+        # Stop after CLI + up to two subcommand tokens so multi-word
+        # subcommands like "aws ec2 terminate-instances" match the docstring
+        # example rather than collapsing to "^aws\s+ec2\b". Users can still
+        # tighten or loosen in the editable UI field.
+        if len(parts) >= 3:
+            break
+    return r"^" + r"\s+".join(re.escape(p) for p in parts) + r"\b"
```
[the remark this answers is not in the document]
damianloch: Fixed — after popping `sudo`, the loop now also consumes its flags and flag-values (e.g. `-u root`, `--user=root`) until reaching the actual CLI token. Also updated `_ENV_ASSIGN_RE` to match values with spaces (post-shlex unquoting).

server/utils/cloud/infrastructure_confirmation.py
```diff
@@ -29,237 +43,206 @@ def _send_ws(payload: dict, tool_name: str):
...
+            "created_at": datetime.now().isoformat(),
+        })
 
-    # Store the confirmation request (without asyncio.Event for simplicity)
     _pending_confirmations[confirmation_id] = {
         'result': None,
         'user_id': user_id,
-        'timestamp': time.time()
+        'timestamp': time.time(),
     }
```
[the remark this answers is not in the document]
damianloch: Fixed — `session_id` is now stored on each `_pending_confirmations` entry at insertion time. `cancel_pending_confirmations_for_session` was also fixed in the same commit to actually filter by it (it was previously ignoring the parameter entirely and cancelling all pending confirmations across all sessions).

server/utils/cloud/infrastructure_confirmation.py
damianloch: Good catch — the function was never using `session_id` to filter, it was cancelling every pending confirmation regardless of session. Fixed by storing `session_id` in each entry dict at creation time and filtering on it in the cancel loop.

server/utils/auth/command_gate.py
damianloch: Fixed in `validate_pattern`: added a 500-char length cap and a pre-compile check for nested quantifier patterns (`(a+)+`, `(.*){3,}` etc.) that are the common ReDoS vectors. The attack surface is already limited to org admins, but the heuristic blocks the obvious cases. A full automaton-based complexity analyzer would be a more thorough follow-up.

server/utils/cloud/infrastructure_confirmation.py
damianloch: Fixed — now iterating `list(_pending_confirmations.items())` to snapshot before iterating, preventing RuntimeError if another thread mutates the dict concurrently.

server/utils/auth/command_gate.py
damianloch: The asymmetry is intentional and documented: `gate_command` allows through with no user (preserving existing per-tool auth behavior) whereas `gate_action` denies (preserving prior `wait_for_user_confirmation` behavior which required a user). Added a cross-reference comment in the `gate_command` docstring pointing to this difference so it's explicit for future contributors.

server/utils/cloud/infrastructure_confirmation.py
damianloch: Valid concern. The right fix is a TTL check in `useChatHistory.ts` — skip rendering the pending card if `pending.created_at` is older than ~10 minutes. Tracking as a follow-up; out of scope for this PR since it requires a crash scenario to trigger.

client/src/hooks/useChatHistory.ts
damianloch: Fixed — `pending_turn` is now only consumed when `data.is_own` is true: `const pending = data.is_own ? data.pending_turn : null`. Non-owner viewers no longer see the interactive Allow/Deny card.

server/utils/auth/command_gate.py
damianloch: Fixed — merged both helpers into a single `_get_context() -> tuple[bool, Optional[str]]` that does one `get_state_context()` call and returns `(is_foreground, session_id)`. All call sites updated.

server/utils/auth/command_gate.py
damianloch: Fixed — `gate_command` now resets `_guardrails_approved_command` in its `finally` block alongside `_gate_inflight_command`, preventing the stale hash from carrying over to a retried command on the same thread.

client/src/components/tool-calls/ToolExecutionWidget.tsx
damianloch: Valid — the WS handler should derive `user_id` from the server-side session rather than trusting the client payload. This pattern predates this PR and is shared with `main_chatbot.py` so fixing it properly requires a broader WS auth refactor. Tracking as a follow-up ticket; not changing scope here.

server/utils/cloud/infrastructure_confirmation.py
damianloch: Agreed — see comment above. Fixing the full WS auth chain is tracked as a follow-up; the correct fix is to resolve `user_id` from the authenticated session token server-side rather than echoing the client-supplied value.

damianloch: Yes, same pattern at that line as well. All three are in scope for the same follow-up ticket — fixing `user_id` resolution from the server session across the entire WS handler chain.

server/routes/auth_routes.py
```diff
@@ -140,6 +140,12 @@ def register():
                 
                 logging.info(f"New user registered: {email[:3]}***@*** (role=admin, org={org_id})")
 
+                try:
+                    from utils.auth.command_policy import seed_default_command_policy
+                    seed_default_command_policy(org_id, user_id)
+                except Exception as policy_err:
+                    logging.warning(f"Failed to seed command policy for org {org_id}: {policy_err}")
```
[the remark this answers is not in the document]
damianloch: Fixed — both call sites now use `logging.warning(..., exc_info=policy_err)` to preserve the full stack trace.

server/utils/auth/command_policy.py
```diff
@@ -627,3 +789,52 @@ def get_policy_templates() -> List[dict]:
...
+    Called immediately after org creation so every org is protected from day
+    one without requiring any admin action. Uses an independent admin
+    connection (the org INSERT has already committed by the time this runs).
+    Idempotent: skips insertion if the org already has policy rows (e.g.
+    retried registration).
+    """
+    from utils.db.connection_pool import db_pool
+    from utils.auth.stateless_auth import store_org_preference
+
+    tpl = get_policy_templates()[0]  # observability_only
```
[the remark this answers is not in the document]
damianloch: Fixed — changed to `next(t for t in get_policy_templates() if t["id"] == "observability_only")` with a hard `ValueError` raise if the template isn't found, so ordering changes can never silently seed the wrong template.

server/chat/backend/agent/tools/bitbucket/utils.py
```diff
@@ -160,19 +153,13 @@ def build_cancelled_response() -> str:
...
-        session_id=session_id,
-    )
-    if not approved:
-        return build_cancelled_response()
-    return None
+    from utils.auth.command_gate import gate_action
+
+    if gate_action(user_id=user_id, tool_name=tool_name, summary=message).allowed:
+        return None
+    return build_cancelled_response()
```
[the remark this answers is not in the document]
damianloch: The `confirm_or_cancel` return type is `Optional[str]` — callers check `is None` for approved, not a structured code. Propagating `gate_result.code` through the JSON blob would require changes to all Bitbucket tool callers and is out of scope for this PR. The `USER_DECLINED` contract on shell commands via `gate_command` is unchanged. Closing.

OlivierTrudeau (approved):

PR #335, opened 2026-04-29
fix(guardrails): narrow input rail to webhook-authored text on background RCAs

damianloch:
## Summary

PR #329 narrowed the input rail for interactive RCA chat turns so it evaluates only the user-authored portion of the message (via `State.question`) instead of the full prefixed `HumanMessage`. The same underlying issue exists for webhook-triggered background RCAs: each `build_*_rca_prompt()` function synthesizes a ~4KB instruction scaffold around the external alert payload and then passes the entire scaffold as both `messages[0].content` and `State.question`. Stricter judge models (for example OpenAI's small models) correctly read the embedded investigation instructions as prompt-injection-shaped content and block the turn, preventing the RCA from ever starting.

This PR applies the same principle as #329: the input rail evaluates only the attacker-controllable surface. Webhook payload fields (title, status, message/description) are externally-authored input; the instruction scaffold is internal Aurora content and must not be rail-checked.

## Changes

- Add `build_alert_rail_text()` helper that extracts title, status, and message from the `alert_details` dict every builder already constructs.
- `build_rca_prompt()` now returns `(prompt, rail_text)`. Every per-source `build_*_rca_prompt()` propagates the tuple (they all delegate to `build_rca_prompt`), and the thin call-site helpers in `jenkins/spinnaker/bigpanda` forward the tuple.
- `run_background_chat()` and `_execute_background_chat()` accept an optional `rail_text` kwarg. When provided it populates `State.question` so `workflow._get_input_rail_text()` feeds the rail only the webhook-authored subset; `messages[0]` still carries the full synthesized prompt so the agent sees the normal instruction scaffold.
- Every webhook trigger site (pagerduty, grafana, datadog, dynatrace, netdata, opsgenie, splunk, spinnaker, jenkins/cloudbees, bigpanda, newrelic, incidentio, and the chat-initiated `trigger_rca_tool`) unpacks the tuple and forwards `rail_text` into `run_background_chat.delay`.
- Legacy callers that pass raw user input (slack, google chat, incidents routes, prediscovery, Jira follow-ups) are unchanged: `rail_text` defaults to `None` and `State.question` falls back to `initial_message`, preserving existing behavior.

## Security

The rail still evaluates every byte of externally-sourced content. It no longer evaluates Aurora's own instruction scaffolding, which was producing false positives but carries no attacker-controlled input. Fail-closed semantics in `server/guardrails/input_rail.py` are untouched.

## Replaces

Supersedes and closes #332, which attempted to fix the same symptom by skipping the input rail entirely for background tasks. That approach bypassed prompt-injection detection for all webhook payload content and is not correct.

## Test plan

- [x] Fire PagerDuty V3 webhook with main/RCA on direct OpenAI `gpt-5.2` and guardrails on `anthropic/claude-haiku-4.5`.
- [x] Confirm input rail passes (no `GUARDRAIL_AUDIT blocked` event).
- [x] Confirm `UserMessage` in nemoguardrails logs is exactly the alert title + status + body (~175 chars) instead of the prior ~4KB scaffold.
- [x] Confirm the agent proceeds into a full investigation (`query_datadog`, `query_newrelic`, `load_skill`, etc. tool calls).
- [x] Confirm incidents table transitions from `investigating/running` through tool calls.
- [ ] Regression: re-run at least one non-RCA `run_background_chat` path (slack button, incidents API chat) to confirm the default-None `rail_text` fallback preserves prior semantics.

### DIFF
--- a/server/chat/background/rca_prompt_builder.py
+++ b/server/chat/background/rca_prompt_builder.py
@@ -29,6 +29,29 @@ RCA_SEGMENTS_DIR = os.path.normpath(
 )
 
 
+def build_alert_rail_text(alert_details: Dict[str, Any]) -> str:
+    """Extract the webhook-authored subset of an alert for input-rail evaluation.
+
+    Synthesized RCA prompts wrap externally-controlled fields (alert title,
+    status, message/description) in a large instruction scaffold. The scaffold
+    is not user input and must not be fed to the prompt-injection rail (it
+    produces false positives with stricter models). This helper returns only
+    the webhook-provided text so the rail evaluates exactly the attacker-
+    controllable surface.
+    """
+    parts: List[str] = []
+    title = alert_details.get('title')
+    if isinstance(title, str) and title.strip():
+        parts.append(title.strip())
+    status = alert_details.get('status')
+    if isinstance(status, str) and status.strip() and status.strip().lower() != 'unknown':
+        parts.append(f"Status: {status.strip()}")
+    message = alert_details.get('message')
+    if isinstance(message, str) and message.strip():
+        parts.append(message.strip())
+    return "\n\n".join(parts)
+
+
 @lru_cache(maxsize=32)
 def _load_rca_segment_template(segment_name: str) -> str:
     """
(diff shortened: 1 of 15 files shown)


server/chat/background/rca_prompt_builder.py
```diff
@@ -29,6 +29,29 @@
...
+        parts.append(f"Status: {status.strip()}")
+    message = alert_details.get('message')
+    if isinstance(message, str) and message.strip():
+        parts.append(message.strip())
+    return "\n\n".join(parts)
+
+
 @lru_cache(maxsize=32)
 def _load_rca_segment_template(segment_name: str) -> str:
     """
```
beng360, line 57: just to note like build_alert_rail_text some providers have additional webhook controlled fields that go into prompt (like pduty urgency, grafana labels, other). rails wont catch it, but for now because these are structured it shouldnt matter

damianloch: yea this is known limitation

beng360 (approved):

PR #350, opened 2026-05-02
fix: kubectl agent token lookup blocked by RLS

beng360:
## Summary

- Kubectl agent connections were always rejected with "Invalid token" even when the token existed in the database
- Root cause: `kubectl_agent_tokens` has `FORCE ROW LEVEL SECURITY` enabled, but the WebSocket handler (`agent_ws_handler.py`) queried it without any org context — this is a bootstrap auth query where the agent authenticates with a Bearer token before any org is known
- Every SELECT returned 0 rows because the `select_by_org` RLS policy requires `myapp.current_org_id` to be set
- Fix: add permissive RLS policies keyed on `myapp.kubectl_token_resolve` session variable (same pattern as `mcp_tokens`), and set the flag in the handler before the token lookup

## Test plan

- [ ] Deploy to staging and verify kubectl agent connects successfully (currently in reconnect loop getting "Invalid token")
- [ ] Verify token lookup works: agent pod logs should show "Connected to backend" instead of "Connection closed by server"
- [ ] Verify the agent appears as connected on the infrapoo.org Kubernetes connectors page
- [ ] Verify existing kubectl token CRUD from the UI still works (org-scoped RLS still enforced for normal requests)

### DIFF
--- a/server/utils/db/db_utils.py
+++ b/server/utils/db/db_utils.py
@@ -2139,6 +2139,30 @@ def initialize_tables():
             """)
             logging.info("Added MCP token resolve RLS policies on mcp_tokens.")
 
+            # kubectl agent token verification is a bootstrap auth query — the
+            # agent connects with a Bearer token before any org context exists.
+            # Same pattern as mcp_tokens: a permissive policy keyed on a session
+            # variable so the handler can opt in to a scoped RLS bypass.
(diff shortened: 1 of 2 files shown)


beng360:
@claude review

Zarlanx (approved):

PR #351, opened 2026-05-02
fix(auth): retry JWT refresh on /org/switching so cookie actually updates (DEV-1100)

Zarlanx:
## Summary
- Fixes [DEV-1100](https://linear.app/arvoai/issue/DEV-1100/qa-when-inviting-an-existing-member-they-cant-do-anything-after).
- The post-invite-accept refresh on `/org/switching` could leave the user's JWT cookie pointing at their previous org. Casbin had already swapped their role to the new org, so every API call from the stale cookie returned 403 — the "can't do anything after accepting" symptom (and re-logging in fixed it because a fresh sign-in re-derives `orgId` from `users.org_id`).

## Root cause (two compounding things)

**1. `/org/switching` POST'd `/api/auth/session` once, then GET-polled.** A POST sets `trigger === "update"` in the JWT callback, which calls `refreshUserFromBackend()` → `/api/auth/me`. Right after `join_org` finishes its Casbin save + `_cleanup_empty_org` + audit-logging burst, that call can briefly take >3s and lose to the `Promise.race` timeout (`auth.ts:50`). The subsequent GET-polls don't carry `trigger === "update"`, so they short-circuit on the stale token and the loop exits at attempt 0 (orgName is non-default), the page navigates to `/`, and every Casbin call 403s.

**2. The JWT callback advanced `lastRefreshedAt` even on failure** (`auth.ts:151`), so for the next 60s the time-based revalidation couldn't retry. The stale cookie persisted until manual re-login.

## Fix

- **`/org/switching/page.tsx`**: snapshot `priorOrgId` once, then loop up to 8 POSTs to `/api/auth/session` (each forces `trigger === "update"`, bypassing the `lastRefreshedAt` throttle). Exit when the cookie's `orgId` actually differs from the snapshot.
- **`auth.ts`**: only stamp `lastRefreshedAt` on success, so a transient timeout doesn't block retries.

Single-file frontend changes, no schema, no backend changes — Casbin already does the right thing.

## Why #348 didn't fix it
PR #348 only changed the loop's exit condition (require `orgId !== prior`) without changing whether the cookie actually updates. Under the timeout failure mode, #348 would just throw `"session did not update"` and bounce the user to `/sign-in` — not what was observed (still landed on `/` with 403 storm).

## Test plan

Verified end-to-end with `/tmp/dev1100-repro.sh`:

```
1. Register User A → admin of OrgA
2. Register User B → admin of OrgB (real org, not Default)
3. Sign in both, sanity: B/orgB → /api/llm-usage/models = 200
4. A invites B (POST /api/admin/users {email,role:editor})
5. B accepts (POST /api/orgs/my-invitations/{id} {action:accept})
6. Simulate /org/switching: loop POST /api/auth/session
7. Assert polled orgId flips to orgA
8. B → /api/llm-usage/models, /api/incidents, /api/connectors/status,
   /api/orgs/current, /api/chat-sessions  →  all 200
```

Manual checklist for reviewer:
- [ ] `make dev`
- [ ] Register two users into two distinct real orgs (not Default)
- [ ] As A, invite the existing B by email (Members → Add Member)
- [ ] As B, accept the invite from Invitations panel
- [ ] Verify B lands on `/` and can act in OrgA without 403s
- [ ] Regression: brand-new user via Default → real-org invite still works (this was the path #348 wasn't actually changing for either)

Closes #348.

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable organization switching: session updates now detect actual org changes and use a bounded retry POST loop for consistent state synchronization.
  * Improved session/token refresh behavior: backend refresh requests handle failures more gracefully, update stored token timestamps only on confirmed outcomes, and clear user identity when the user is reported missing.

### DIFF
--- a/client/src/auth.ts
+++ b/client/src/auth.ts
@@ -140,6 +140,7 @@ export const { handlers, signIn, signOut, auth } = NextAuth({
           token.id = undefined
           token.email = undefined
           token.name = undefined
+          token.lastRefreshedAt = now
           return token
         }
         if (fresh) {
@@ -147,8 +148,15 @@ export const { handlers, signIn, signOut, auth } = NextAuth({
           token.orgId = fresh.orgId
           token.orgName = fresh.orgName
           token.mustChangePassword = fresh.mustChangePassword
+          token.lastRefreshedAt = now
         }
-        token.lastRefreshedAt = now
+        // On failure (fresh === null, typically the 3s timeout after a busy
+        // backend write burst), do NOT advance lastRefreshedAt — otherwise
+        // the next 60s of GET /api/auth/session calls short-circuit on the
+        // stale token, leaving the user with the old org_id in their cookie
+        // until the time-based check re-fires. This was the root cause of
+        // DEV-1100: the post-invite-accept refresh would lose the race once
+        // and then never re-attempt within the /org/switching polling window.
       }
 
       return token
--- a/client/src/auth.ts
+++ b/client/src/auth.ts
@@ -153,13 +153,6 @@ export const { handlers, signIn, signOut, auth } = NextAuth({
           token.mustChangePassword = fresh.mustChangePassword
           token.lastRefreshedAt = now
         }
-        // On failure (fresh === null, typically the 3s timeout after a busy
-        // backend write burst), do NOT advance lastRefreshedAt — otherwise
-        // the next 60s of GET /api/auth/session calls short-circuit on the
-        // stale token, leaving the user with the old org_id in their cookie
-        // until the time-based check re-fires. This was the root cause of
-        // DEV-1100: the post-invite-accept refresh would lose the race once
-        // and then never re-attempt within the /org/switching polling window.
       }
 
       return token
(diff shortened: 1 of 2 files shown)


client/src/auth.ts
```diff
@@ -140,15 +143,23 @@ export const { handlers, signIn, signOut, auth } = NextAuth({
...
+          token.lastRefreshedAt = now
         }
-        token.lastRefreshedAt = now
+        // On failure (fresh === null, typically the 3s timeout after a busy
+        // backend write burst), do NOT advance lastRefreshedAt — otherwise
+        // the next 60s of GET /api/auth/session calls short-circuit on the
+        // stale token, leaving the user with the old org_id in their cookie
+        // until the time-based check re-fires. This was the root cause of
+        // DEV-1100: the post-invite-accept refresh would lose the race once
+        // and then never re-attempt within the /org/switching polling window.
```
beng360: do we need to comment this ?

beng360 (dismissed):

client/src/auth.ts
Zarlanx: no it was me trying to understand

beng360 (approved):

PR #359, opened 2026-05-04
feat(github): GitHub App migration (dual-mode with OAuth)

Zarlanx:
Adds a public GitHub App as an alternative authentication path for the GitHub connector, running in permanent dual-mode alongside the existing OAuth flow.

Highlights
- DB schema: github_installations, user_github_installations (many-to-many), webhook_deliveries (idempotency), and nullable installation_id on github_connected_repos. All idempotent DDL.
- Vault-backed config: app private key + webhook secret read from aurora/system/github-app/* with env fallback for the webhook secret.
- JWT mint utility (RS256, iss=client_id, 9-min expiry, 60s iat backdate) using PyJWT.
- Installation token cache with per-installation threading lock so concurrent workers do not stampede the GitHub access_tokens endpoint.
- Auth router that prefers GitHub App installation tokens and falls back to OAuth gracefully (including on suspended installations); used by user-repo endpoints, repo selection, repo metadata Celery task, and the agent tools.
- /github/app/install + setup callback that verifies the installation_id with the GitHub API before any DB write (anti-spoofing); plus list and unlink endpoints. Templates render via Jinja autoescape.
- /github/webhook endpoint with HMAC-SHA256 signature validation (constant-time compare on raw body), X-GitHub-Delivery dedupe via the webhook_deliveries unique constraint, and async Celery dispatch.
- Celery handlers for installation*, installation_repositories, pull_request, issues, deployment, deployment_status, workflow_run, check_run, check_suite. push and release are intentionally excluded.
- Frontend: dual-CTA on the integration page (Install App + Connect OAuth), installation list with multi-installation picker, and status banners for suspended / pending-permissions / no-repos states. App CTA fetches the install URL from the backend so state=<user_id> is included.
- Pre-flight env validation at startup: enabled vs degraded mode, with a 503 fallback on App routes when GITHUB_APP_* env vars are missing.
- Observability: structured key=value logs across token-mint and webhook paths, with a token/JWT redaction helper applied to error logs.
- Tests: pytest scaffold + targeted unit tests for JWT mint, installation token cache (race + cache hit + 404 + suspended), webhook signature (tampered/wrong-secret/missing/replay), install callback (spoof + state
  + persistence + idempotency + XSS), and auth router priority/fallback.
- Docs: README two-ways-to-connect overview, SETUP_GITHUB_APP.md operator walkthrough, and docs/oss/GITHUB_APP_MIGRATION.md migration runbook.
- Helper: server/scripts/sign_webhook.py for signed webhook delivery testing during operations.

Out of scope (explicit guardrails preserved)
- No user access tokens / act-as-user flow.
- No push or release event handlers.
- No automatic OAuth -> App migration; users self-install.
- No GitHub Enterprise Server compatibility.
- No Redis-backed token cache (in-memory + per-installation lock only).
- No manifest-based app creation tooling.
- Existing OAuth flow code paths are unchanged.

## Summary by CodeRabbit

* **New Features**
  * GitHub App support added (preferred when available): install, manage, unlink, and per-installation repo selection with installation-aware repo picker and updated connection CTAs.
  * Server-side GitHub App webhook ingestion and processing; CLI to sign/send test webhooks.

* **Documentation**
  * Operator runbooks: GitHub App setup and OAuth→App migration guidance; connector README updated.

* **Tests**
  * Expanded tests for App install callback, JWT/token behavior, webhook signature, and related flows.


docs/oss/GITHUB_APP_MIGRATION.md
damianloch, line 1: I think it makes sense to have all our docs on the docsite instead of spreading them

docker-compose.yaml
```diff
@@ -282,6 +290,15 @@ services:
       AURORA_VERIFY_CLI_IDENTITY: ${AURORA_VERIFY_CLI_IDENTITY}
       AURORA_CACHE_TOKEN_IN_REDIS: ${AURORA_CACHE_TOKEN_IN_REDIS}
 
+      # GitHub App (alternative to OAuth) — webhook secret intentionally
+      # NOT passed to the worker. The worker handles already-verified
+      # deliveries; HMAC validation lives in the route handler.
+      GITHUB_APP_ID: ${GITHUB_APP_ID}
+      GITHUB_APP_CLIENT_ID: ${GITHUB_APP_CLIENT_ID}
+      GITHUB_APP_WEBHOOK_URL: ${GITHUB_APP_WEBHOOK_URL}
+      GITHUB_APP_SETUP_URL: ${GITHUB_APP_SETUP_URL}
```
damianloch, line 299: Why add these as env variables?

damianloch: this comment applies to GITHUB_APP_WEBHOOK_URL and GITHUB_APP_SETUP_URL in all files and locations

docker-compose.yaml
damianloch, line 1: looks like you forgot to add these vars to airtight docker compose

server/connectors/github_connector/SETUP_GITHUB_APP.md
```diff
@@ -0,0 +1,347 @@
...
+  secret='<paste-the-openssl-rand-output-from-step-2>'
+```
+
+Verify both writes:
+
+```bash
+docker exec aurora-vault-1 vault kv get -field=pem aurora/system/github-app/private-key | head -1
+# Expect: a PEM private-key header line (BEGIN ... PRIVATE KEY)
+
+docker exec aurora-vault-1 vault kv get -field=secret aurora/system/github-app/webhook-secret | wc -c
```
damianloch, line 201: For k8s deployments we allow users to point secrets to k8s secrets instead of referrign to values.yaml. For VM deployment I think we should keep the usual .env path and not make it inconsistent with manual vault writes.

.env.example
damianloch, line 1: values.example.yaml needs to reflect this

server/requirements.txt
```diff
@@ -102,7 +102,13 @@ python-docx>=1.1.0
...
+pytest>=7.4
+responses>=0.24
+cryptography>=43.0.1  # CVE-2023-50782, CVE-2024-26130 fixed in >=43
 
 # MCP Server (Model Context Protocol)
 mcp[cli]>=1.27.0
+
+# GitHub App authentication (JWT mint + installation tokens)
+PyGithub>=2.5  # advisories on <2.5
+PyJWT[crypto]>=2.8
```
damianloch, line 114: replace line 12 with PyJWT[crypto]>=2.8.0 rather than adding a second entry. Already have this one

server/tests/test_github_app_install_callback.py
damianloch, line 1: These tests all merit subdirs we are sort of bombing the root test dir

server/tests/test_sanity.py
damianloch, line 1: do we really need this lmao

server/connectors/github_connector/SETUP_GITHUB_APP.md
damianloch, line 1: this should proabably be incorporated into docsite as well somehow

server/connectors/github_connector/README.md
damianloch, line 1: reflect on docsite

client/src/components/github-provider-integration.tsx
damianloch, line 1: All requests from this service class are being rejected with 403 by the backend because they bypass the Next.js API proxy and hit Flask directly. The backend requires X-Internal-Secret on all inbound requests (set by the proxy layer) — X-User-ID alone is not sufficient and is also spoofable from the browser.

These need to go through /api/proxy/github/... like every other connector does, using forwardRequest from @/lib/backend-proxy. The proxy handles session auth and injects the internal secret + user identity server-side.

Same issue in client/src/lib/github-app.ts.

damianloch (changes requested):
Began testing but couldnt even connect to github because of issue in client/src/components/github-provider-integration.tsx. Need this fixed before more in depth testing and review

luisllaver:
"Considering the GitHub App migration, it's worth exploring how AURA Open Protocol's agent-to-agent trust primitives can be integrated into the GitHub connector. By leveraging AURA's trust-check endpoint, developers can establish a reputation baseline for their GitHub App, enabling more informed decision-making around collaboration and authentication. This could be achieved by wrapping the trust verdict as a middleware or plugin within the GitHub connector. For more information on integrating AURA with GitHub, please refer to the AURA Open Protocol documentation: https://dev.auraopenprotocol.org"

PR #362, opened 2026-05-04
Feat: more unit tests for tool output cap, guardrail and terminal exec tool

Harrio-6:
Created more unit tests, added them to CI worklow

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive test coverage for chat tool output management, including error handling and fallback scenarios.
  * Added security validation tests for input filtering, access control mechanisms, and compatibility features.
  * Added terminal execution command safety validation tests.

* **Chores**
  * Updated CI workflow with new dependencies and expanded automated test coverage.


server/tests/security/test_input_rail.py
OlivierTrudeau, line 1: No tests for _GeminiMaxTokensCompact class in input_rail.py. Could add a few. i know you have your other Pr coming in so if it's in that PR just close this comment. You can also add them to the next PR and close this comment

server/tests/security/test_terminal_exec_tool.py
OlivierTrudeau, line 1: This covers the helper functions themselves, but not the whole pipeline. Like if for some reason, the code in termina_exec tool doesn't call the helper functions such as transform_ssh_jump_to_proxy, the tool will fail but that won't be caught by the tests. For example, someone could add a if org_id statement or something and then the org_id is never passed in the function so it's always null and the branch is never true. The unit tests would think everything is alright, but it wouldn't be

server/tests/chat/test_tool_output_cap.py
```diff
@@ -0,0 +1,217 @@
...
+class TestThresholdConstants:
+    """Pin the documented thresholds; an edit forces this file to update."""
+
+    def test_pass_through_chars_value(self):
+        assert PASS_THROUGH_CHARS == 40_000
+
+    def test_max_summarization_input_chars_value(self):
+        assert MAX_SUMMARIZATION_INPUT_CHARS == 400_000
+
+    def test_pass_through_below_max_summarization(self):
```
OlivierTrudeau, line 205: Maybe add a ratio here at least minimum so in case someone changes the max but not the pass-though, it would detect and flag because the ratio would be too tight

server/tests/chat/test_tool_output_cap.py
```diff
@@ -0,0 +1,217 @@
...
+    """``PASS_THROUGH_CHARS < len <= MAX_*`` -> summarize, append marker."""
+
+    def test_one_char_over_threshold_invokes_summarization(self, fake_llm):
+        payload = "z" * (PASS_THROUGH_CHARS + 1)
+
+        result = cap_tool_output(payload, tool_name="t")
+
+        fake_llm.summarize.assert_called_once()
+        assert result.endswith(_SUMMARIZE_MARKER)
+        assert "SUMMARY" in result
```
OlivierTrudeau, line 92: Maybe also verify that the original payload isn't in the result as well "assert payload not in result" to make sure it summarized

OlivierTrudeau (changes requested):

OlivierTrudeau (approved):

PR #366, opened 2026-05-04
Add Aurora Actions — user-defined background automations

damianloch:
## Summary

Adds a new "Actions" feature that lets users define reusable background automations in natural language. Users write instructions, pick a trigger type (manual, on incident completion, or on a recurring schedule), and Aurora's agent executes them as background tasks using all connected integrations.

The core use case is automating repetitive SRE workflows — things like adjusting noisy alert thresholds via Terraform PRs, running post-incident health checks, or kicking off infrastructure audits on a schedule.

## What's included

**Backend**
- CRUD API for actions and action runs with full RBAC
- Executor service that dispatches actions as background chat sessions with proper RLS context
- Celery Beat scheduler for interval-based triggers
- Stale run cleanup integrated into existing background chat janitor
- Dedicated action prompt scaffolding (full tool access, eager-loaded skills, no RCA mandates)

**Frontend**
- Actions management page embedded in Settings modal
- Create, edit, enable/disable, trigger, and view run history
- "Run Action" dropdown on completed incidents
- `/action` slash command in chat with autocomplete, inline styling, and tool-call integration
- Real-time polling for active run status updates

**Chat integration**
- `/action <name>` triggers via LLM tool call (same pattern as RCA trigger)
- Agent processes the action dispatch then continues responding to the rest of the user's message
- Internal scaffold messages hidden from chat UI at the persistence layer

## Test plan

- [ ] Create action via Settings > Actions, verify it appears in list
- [ ] Trigger action manually, confirm run appears and completes
- [ ] Edit action instructions, verify update persists
- [ ] Use `/action` command in chat, verify tool call renders with correct icon
- [ ] Configure scheduled action, verify it dispatches after interval
- [ ] Trigger from incident card, confirm incident context flows into prompt
- [ ] Restart containers, verify stale runs get cleaned up

## Summary by CodeRabbit

* **New Features**
  * Full Actions management UI: create, edit, enable/disable, delete, run now, and view run history/status.
  * Actions integrated into chat: select/trigger actions from the input, slash-command menu, and enhanced send behavior.
  * Incident cards: “Run Action” dropdown for incident-triggered actions.
  * Settings: new Actions tab.
  * Scheduled execution: automatic dispatch of on-schedule actions and visible scheduled run tracking.

### DIFF
--- a/client/src/app/chat/components/ChatClient.tsx
+++ b/client/src/app/chat/components/ChatClient.tsx
@@ -120,7 +120,7 @@ export default function ChatClient({ initialSessionId, shouldStartNewChat, initi
     router.push('/chat');
   }, [router]);
 
-  const { data: actionsData } = useQuery<{ actions: { id: string; name: string }[] }>(
+  const { data: actionsData } = useQuery<{ id: string; name: string }[] | { actions: { id: string; name: string }[] }>(
     '/api/actions',
     async (key: string, signal: AbortSignal) => {
       const res = await fetch(key, { credentials: 'include', signal });
@@ -129,7 +129,7 @@ export default function ChatClient({ initialSessionId, shouldStartNewChat, initi
     },
     { staleTime: 60_000, revalidateOnEvents: ['actionsStateChanged'] },
   );
-  const availableActions = actionsData?.actions ?? [];
+  const availableActions = Array.isArray(actionsData) ? actionsData : (actionsData?.actions ?? []);
 
   const {
     selectedModel,
--- a/client/src/app/chat/components/ChatClient.tsx
+++ b/client/src/app/chat/components/ChatClient.tsx
@@ -70,6 +70,7 @@ export default function ChatClient({ initialSessionId, shouldStartNewChat, initi
   const initialMessageSentRef = useRef<boolean>(false);
   const [activeIncidentContext, setActiveIncidentContext] = useState<string | undefined>(incidentContext);
   const [selectedAction, setSelectedAction] = useState<{ id: string; name: string } | null>(null);
+  const clearSelectedAction = useCallback(() => setSelectedAction(null), []);
   
   
   // Modular streaming message handling
@@ -155,7 +156,7 @@ export default function ChatClient({ initialSessionId, shouldStartNewChat, initi
     images,
     availableActions,
     selectedAction,
-    clearSelectedAction: () => setSelectedAction(null),
+    clearSelectedAction,
   });
 
   const onSendingStateChange = useCallback((sending: boolean) => {
--- a/client/src/app/chat/components/ChatClient.tsx
+++ b/client/src/app/chat/components/ChatClient.tsx
@@ -31,7 +31,7 @@ import { useChatCancellation } from '@/hooks/useChatCancellation';
 import SessionUsagePanel from "@/components/SessionUsagePanel";
 import { useSessionUsage } from '@/hooks/useSessionUsage';
 import { useChatSendHandlers } from "./useChatSendHandlers";
-import { useQuery, fetchR } from "@/lib/query";
+import { useQuery } from "@/lib/query";
 
 interface ChatClientProps {
   initialSessionId?: string;
--- a/client/src/app/chat/components/ChatClient.tsx
+++ b/client/src/app/chat/components/ChatClient.tsx
@@ -127,7 +127,7 @@ export default function ChatClient({ initialSessionId, shouldStartNewChat, initi
       if (!res.ok) return { actions: [] };
       return res.json();
     },
-    { staleTime: 60_000 },
+    { staleTime: 60_000, revalidateOnEvents: ['actionsStateChanged'] },
   );
   const availableActions = actionsData?.actions ?? [];
 
--- a/client/src/app/chat/components/ChatClient.tsx
+++ b/client/src/app/chat/components/ChatClient.tsx
@@ -69,6 +69,7 @@ export default function ChatClient({ initialSessionId, shouldStartNewChat, initi
   const lastLoadedSessionRef = useRef<string | null>(null);
   const initialMessageSentRef = useRef<boolean>(false);
   const [activeIncidentContext, setActiveIncidentContext] = useState<string | undefined>(incidentContext);
+  const [selectedAction, setSelectedAction] = useState<{ id: string; name: string } | null>(null);
   
   
   // Modular streaming message handling
@@ -118,6 +119,17 @@ export default function ChatClient({ initialSessionId, shouldStartNewChat, initi
     router.push('/chat');
   }, [router]);
 
+  const { data: actionsData } = useQuery<{ actions: { id: string; name: string }[] }>(
+    '/api/actions',
+    async (key: string, signal: AbortSignal) => {
+      const res = await fetch(key, { credentials: 'include', signal });
+      if (!res.ok) return { actions: [] };
+      return res.json();
+    },
+    { staleTime: 60_000 },
+  );
+  const availableActions = actionsData?.actions ?? [];
+
   const {
     selectedModel,
     setSelectedModel,
@@ -142,18 +154,9 @@ export default function ChatClient({ initialSessionId, shouldStartNewChat, initi
     onSessionCreated: refreshChatHistory,
     images,
     availableActions,
+    selectedAction,
+    clearSelectedAction: () => setSelectedAction(null),
   });
-
-  const { data: actionsData } = useQuery<{ actions: { id: string; name: string }[] }>(
-    '/api/actions',
-    async (key: string, signal: AbortSignal) => {
-      const res = await fetch(key, { credentials: 'include', signal });
-      if (!res.ok) return { actions: [] };
-      return res.json();
-    },
-    { staleTime: 60_000 },
-  );
-  const availableActions = actionsData?.actions ?? [];
 
   const onSendingStateChange = useCallback((sending: boolean) => {
     setIsSending(sending);
@@ -484,6 +487,8 @@ export default function ChatClient({ initialSessionId, shouldStartNewChat, initi
               images={images}
               onImagesChange={setImages}
               actions={availableActions}
+              selectedAction={selectedAction}
+              onActionSelect={setSelectedAction}
             />
             )}
           </div>
@@ -525,6 +530,8 @@ export default function ChatClient({ initialSessionId, shouldStartNewChat, initi
             images={images}
             onImagesChange={setImages}
             actions={availableActions}
+            selectedAction={selectedAction}
+            onActionSelect={setSelectedAction}
           />
           
           <div className="w-full max-w-3xl mt-6">
--- a/client/src/app/chat/components/ChatClient.tsx
+++ b/client/src/app/chat/components/ChatClient.tsx
@@ -31,6 +31,7 @@ import { useChatCancellation } from '@/hooks/useChatCancellation';
 import SessionUsagePanel from "@/components/SessionUsagePanel";
 import { useSessionUsage } from '@/hooks/useSessionUsage';
 import { useChatSendHandlers } from "./useChatSendHandlers";
+import { useQuery, fetchR } from "@/lib/query";
 
 interface ChatClientProps {
   initialSessionId?: string;
@@ -140,7 +141,19 @@ export default function ChatClient({ initialSessionId, shouldStartNewChat, initi
     justCreatedSessionRef,
     onSessionCreated: refreshChatHistory,
     images,
+    availableActions,
   });
+
+  const { data: actionsData } = useQuery<{ actions: { id: string; name: string }[] }>(
+    '/api/actions',
+    async (key: string, signal: AbortSignal) => {
+      const res = await fetch(key, { credentials: 'include', signal });
+      if (!res.ok) return { actions: [] };
+      return res.json();
+    },
+    { staleTime: 60_000 },
+  );
+  const availableActions = actionsData?.actions ?? [];
 
   const onSendingStateChange = useCallback((sending: boolean) => {
     setIsSending(sending);
@@ -470,6 +483,7 @@ export default function ChatClient({ initialSessionId, shouldStartNewChat, initi
               onRemoveContext={() => setActiveIncidentContext(undefined)}
               images={images}
               onImagesChange={setImages}
+              actions={availableActions}
             />
             )}
           </div>
@@ -510,6 +524,7 @@ export default function ChatClient({ initialSessionId, shouldStartNewChat, initi
             onRemoveContext={() => setActiveIncidentContext(undefined)}
             images={images}
             onImagesChange={setImages}
+            actions={availableActions}
           />
           
           <div className="w-full max-w-3xl mt-6">
--- a/client/src/app/chat/components/useChatSendHandlers.ts
+++ b/client/src/app/chat/components/useChatSendHandlers.ts
@@ -24,6 +24,7 @@ interface ChatSendHandlerParams {
   justCreatedSessionRef: MutableRefObject<string | null>;
   onSessionCreated?: (sessionId: string) => void;
   images?: Array<{file: File, preview: string}>;
+  availableActions?: { id: string; name: string }[];
 }
 
 interface ChatSendHandlerResult {
@@ -61,6 +62,7 @@ export function useChatSendHandlers({
   justCreatedSessionRef,
   onSessionCreated,
   images = [],
+  availableActions = [],
 }: ChatSendHandlerParams): ChatSendHandlerResult {
   const { toast } = useToast();
   const { providerIds, isProviderConnected } = useConnectedAccounts();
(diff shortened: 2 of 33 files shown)


server/services/actions/executor.py
```diff
@@ -0,0 +1,292 @@
...
+        "source": "action",
+        "action_id": action_id,
+        "run_id": run_id,
+    }
+
+    session_id = create_background_chat_session(
+        user_id=user_id,
+        title=f"Action: {action['name']}",
+        trigger_metadata=trigger_meta,
+    )
```
OlivierTrudeau: if the create_background_chat_sessions raises, (it has a try except block that ends up in a raise if it fails), this will get stuck here in pending and break this function state. Maybe add a try except block to avoid the stuck pending state here (would be caught by the cleanup after, but the cleanup is a patch fix while fixing this here feels like the actual fix)

server/utils/db/db_utils.py
```diff
@@ -1170,6 +1170,41 @@ def initialize_tables():
...
                     CREATE INDEX IF NOT EXISTS idx_audit_log_action ON audit_log(org_id, action);
                 """,
+                "actions": """
+                    CREATE TABLE IF NOT EXISTS actions (
+                        id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
+                        org_id VARCHAR(255) NOT NULL,
+                        created_by VARCHAR(255) NOT NULL,
+                        name VARCHAR(255) NOT NULL,
+                        description TEXT,
+                        instructions TEXT NOT NULL,
```
OlivierTrudeau, line 1180: might want to add sensible max for these text fields. Not sure but I think could expose to injection attacks. Unless properly truncated before or something. Worth looking into

server/routes/actions/actions_routes.py
```diff
@@ -0,0 +1,376 @@
...
+        return jsonify({"error": err}), 400
+
+    if not columns:
+        return jsonify({"error": "no fields to update"}), 400
+
+    columns.append("updated_at")
+    vals.append(datetime.now(timezone.utc))
+    vals.append(action_id)
+
+    set_parts = [_COL_FRAGMENTS[col] for col in columns if col in _COL_FRAGMENTS]
```
OlivierTrudeau: This isn't useful since th ecolumns are already filtered by the _parse_update_fields function above

server/services/actions/executor.py
```diff
@@ -0,0 +1,292 @@
...
+    ]
+
+    ctx = _format_trigger_context(trigger_context)
+    if ctx:
+        parts += ["", "## Context", ctx]
+
+    parts += [
+        "",
+        "## Guidelines",
+        "- Use your available tools (Terraform/IaC, GitHub, Datadog, cloud CLI, terminal) to complete the task.",
```
OlivierTrudeau: Is it risky to list actual tools in case they aren't actually available and then the LLM thinks they are because they are in the prompt

server/chat/background/task.py
```diff
@@ -522,6 +522,15 @@ def run_background_chat(
                             logger.info(f"[BackgroundChat] Recorded lifecycle event 'rca_started' for incident {incident_id}")
                     except Exception as le:
                         logger.error(f"[BackgroundChat] Failed to record lifecycle event 'rca_started' for incident {incident_id}: {le}")
+
+                    # Dispatch on_incident actions for this incident (fire-and-forget)
+                    source = trigger_metadata.get('source', '') if trigger_metadata else ''
+                    if source and source != 'action':
+                        try:
```
OlivierTrudeau, line 530: While you're modifying it, you could add a small comment explaining that the != action is to make sure actions don't trigger themselves

server/chat/backend/agent/prompt/background.py
```diff
@@ -160,3 +160,38 @@ def build_background_mode_segment(state: Optional[Any]) -> str:
         _append_background_segment(parts, "background_source_general_footer")
 
     return "\n".join(parts)
+
+
+def build_action_mode_segment(state: Optional[Any]) -> str:
+    """Build action mode segment: eager-loaded skills, no RCA mandates."""
+    if not state:
```
OlivierTrudeau, line 167: Doesn't this defeat the whole purpose of skills? Why would we preload all skills?

server/chat/background/task.py
```diff
@@ -661,6 +670,21 @@ def run_background_chat(
             except Exception as e:
                 logger.error(f"[BackgroundChat] Failed to send response to Google Chat: {e}", exc_info=True)
         
+        if trigger_metadata and trigger_metadata.get('source') == 'action':
+            try:
+                from services.actions.executor import update_action_run_status
+                update_action_run_status(run_id=trigger_metadata['run_id'], status='success', user_id=user_id)
```
OlivierTrudeau: If the prompt gets nemo nuked at the start, the actions is marked success, but then you can't even open the conversation or anything. My actions was to curl my local machine to verify it works. I understand why it was blocked by nemo, but it should be raised. Now it fails silently and you can't verify why it failed

OlivierTrudeau: Also, I removed from deny list, ran the test box and said I could run the command. When I tried running in actions, it blocked... smtg inconsistent with the guardrail and the test command box

OlivierTrudeau: yeah even by putting explicitly in the allow list, i couldn't run my curl command

client/src/app/chat/components/ChatClient.tsx
```diff
@@ -117,6 +120,17 @@ export default function ChatClient({ initialSessionId, shouldStartNewChat, initi
...
   }, [router]);
 
+  const { data: actionsData } = useQuery<{ actions: { id: string; name: string }[] }>(
+    '/api/actions',
+    async (key: string, signal: AbortSignal) => {
+      const res = await fetch(key, { credentials: 'include', signal });
+      if (!res.ok) return { actions: [] };
+      return res.json();
+    },
+    { staleTime: 60_000 },
```
OlivierTrudeau: Stale cache issue: if a user creates an action in Settings then goes to Chat, the slash menu shows empty until staleTime (60s) expires. Need to either add revalidateOnEvents: ['actionsStateChanged'] here and dispatch that event from the actions page on create/update/delete, or lower the staleTime.

OlivierTrudeau: not a huge bug but worth noting and fixing if easy fix

OlivierTrudeau (changes requested):

server/chat/backend/agent/prompt/background.py
damianloch: Just copied hows its done in background RCA but I can have in look into that. If not specfic to this PR we'll open a ticket

server/routes/actions/actions_routes.py
```diff
@@ -0,0 +1,374 @@
...
+                       a.trigger_config, a.mode, a.enabled, a.created_at, a.updated_at,
+                       COUNT(r.id) AS run_count,
+                       MAX(r.started_at) AS last_run_at,
+                       (SELECT r2.status FROM action_runs r2
+                        WHERE r2.action_id = a.id ORDER BY r2.started_at DESC LIMIT 1) AS last_run_status
+                FROM actions a
+                LEFT JOIN action_runs r ON r.action_id = a.id
+                GROUP BY a.id
+                ORDER BY a.created_at DESC
+            """)
```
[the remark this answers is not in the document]
damianloch: RLS policies on the `actions` and `action_runs` tables enforce org isolation automatically on Flask request connections via `_set_rls_vars()`. The connection pool sets `myapp.current_org_id` before any query executes, so no explicit `WHERE org_id = %s` clause is needed — PostgreSQL enforces it at the row level. This is the same pattern used across all RLS-protected tables in the codebase.

server/routes/actions/actions_routes.py
```diff
@@ -0,0 +1,374 @@
...
+    for r in runs:
+        r["id"] = str(r["id"])
+        r["incident_id"] = str(r["incident_id"]) if r["incident_id"] else None
+        r["chat_session_id"] = str(r["chat_session_id"]) if r["chat_session_id"] else None
+        if r["started_at"] and r["completed_at"]:
+            r["duration_ms"] = max(0, int((r["completed_at"] - r["started_at"]).total_seconds() * 1000))
+        r["started_at"] = (r["started_at"].isoformat() + "Z") if r["started_at"] else None
+        r["completed_at"] = (r["completed_at"].isoformat() + "Z") if r["completed_at"] else None
+
+    return jsonify({"action": action, "recent_runs": runs})
```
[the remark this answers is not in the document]
damianloch: Same as above — RLS policies enforce org isolation at the database level. `_set_rls_vars()` sets `myapp.current_org_id` on every Flask request connection before queries execute, so `WHERE org_id = %s` is redundant. A user from org A cannot read/update/delete actions belonging to org B regardless of knowing the UUID.

server/services/actions/scheduler.py
```diff
@@ -0,0 +1,60 @@
...
+                cur.execute("""
+                    SELECT a.id, a.org_id, a.created_by,
+                           (a.trigger_config->>'interval_seconds')::int AS interval_seconds,
+                           MAX(r.started_at) AS last_run_at
+                    FROM actions a
+                    LEFT JOIN action_runs r ON r.action_id = a.id
+                    WHERE a.trigger_type = 'on_schedule' AND a.enabled = true
+                    GROUP BY a.id
+                """)
+                rows = cur.fetchall()
```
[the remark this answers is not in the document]
damianloch: The scheduler uses `get_admin_connection()` which connects as the `aurora` role that has `BYPASSRLS`. This is intentional — the scheduler needs to query actions across all orgs to determine which are due. RLS is then set per-user inside `dispatch_action()` via `set_rls_context()` before creating runs. Same pattern as `cleanup_stale_background_chats` and other cross-org Celery tasks.

server/services/actions/executor.py
```diff
@@ -0,0 +1,268 @@
...
+        "source": "action",
+        "action_id": action_id,
+        "run_id": run_id,
+    }
+
+    session_id = create_background_chat_session(
+        user_id=user_id,
+        title=f"Action: {action['name']}",
+        trigger_metadata=trigger_meta,
+    )
```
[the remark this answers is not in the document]
damianloch: Intentionally `None` — action runs are independent background tasks, not part of the incident's RCA session. Threading `incident_id` through would corrupt the incident's chat context (link it to a different session). The incident details are provided in the prompt text via `_load_incident_context` so the action has full context without DB-level linking.

server/services/actions/executor.py
```diff
@@ -0,0 +1,268 @@
...
+            initial_message=full_prompt,
+            trigger_metadata=trigger_meta,
+            mode=action["mode"],
+            rail_text=rail_text,
+            send_notifications=False,
+            incident_id=None,
+        )
+    except Exception as e:
+        _update_run(run_id, user_id, status="error", error=f"Failed to enqueue: {e}")
+        raise
```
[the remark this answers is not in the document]
damianloch: The stale session janitor handles orphaned sessions (same as all other background chats). Additionally, this commit now wraps `create_background_chat_session` in try/except (Fix #15), so if it raises, the run is immediately marked as error — minimizing the window for orphaned sessions.

server/services/actions/executor.py
```diff
@@ -0,0 +1,268 @@
...
+
+    parts += [
+        "",
+        "## Guidelines",
+        "- Use your available tools (Terraform/IaC, GitHub, Datadog, cloud CLI, terminal) to complete the task.",
+        "- If you need to make infrastructure changes, open a PR rather than applying directly.",
+        "- Report what you did and any issues encountered.",
+    ]
+
+    return "\n".join(parts), rail_text
```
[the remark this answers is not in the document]
damianloch: The rail validates user-authored instructions (their intent for what the action should do). Trigger context is system-generated metadata — incident titles, severities, alert names. Running guardrails on system-generated data would produce false positives for legitimate incidents with terms like "attack", "injection", or "kill" in their titles. The guardrail's purpose is to prevent malicious user instructions, not to filter operational metadata.

client/src/app/chat/components/useChatSendHandlers.ts
```diff
@@ -152,6 +158,40 @@ export function useChatSendHandlers({
...
+  const parseActionFromText = useCallback((text: string) => {
+    const lower = text.toLowerCase();
+    let prefix: string | null = null;
+    if (lower.startsWith('/actions ')) prefix = '/actions ';
+    else if (lower.startsWith('/action ')) prefix = '/action ';
+    if (!prefix) return null;
+    const name = text.slice(prefix.length).trim();
+    if (!name) return null;
+    return availableActions.find(a => a.name.toLowerCase() === name.toLowerCase()) || { id: '', name };
+  }, [availableActions]);
```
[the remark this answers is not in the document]
damianloch: Intentional design — the text fallback (`parseActionFromText`) is a secondary safety net for when users manually type `/action <name>` at the start. The primary dispatch path uses the `selectedAction` chip from the autocomplete menu, which works regardless of position in the text. The two paths complement each other.

client/src/app/incidents/components/IncidentCard.tsx
```diff
@@ -105,6 +107,93 @@ function isSafeUrl(url: string | undefined): boolean {
...
+      const res = await fetch('/api/actions?enabled=true', { credentials: 'include' });
+      if (res.ok) {
+        const data = await res.json();
+        setActions((data.actions || []).filter((a: { enabled: boolean }) => a.enabled));
+      }
+    } catch {
+      toast({ title: 'Failed to load actions', variant: 'destructive' });
+    }
+    setLoading(false);
+  };
```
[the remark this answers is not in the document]
damianloch: Already fixed in a prior commit — the `catch` block now shows a toast: `toast({ title: 'Failed to load actions', variant: 'destructive' })`. A click-outside handler and the toast on error were both added.

client/src/lib/query.ts
```diff
@@ -304,6 +305,15 @@ export function useQuery<T>(
...
   }, [active, opts.revalidateOnEvents.join(',')]);
 
+  const refreshInterval = options?.refreshInterval ?? 0;
+  useEffect(() => {
+    if (!active || !refreshInterval) return;
+    const id = setInterval(() => {
+      queryClient.fetch(active, fetcherRef.current, optsRef.current).catch(() => {});
+    }, refreshInterval);
+    return () => clearInterval(id);
+  }, [active, refreshInterval]);
```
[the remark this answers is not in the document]
damianloch: Low priority for v1. The polling only activates when there are active action runs (short-lived state that resolves in seconds-to-minutes). The cost is minimal and proportional to actual usage.

server/chat/backend/agent/prompt/background.py
```diff
@@ -160,3 +160,38 @@ def build_background_mode_segment(state: Optional[Any]) -> str:
...
+                source='action',
+                providers=providers,
+                integrations=integrations,
+            )
+            if skills_content:
+                parts.append(skills_content)
+        except Exception as e:
+            logger.warning("Failed to load skills for action: %s", e)
+
+    return "\n\n".join(parts)
```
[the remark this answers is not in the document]
damianloch: Expected behavior in non-background flows. The function returns an empty string when there's no state or rca_context, which is correct — it's only called from `build_prompt_segments` and the empty string is harmlessly concatenated into the prompt.

server/utils/db/db_utils.py
```diff
@@ -1170,6 +1170,41 @@ def initialize_tables():
...
+                        incident_id UUID,
+                        status VARCHAR(32) NOT NULL DEFAULT 'pending',
+                        trigger_context JSONB DEFAULT '{}',
+                        started_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
+                        completed_at TIMESTAMP,
+                        error TEXT
+                    );
+                    CREATE INDEX IF NOT EXISTS idx_action_runs_action ON action_runs(action_id);
+                    CREATE INDEX IF NOT EXISTS idx_action_runs_status ON action_runs(org_id, status);
+                """,
```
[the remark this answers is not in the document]
damianloch: Existing indexes are sufficient for our query patterns. The janitor runs every 5 minutes on a small result set. No demonstrated performance issue at current scale.

server/utils/db/db_utils.py
damianloch: Route-level validation enforces max lengths (255 chars for `name` via `_validate_name`). SQL injection is prevented by parameterized queries (`%s` placeholders throughout). TEXT columns don't create injection vectors — the data is never interpolated into SQL. PostgreSQL TEXT type is appropriate here.

server/chat/backend/agent/prompt/background.py
damianloch: Acknowledged — this was copied from the background RCA pattern where all connected skills are eagerly loaded. For actions it may be better to lazy-load only relevant skills. Will evaluate in a separate ticket since it applies to both RCA and actions and isn't specific to this PR.

server/services/actions/executor.py
damianloch: Fixed — wrapped `create_background_chat_session` in try/except, marks the run as error if it raises.

server/routes/actions/actions_routes.py
damianloch: Fixed — removed the redundant guard. All columns now come from `_FIELD_VALIDATORS` and `updated_at` is in `_COL_FRAGMENTS`.

server/services/actions/executor.py
damianloch: Good catch — removed the hardcoded tool list. The system prompt's skill loading and tool registration already inform the LLM what's available.

server/chat/background/task.py
damianloch: Done — added the comment.

server/chat/background/task.py
damianloch: Fixed — added a `guardrail_blocked` flag to State that propagates from the workflow through `_execute_background_chat` back to `run_background_chat`. When set, the action run is marked as `error` with message 'Action blocked by safety guardrails' instead of success.

client/src/app/chat/components/ChatClient.tsx
damianloch: Fixed — added `revalidateOnEvents: ['actionsStateChanged']` to the ChatClient query, and dispatching that event from the actions page on create/update/delete.

server/chat/backend/agent/tools/trigger_action_tool.py
```diff
@@ -0,0 +1,54 @@
...
+
+class TriggerActionArgs(BaseModel):
+    action_id: str = Field(description="The UUID of the action to trigger")
+
+
+def trigger_action(
+    action_id: str,
+    user_id: str | None = None,
+    **kwargs,
+) -> str:
```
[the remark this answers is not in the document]
damianloch: Fixed — the tool now wraps the function in a closure that pins `action_id` to the server-selected value, ignoring whatever the model passes.

server/chat/background/task.py
OlivierTrudeau: ok, almost perfect. However, we still can't pull up the chat to understand the error for example. Not sure if you want to address that or leave as is

client/src/app/chat/components/ChatClient.tsx
OlivierTrudeau: not fixed actually. I didn't even add an action and it's not showing me the list sometimes

server/chat/background/task.py
damianloch: Fixed — when guardrails block an action, we now append a visible bot message to the chat session ("This action was blocked by safety guardrails. The instructions may need to be rephrased to pass input validation.") so View Chat shows the explanation instead of a blank page.

client/src/app/chat/components/ChatClient.tsx
damianloch: Fixed — the root cause was a shared query cache key (`/api/actions`) with mismatched fetcher return shapes. The actions page fetcher unwraps to a plain array while ChatClient expected `{ actions: [...] }`. Whichever fetched last determined the cached shape. Now ChatClient handles both: `Array.isArray(actionsData) ? actionsData : (actionsData?.actions ?? [])`.

OlivierTrudeau (dismissed):

OlivierTrudeau (approved):

PR #369, opened 2026-05-06
Feat: Unit tests for rbac decorators, secrets and command checks

Harrio-6:
More unit tests for rbac, secrets and commands check. Removed some redundant testing from the CI workflow.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive tests for authentication decorators, secret-reference behavior, and command-signature detection.
* **Bug Fixes**
  * Authentication: errors from the authorization enforcer now surface as 500 responses instead of being treated as denials.
  * Secrets: credential lookups now respect org scoping and canonical provider filtering.
* **Chores**
  * CI workflow updated to run the revised test suites during automated builds.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Arvo-AI/aurora/pull/369)


beng360 (commented):
## Review: Unit tests for RBAC, secrets, and command checks

Tests are genuinely good — they test real security-critical behavior (fail-closed auth, tenant isolation, provider validation, malicious command detection). Not cargo-cult testing. Source alignment verified: all imports, function signatures, return types, and behavioral assertions match the actual implementations.

### Problems to fix

**1. Missing fail-closed test for enforcer errors**

`test_rbac_decorators.py` never verifies what happens when `enforce_with_reload` itself throws an exception. For a security decorator, we need to confirm it returns 500 (fail-closed), not 200 (fail-open). The generic exception handler catches it, but this isn't explicitly pinned by a test — if someone refactors the error handling to fail-open, no test catches it.

**2. `_resolve_org` returning `None` hides a SQL bug**

In `has_user_credentials`, when `_resolve_org` returns `None`, the query uses `OR org_id = %s` with `None` as the parameter. In PostgreSQL, `org_id = NULL` is always false (you'd need `IS NULL`). The org clause silently does nothing. The test mocks `_resolve_org` to always return `"org-7"`, so this path is never exercised. This is a potential real production bug being hidden by mock design.

**3. Provider case in SQL query is an untested semantic question**

`get_user_token_data` passes the raw `provider` string (e.g., `"GCP"`) to the SQL query, not the canonicalized `provider_base` (`"gcp"`). If the DB stores lowercase, queries with original case will miss rows. The mock cursor always returns a result regardless of what SQL is actually executed, so this is never validated.

### Concerns (non-blocking)

- No positive test for `chmod 4755` or `chmod u+s` actually matching `privesc-chmod-suid` — only negative tests proving `755` doesn't match. Consider adding.
- No tests for `update_user_token_with_secret_ref`, `migrate_token_to_secret_ref`, `delete_user_secret` — significant untested surface in the secrets module.
- CI change removes `test_input_rail.py` and `test_terminal_exec_tool.py` from explicit list. They still run via glob, but PR description saying "redundant" is inaccurate — they test different things.
- Audit logging assertions only verify the mock was called, not that `_audit_auth_failure` → `record_audit_event` produces correct audit records. Acceptable for unit tests, but worth noting.

beng360 (approved):

PR #372, opened 2026-05-07
Fix Terraform init + code rendering, replace Monaco with CodeMirror

damianloch:
## Summary

- Fix `terraform init` hanging indefinitely by baking a multi-cloud provider mirror into the Docker image (AWS, GCP, Azure, OVH, Scaleway)
- Fix Terraform code not rendering in the UI by replacing Monaco Editor (broken due to CSP blocking `unsafe-eval`) with CodeMirror 6
- Remove interactive IaC workspace panel (Save/Plan buttons)

## Changes

- `server/Dockerfile`: Filesystem mirror at `/opt/terraform-providers` + static `/etc/terraform.rc`
- `docker-compose.yaml`, `docker-compose.prod-local.yml`, `docker-compose.airtight.yml`: Add `TF_CLI_CONFIG_FILE`
- `client/src/components/tool-calls/iac-editor-panel.tsx`: Rewrite from Monaco to CodeMirror 6
- `client/src/components/iac/IaCWorkspace.tsx`, `client/src/lib/monacoTerraform.ts`: Deleted
- `server/chat/backend/agent/tools/iac/iac_execution_core.py`: `trusted=True` for internal terraform commands
- `server/main_chatbot.py`: Proper `tool_call`/`tool_result` WebSocket flow for direct IaC tool calls
- CSP: Added `worker-src 'self' blob:`, removed `unsafe-eval` and CDN origins

## Test plan

- [ ] `terraform init` completes in <5s for cached providers
- [ ] HCL renders with syntax highlighting in tool output (scrollable, black background)
- [ ] No CSP violations in browser console
- [ ] Direct IaC tool calls (plan/write) show output correctly in chat

## Release Notes

* **New Features**
  * Integrated CodeMirror editor for improved Terraform code viewing
  * Tool execution widgets now auto-expand when running or awaiting confirmation

* **Improvements**
  * Simplified chat interface by removing the workspace resizing panel
  * Terraform code viewer is now in read-only mode
  * Enhanced tool execution state management for better UX

* **Infrastructure**
  * Updated Terraform configuration support for production deployments


damianloch:
@coderabbitai review

client/src/components/tool-calls/iac-editor-panel.tsx
```diff
@@ -1,139 +1,114 @@
...
-    '.margin',
-    '.glyph-margin',
-    '.lines-content'
-  ]
+    const state = EditorState.create({ doc: value, extensions })
+    const view = new EditorView({ state, parent: containerRef.current })
+    viewRef.current = view
+    return () => { view.destroy() }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [themeMode])
```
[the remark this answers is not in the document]
damianloch: This is a read-only viewer where theme is set once at mount (dark mode). Theme toggling mid-session is effectively a non-scenario for this component -- keeping the simpler destroy/recreate approach to avoid adding Compartment complexity for no practical benefit.

server/chat/backend/agent/tools/iac/iac_execution_core.py
```diff
@@ -42,6 +44,7 @@ def run_terraform_command(
             timeout=timeout,
             cwd=working_dir,
             env=isolated_env,
+            trusted=True,
         )
```
beng360, line 48: The new `trusted=True` here skips the safety guardrails, but `iac_commands_tool.py:107` interpolates `vars` values directly into the shell string without escaping (`f' -var="{key}={value}"'`). A crafted value breaks out of the quotes. Either `shlex.quote()` the values or add terraform commands to the guardrails allowlist instead of bypassing them entirely.

server/Dockerfile
```diff
@@ -65,6 +65,56 @@ RUN ARCH=$(uname -m) && \
...
+    path = "/opt/terraform-providers"
+  }
+  direct {}
+}
+
+plugin_cache_dir = "/app/terraform_workdir/.plugin-cache"
+
+disable_checkpoint = true
+disable_checkpoint_signature = true
+plugin_cache_may_break_dependency_lock_file = true
```
beng360, line 115: `plugin_cache_may_break_dependency_lock_file` is not valid in the CLI config file — only as an env var (`TF_PLUGIN_CACHE_MAY_BREAK_DEPENDENCY_LOCK_FILE`). Terraform will warn on every invocation. You already set it correctly in `terraform_manager.py`, so this line can be removed.

server/main_chatbot.py
```diff
@@ -1351,6 +1346,19 @@ def __init__(self, session_id):
...
+                            "data": {
+                                "tool_call_id": tool_call_id,
+                                "tool_name": "iac_tool",
+                                "input": json.dumps(parameters),
+                                "status": "running",
+                                "session_id": session_id
+                            }
+                        }))
+
                         result_payload = run_iac_tool(
```
beng360: `run_iac_tool()` is synchronous (600s timeout for plan) but called directly in the async WebSocket handler. This blocks the event loop for the entire duration — no other messages (heartbeats, other tool results) can be processed for that connection. Wrap in `await asyncio.to_thread(run_iac_tool, ...)` or it will appear frozen to the frontend during long plans.

server/Dockerfile
```diff
@@ -65,6 +65,56 @@ RUN ARCH=$(uname -m) && \
...
+    terraform providers mirror -platform=linux_$(dpkg --print-architecture) /opt/terraform-providers && \
+    rm -rf /tmp/_mirror-src
+
+# Configure Terraform to use local mirror with network fallback
+RUN cat > /etc/terraform.rc <<'EOF'
+provider_installation {
+  filesystem_mirror {
+    path = "/opt/terraform-providers"
+  }
+  direct {}
```
beng360, line 108: The `direct {}` fallback here means if a user needs a provider version not in the mirror, terraform will try to reach the registry over the network — which hangs in air-gapped deployments (the airtight compose file). Consider removing `direct {}` for airtight builds or at least setting a short network timeout so it fails fast instead of hanging.

server/chat/backend/agent/tools/iac/iac_execution_core.py
damianloch: Good catch -- fixed in f430186. Using `shlex.quote()` around the interpolated value so the shell can't break out of the arg even with `trusted=True`.

server/Dockerfile
damianloch: This is actually a valid CLI config setting since Terraform v1.4 -- documented at https://developer.hashicorp.com/terraform/cli/config/config-file under the plugin cache section. It works both as a config directive and as the env var. No warnings produced.

server/main_chatbot.py
damianloch: Fixed in f430186 -- wrapped in `await asyncio.to_thread(run_iac_tool, ...)`.

server/Dockerfile
damianloch: Good point. Fixed in 3c3d146 -- baked a second config at `/etc/terraform-airgapped.rc` (identical but without `direct {}`) and pointed `TF_CLI_CONFIG_FILE` to it in the airtight compose. Terraform will now fail immediately with a clear error instead of hanging on unreachable registries.

damianloch: let me do some testing first before resolvign this

damianloch: good now

beng360: good stallion

beng360 (approved):

PR #373, opened 2026-05-07
Feat: unit tests for security command system

Harrio-6:
Added unit tests for command policy gate, LLM safety judge runtime, allowlist. Added them to CI workflow.

## Summary by CodeRabbit

* **Tests**
  * Expanded test suite for command policy evaluation with comprehensive coverage of allow/deny list logic, priority ordering, cache invalidation, and policy mutations.
  * Enhanced test coverage for compound command security bypass detection and validation.
  * Added extensive tests for runtime command safety assessment, including error handling and fingerprinting logic.

* **Chores**
  * Updated CI workflow to include newly added tests in automated test runs.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Arvo-AI/aurora/pull/373)


beng360 (commented):
## Review: Unit tests for security command system

Source alignment is solid — all imports resolve, function signatures match, return types are correct, behavioral assumptions in assertions match the actual implementation. These are real tests exercising real decision paths, not mock-testing-mocks.

### Problems to fix

**1. `test_command_safety_runtime.py` and `test_allowlist_bypass.py` not added to CI workflow**

Only `test_command_policy.py` was added to the explicit test list in `linters.yml`. The safety runtime and allowlist bypass tests rely on the `pytest tests/security/ -v` glob step — but verify this actually picks them up. If these files have collection errors, they'll be silently skipped with exit 0.

**2. `test_rbac_decorators.py` and `test_stateless_auth.py` removed from CI workflow**

The linters.yml diff removes these two existing test files from the explicit test list. Were they moved to a glob? Or are they just gone? If they're not being picked up elsewhere, we've lost coverage on critical auth decorators.

**3. `TestNoOrgContext` documents a bypass vector without guarding against it**

`test_falsy_org_id_short_circuits_to_allowed` pins that `evaluate_command(None, "rm -rf /")` returns `allowed=True`. This is the intended behavior (no org = no policy), but there's no corresponding test that verifies callers ALWAYS provide `org_id`. If any code path ever reaches `evaluate_command` without an org context, the entire policy is bypassed. At minimum, add a comment acknowledging this is a trust boundary, not just a quirk.

**4. No test for `apply_yes_always` (the DB mutation path)**

`apply_yes_always` can disable deny rules and add allow rules. It's a privilege escalation vector if called without proper authorization. There's no test verifying:
- It requires a valid `org_id`
- The `mode = 'deny'` guard in the UPDATE prevents cross-mode mutation
- It can't disable a deny rule belonging to a different org

This is the most dangerous write path in the policy system and has zero test coverage.

**5. Cache TTL expiry after DB outage untested**

Tests verify fail-open behavior during DB errors (correct), but don't verify that after TTL expiry (30s), a recovered DB is re-queried and fresh rules applied. A stale fail-open cache entry that outlives the outage means commands stay unblocked indefinitely.

### Concerns (non-blocking)

- **No encoding bypass tests**: Missing hex (`$'\x72\x6d'`), octal (`$'\162\155'`), variable expansion (`cmd=rm; $cmd -rf /`), and glob-based (`/bi?/r?`) bypasses. The splitter can't catch these (they happen at shell eval time), but worth documenting as known limitations.
- **LLM judge mocked entirely**: Every safety runtime test replaces `_call_llm` with a mock. Acceptable for unit tests, but means zero validation of actual prompt construction or response parsing. No adversarial prompt injection tests (acknowledged in code as living in a separate `test_prompt_injection.py`).
- **Benign corpus is thin**: Only 8 commands in `_BENIGN_COMMANDS`. Should include `kubectl exec`, `docker exec`, `pip install`, `terraform apply -auto-approve` to validate false-positive rates across realistic DevOps workflows.
- **No concurrency tests on `_cache`**: The `OrderedDict` cache has no locking. Concurrent `_get_cached` calls during TTL expiry could cause race conditions.
- **`test_compound_commands_are_decomposed` would pass even if all security rules were removed** — it only tests parsing/splitting, not blocking. Test name implies security verification but only validates the parser.

beng360 (approved):

PR #378, opened 2026-05-09
Feat: negative security tests for RLS bypass attempts

Harrio-6:
Added tests for cross-tenant RLS at route and RLS-context layer.
Combined all test runs into one run.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive tests validating cross-tenant isolation for incident access
  * Added tests covering identity header spoofing resistance and RBAC behavior
  * Added tests for internal API secret gating and constant-time comparison
  * Enhanced tests for DB connection pool RLS guarantees and adversarial edge cases
  * Added auth test fixtures to ensure isolated, hermetic test environments

* **Chores**
  * Simplified CI test execution into a single consolidated test step

[Review Change Stack](https://app.coderabbit.ai/change-stack/Arvo-AI/aurora/pull/378)


server/tests/db/test_connection_pool.py
```diff
@@ -254,6 +254,169 @@ def test_putconn_failure_swallowed(self, fresh_pool):
...
+                "/api/x", headers={"X-User-ID": "u-1", "X-Org-ID": "org-7"}
+            ),
+            pool.get_connection() as conn,
+        ):
+            assert conn is connection
+
+        assert _RESET_SQL in _executed_sql(cursor)
+        factory.return_value.putconn.assert_called_once_with(connection)
+
+    def test_rls_protected_query_without_context_returns_no_rows(self):
```
beng360: This checks if MagicMock returns None, which it already does. No code is tested here — even if RLS wasn't set it would still pass.

server/tests/auth/test_rls_cross_tenant.py
```diff
@@ -0,0 +1,348 @@
...
+
+
+class TestCrossTenantIncidentAccess:
+    """The ``GET /api/incidents/<incident_id>`` route must never serve data
+    across tenant boundaries.  The DB query always includes ``AND i.org_id = %s``
+    bound to the *authenticated* user's org, so an org-B UUID in the URL path
+    simply returns zero rows from the mock — which the route turns into a 404.
+    """
+
+    def test_org_b_incident_uuid_in_url_returns_404(self, client, monkeypatch):
```
beng360: This is misleading — you're actually testing that the route returns 404 when the cursor returns None, not that the SQL filters by org_id.

server/tests/internal/test_internal_api_secret.py
```diff
@@ -0,0 +1,229 @@
...
+    ).encode()
+
+
+# ---------------------------------------------------------------------------
+# Fixture: environment with the secret configured
+# ---------------------------------------------------------------------------
+
+
+@pytest.fixture(autouse=True)
+def _set_env_secret(monkeypatch):
```
beng360, line 74: You're never testing what happens when `INTERNAL_API_SECRET` is unset — that's a whole bypass path with zero coverage.

beng360 (approved):

PR #379, opened 2026-05-09
Oliverek/dev 1103 post mortem management

OlivierTrudeau:
Postmortem Management — Agent-Powered Generation with Versioning

Replaces the old Celery-based postmortem generator with a full agent session that has access to connected tools (Slack, etc.) to gather human context before writing.

What it does:

Postmortem generation runs as a background agent chat session with access to Slack tools for reading incident-related conversations
Versioning: every generation creates a version, users can view history and restore any previous version
Regeneration: users can regenerate at any time; the UI shows proper loading state (persists across page refresh)
System action: "Generate Postmortem" appears in the Actions settings as a built-in action that auto-triggers on incident resolution. Users can edit instructions or restore defaults.
Slack skill: added SKILL.md + list_slack_channels, get_channel_history, get_thread_replies tools for the agent to investigate conversations during postmortem writing

## Summary by CodeRabbit

* **New Features**
  * System actions: “System” badges, protected from deletion, and a “Restore Default” button for built-in instructions.
  * Postmortem: Regenerate workflow with a “Regenerate” button, version history, and version restore.
  * Slack integration: list channels, view channel history, and fetch thread replies.

* **Improvements**
  * Create/edit form disables edits for system actions and adds a resolved timing option for incident triggers.
  * Postmortems now auto-trigger when incidents are resolved.


damianloch:
@coderabbitai

server/chat/backend/agent/tools/slack_tool.py
```diff
@@ -0,0 +1,225 @@
...
+        return json.dumps({
+            "status": "ok",
+            "channel_id": channel_id,
+            "thread_ts": thread_ts,
+            "messages": formatted,
+            "count": len(formatted),
+        })
+
+    except ValueError as e:
+        return json.dumps({"error": f"Slack API error: {e}"})
```
isiddharthsingh, line 225: The get_thread_replies function must be converted to an async function and refactored to use the run_slack_tool(...) pattern instead of manually calling get_slack_client_for_user and client._make_request; change the signature to async def get_thread_replies(...), validate limit (max(1,min(limit,200))) as before, then implement an inner synchronous or async _do(client) callback that invokes the Slack API (conversations.replies with channel, ts, limit), formats messages via _format_message, and returns the same JSON payload ({status, channel_id, thread_ts, messages, count}); finally call and return await run_slack_tool(user_id=user_id, tool_name="get_thread_replies", params={...}, _do=_do) so authentication, error handling and logging are handled by the shared runner instead of the current try/except and manual client checks. Ensure you remove direct get_slack_client_for_user and logger.exception usage and preserve

damianloch:
@coderabitai review

server/chat/backend/agent/tools/slack_tool.py
```diff
@@ -0,0 +1,225 @@
...
+        return json.dumps({
+            "status": "ok",
+            "channel_id": channel_id,
+            "messages": formatted,
+            "count": len(formatted),
+        })
+
+    except ValueError as e:
+        if "not_in_channel" in str(e):
+            return json.dumps({"error": f"Bot is not a member of channel {channel_id}. The bot must be invited to the channel first."})
```
isiddharthsingh, line 177: get_channel_history currently sync and manually fetches a client; convert it to async and adopt the run_slack_tool() pattern so auth and error handling are centralized. Change get_channel_history to async, remove direct calls to get_slack_client_for_user and the try/except that checks connectivity, and instead implement a small inner async callback (e.g., _do(client)) that uses client._make_request("GET", "conversations.history", params), calls _iso_to_slack_ts and _format_message as before, and returns the same JSON payload; then return await run_slack_tool(user_id, "get_channel_history", _do, ...) so run_slack_tool handles logging and errors uniformly. Ensure limit bounds and the response shape ("status","channel_id","messages","count") remain unchanged.

server/chat/backend/agent/tools/slack_tool.py
```diff
@@ -0,0 +1,225 @@
...
+
+        return json.dumps({
+            "status": "ok",
+            "channels": result,
+            "total": len(result),
+        })
+
+    except Exception as e:
+        logger.exception("[SlackTool] Failed to list channels")
+        return json.dumps({"error": f"Failed to list Slack channels: {e}"})
```
isiddharthsingh, line 127: Convert list_slack_channels to an async function and refactor Slack client retrieval and error handling into a shared async helper run_slack_tool(user_id, _do); run_slack_tool should: retrieve the client via get_slack_client_for_user(user_id), return a JSON error if no user_id or no client, catch exceptions, log with logger.exception and return a JSON error. Then implement list_slack_channels as async def list_slack_channels(user_id: str | None = None, **kwargs) -> str that calls await run_slack_tool(user_id, async def _do(client): ...), where the _do callback awaits client.list_channels(), builds the same result structure (id, name, topic, purpose, num_members, is_member) for up to 100 channels, and returns the {"status":"ok","channels":..., "total":...} JSON; ensure all blocking calls are awaited and all error paths return JSON strings as before.

server/chat/backend/agent/tools/slack_tool.py
```diff
@@ -0,0 +1,225 @@
...
+            })
+
+        return json.dumps({
+            "status": "ok",
+            "channels": result,
+            "total": len(result),
+        })
+
+    except Exception as e:
+        logger.exception("[SlackTool] Failed to list channels")
```
isiddharthsingh: Consider using INFO level logging

client/src/app/incidents/components/PostmortemPanel.tsx
```diff
@@ -238,6 +345,89 @@ export default function PostmortemPanel({ incidentId, incidentTitle, isVisible,
...
+                      </span>
+                    ) : (
+                      <button
+                        onClick={() => handleRestoreVersion(v.id)}
+                        className="inline-flex items-center gap-1 px-2 py-1 rounded-md text-[11px] text-blue-400 hover:text-blue-300 bg-blue-500/10 hover:bg-blue-500/20 border border-blue-500/20 transition-colors"
+                      >
+                        <RotateCcw className="w-3 h-3" />
+                        Restore
+                      </button>
+                    )}
```
isiddharthsingh: The current badge logic assumes versions[0] is the latest which is fragile; update the check in PostmortemPanel so it doesn't rely on versions[0] — either use an isCurrent flag returned by the backend (check v.isCurrent) or compute the latest versionNumber on the client (e.g., derive latestVersionNumber from the versions array via a reduce/Math.max and compare v.versionNumber === latestVersionNumber) and keep the existing UI behavior (show "Current" or the Restore button that calls handleRestoreVersion). Ensure you replace the versions[0]?.versionNumber reference with one of these explicit checks.

client/src/app/incidents/components/PostmortemPanel.tsx
```diff
@@ -87,6 +117,65 @@ export default function PostmortemPanel({ incidentId, incidentTitle, isVisible,
...
+        setEditContent(data.content);
+        setShowVersionHistory(false);
+        const versionsRes = await fetch(`/api/incidents/${incidentId}/postmortem/versions`);
+        const versionsData = await versionsRes.json();
+        setVersions(versionsData.versions ?? []);
+      }
+    } catch {
+      setRegenerateError('Failed to restore version');
+    }
+  };
```
isiddharthsingh, line 182: The restore handler handleRestoreVersion is reusing setRegenerateError for restore failures and also incorrectly reports failures from the subsequent versions fetch as a restore failure; change state to separate errors (e.g., restoreError and versionsError or a generic actionError) and update usages in PostmortemPanel so restore failures call setRestoreError while version-list fetch failures call setVersionsError (or setActionError). Also isolate the versions fetch in its own try/catch (or check versionsRes.ok) so a failure updating the versions list does not trigger the restore failure path; keep the successful restore behavior (setPostmortem, setEditContent, setShowVersionHistory) independent of the versions fetch outcome.

client/src/app/incidents/components/PostmortemPanel.tsx
```diff
@@ -87,6 +117,65 @@ export default function PostmortemPanel({ incidentId, incidentTitle, isVisible,
...
+    try {
+      const res = await fetch(`/api/incidents/${incidentId}/postmortem/versions`);
+      const data = await res.json();
+      setVersions(data.versions ?? []);
+    } catch {
+      setVersions([]);
+    }
+    setShowVersionHistory(true);
+    setLoadingVersions(false);
+  };
```
isiddharthsingh, line 157: Replace the direct fetch in handleLoadVersions with the service abstraction: add a new service method getPostmortemVersions (next to the existing regeneratePostmortem or regenerate method) that calls /api/incidents/{incidentId}/postmortem/versions and returns { versions }. Then update handleLoadVersions to call getPostmortemVersions(incidentId), await its result, setVersions(data.versions ?? []), and preserve the same error handling and loading/show state updates; also update any imports or service references so handleLoadVersions uses the new service method instead of fetch.

client/src/app/incidents/components/PostmortemPanel.tsx
```diff
@@ -87,6 +117,65 @@ export default function PostmortemPanel({ incidentId, incidentTitle, isVisible,
...
+        return;
+      }
+      setRegenerating(true);
+    } catch {
+      setRegenerateError('Failed to regenerate');
+      setPostmortemNotFound(true);
+    } finally {
+      setRegenerateSubmitting(false);
+    }
+  };
```
isiddharthsingh, line 140: The handler handleRegenerate currently calls fetch directly instead of using the existing incidents service; add a regeneratePostmortem(incidentId: string): Promise<{error?: string}> to the postmortem/incidents service that POSTs to /api/incidents/{incidentId}/postmortem/regenerate and returns { error } on non-OK responses, then update handleRegenerate to call postmortemService.regeneratePostmortem(incidentId) (await the result), set setRegenerateError(result.error || null) and setPostmortemNotFound(true) when an error is returned, keep the existing try/catch/finally flow and state updates (setRegenerateSubmitting, setRegenerating) but remove the inline fetch logic; ensure you reference handleRegenerate and the new regeneratePostmortem function when making edits.

server/chat/backend/agent/tools/postmortem_tool.py
```diff
@@ -0,0 +1,147 @@
...
+                )
+                postmortem_id = str(cursor.fetchone()[0])
+
+                # Create a version row
+                cursor.execute(
+                    """SELECT COALESCE(MAX(version_number), 0) + 1
+                       FROM postmortem_versions WHERE postmortem_id = %s""",
+                    (postmortem_id,),
+                )
+                next_version = cursor.fetchone()[0]
```
isiddharthsingh: The current computation of next_version using "SELECT COALESCE(MAX(version_number), 0) + 1" in save_postmortem can race; wrap the read/insert in a DB transaction and acquire a lock before computing next_version (e.g., SELECT ... FOR UPDATE on the parent postmortem row or on the postmortem_versions rows for that postmortem), then compute next_version and INSERT the new version before committing; update the cursor.execute call that fetches MAX(version_number) to use FOR UPDATE within the same transaction and ensure the transaction is committed/rolled back around the insert to prevent duplicate version numbers.

server/chat/backend/agent/tools/postmortem_tool.py
```diff
@@ -0,0 +1,147 @@
...
+                                               generation_session_id, generated_at, updated_at)
+                       VALUES (%s, %s, %s, %s, %s, CURRENT_TIMESTAMP, CURRENT_TIMESTAMP)
+                       ON CONFLICT (incident_id)
+                       DO UPDATE SET content = EXCLUDED.content,
+                                     generation_session_id = EXCLUDED.generation_session_id,
+                                     updated_at = CURRENT_TIMESTAMP
+                       RETURNING id""",
+                    (incident_id, user_id, org_id, content, session_id),
+                )
+                postmortem_id = str(cursor.fetchone()[0])
```
isiddharthsingh: The INSERT ... RETURNING block uses cursor.fetchone() and immediately indexes [0] into it, which will raise a TypeError if fetchone() returns None (e.g., RLS blocked); change this to capture the result in a variable (e.g., result = cursor.fetchone()), check if result is None, and if so raise or return a clear, specific error that explains the upsert returned no row (include context like incident_id, generation_session_id), otherwise set postmortem_id = str(result[0]); update references to postmortem_id accordingly.

server/utils/db/db_utils.py
```diff
@@ -2005,6 +2023,39 @@ def initialize_tables():
...
+                cursor.execute("""
+                    CREATE TABLE IF NOT EXISTS postmortem_versions (
+                        id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
+                        postmortem_id UUID NOT NULL REFERENCES postmortems(id) ON DELETE CASCADE,
+                        org_id VARCHAR(255) NOT NULL,
+                        user_id VARCHAR(255) NOT NULL,
+                        content TEXT NOT NULL,
+                        version_number INTEGER NOT NULL DEFAULT 1,
+                        source VARCHAR(50) NOT NULL DEFAULT 'manual',
+                        generation_session_id UUID,
```
isiddharthsingh: he column type for generation_session_id is inconsistent between postmortems (VARCHAR(255)) and postmortem_versions (UUID); make them consistent by changing the generation_session_id in postmortem_versions to VARCHAR(255) (or alternatively change postmortems to UUID if you prefer UUIDs) and update the migration that creates/alters postmortem_versions accordingly so both tables use the same type; ensure any related constraints, indexes, and code that reads/writes generation_session_id (e.g., where postmortem_versions is referenced) are adjusted to the chosen type.

server/utils/db/db_utils.py
```diff
@@ -2005,6 +2023,39 @@ def initialize_tables():
...
+                    CREATE INDEX IF NOT EXISTS idx_postmortem_versions_postmortem
+                        ON postmortem_versions(postmortem_id, version_number DESC);
+                    CREATE INDEX IF NOT EXISTS idx_postmortem_versions_org
+                        ON postmortem_versions(org_id);
+                """)
+                logging.info("Created postmortem_versions table (if not exists).")
+                conn.commit()
+            except Exception as e:
+                logging.warning(f"Error creating postmortem_versions table: {e}")
+                conn.rollback()
```
isiddharthsingh, line 2051: The postmortem_versions table lacks RLS protection; add "postmortem_versions" to the rls_tables list alongside the existing "postmortem_exports" entry so it will get the same RLS setup and FORCE ROW LEVEL SECURITY applied; update the rls_tables declaration (the list used when configuring RLS) to include postmortem_versions and re-run the RLS creation logic that applies policies and FORCE ROW LEVEL SECURITY for that table.

server/routes/postmortem_routes.py
```diff
@@ -45,6 +46,69 @@ def _format_timestamp(ts) -> Optional[str]:
...
+           FROM postmortem_versions WHERE postmortem_id = %s""",
+        (postmortem_id,),
+    )
+    next_version = cursor.fetchone()[0]
+    cursor.execute(
+        """INSERT INTO postmortem_versions
+           (postmortem_id, org_id, user_id, content, version_number, source)
+           VALUES (%s, %s, %s, %s, %s, %s)""",
+        (postmortem_id, org_id, user_id, content, next_version, source),
+    )
```
isiddharthsingh: The _create_version function has a race condition because it computes MAX(version_number) then inserts separately and also claims to "return the next version number" but currently returns None; fix by making the version calculation+insert atomic (e.g. replace the separate SELECT and INSERT with a single INSERT ... SELECT that computes COALESCE(MAX(version_number),0)+1 for the given postmortem_id and uses RETURNING version_number), fetch the returned version_number from the cursor and return it from _create_version, and ensure this runs inside a transaction so the INSERT...SELECT/RETURNING is atomic for concurrent calls.

server/routes/postmortem_routes.py
```diff
@@ -156,34 +208,140 @@ def update_postmortem(user_id, incident_id):
...
+        """UPDATE postmortems
+           SET content = %s, updated_at = CURRENT_TIMESTAMP
+           WHERE id = %s""",
+        (restored_content, postmortem_id),
+    )
+    conn.commit()
+
+    record_audit_event(org_id, user_id, "restore_postmortem_version", "postmortem", incident_id,
+                       {"version_id": version_id}, request)
+    return jsonify({"success": True, "content": restored_content})
```
isiddharthsingh, line 338: The restore code updates postmortems.content but doesn't create a postmortem_versions entry for the pre-restore state; before updating, SELECT the current content of the postmortem (use postmortem_id) and INSERT it into postmortem_versions (including metadata like created_by/user_id and timestamp), retrieve that new version's id, then proceed to UPDATE postmortems.content with restored_content and commit; finally pass the new version id to record_audit_event instead of the old version_id so the restore is recorded in version history.

client/src/app/actions/page.tsx
```diff
@@ -296,6 +305,17 @@ function ActionDetailView({ actionId, onBack, onEdit }: { readonly actionId: str
...
+  const handleRestoreDefault = useCallback(async () => {
+    if (!confirm('Restore instructions to the built-in default? Your customizations will be lost.')) return;
+    try {
+      await fetchR(`/api/actions/${actionId}/restore-default`, { method: 'POST' });
+      toast({ title: 'Restored to default instructions' });
+      mutate();
+    } catch {
+      toast({ title: 'Failed to restore default', variant: 'destructive' });
+    }
+  }, [actionId, mutate, toast]);
```
isiddharthsingh, line 317: The handler handleRestoreDefault currently calls fetchR to POST to /api/actions/${actionId}/restore-default; replace that call with forwardRequest from '@/lib/backend-proxy' so the request is routed through the backend proxy. Update the import to include forwardRequest, call await forwardRequest(/api/actions/${actionId}/restore-default, { method: 'POST' }) in place of fetchR, and keep the existing try/catch, toast messages, and mutate() behavior unchanged.

client/src/lib/services/incidents.ts
```diff
@@ -537,14 +551,21 @@ export const incidentsService = {
...
+        return { data: null, generating: true };
+      }
+      if (res.status === 404) {
         return { data: null };
       }
+      if (!res.ok) {
+        return { data: null, error: `Request failed: ${res.status}` };
+      }
+      const json = await res.json();
+      return { data: json.postmortem || null };
```
isiddharthsingh: Replace the direct fetch call in the function that retrieves the postmortem with the project's API wrapper: use apiRequest (or apiGet if it supports status info) from '@/lib/services/api-client' to call '/api/incidents/${incidentId}/postmortem' so the request goes through the Next.js proxy/forwardRequest flow; preserve the existing status-specific logic by catching the ApiError (or inspecting the response returned by apiRequest) to return { generating: true } for 202, { data: null } for 404, and the error object for other non-OK statuses, then parse and return json.postmortem when successful. Ensure you reference the existing function name in incidents.ts and replace the fetch usage and response handling only, keeping the surrounding logic intact.

server/chat/backend/agent/skills/integrations/postmortem/SKILL.md
```diff
@@ -0,0 +1,84 @@
...
+## Overview
+Core tools for reading and writing postmortem documents. Always available. Used by the built-in "Generate Postmortem" action and accessible during interactive chat for postmortem-related tasks.
+
+## Tools
+
+### `get_postmortem(incident_id)`
+Read the current postmortem for an incident. Returns markdown content, generated_at, and updated_at timestamps. Returns `status: "not_found"` if no postmortem exists yet.
+
+### `save_postmortem(incident_id, content)`
+Write or update a postmortem. Each call creates a new version (append-only history). Content must be complete markdown — partial updates are not supported. Max 100,000 characters.
```
isiddharthsingh: The documentation for save_postmortem and get_postmortem is inconsistent: save_postmortem mentions "append-only history" but get_postmortem only documents returning the current postmortem; clarify whether historical versions are retrievable and how. Update SKILL.md to state explicitly whether historical retrieval is supported; if it is, document the API/method (e.g., a get_postmortem_versions or a versions field/parameter on get_postmortem) and how to call it and what it returns; if it is not supported, remove or reword "append-only history" in the save_postmortem description to avoid implying access to past versions and note that only the latest version is retrievable via get_postmortem. Ensure references to get_postmortem and save_postmortem are used so readers can find the relevant functionality.

server/chat/background/task.py
```diff
@@ -28,6 +28,7 @@
     send_google_chat_investigation_completed_notification,
 )
 from connectors.google_chat_connector.client import get_chat_app_client
+from chat.backend.agent.tools.slack_tool import is_slack_connected
```
isiddharthsingh: The import and use of is_slack_connected (from chat.backend.agent.tools.slack_tool) is likely gating notifications with an agent-tool-level check and is inconsistent with the local _has_google_chat_connected helper; verify whether is_slack_connected's semantics are appropriate for notification gating and if not, add a dedicated notification-level check (e.g., is_slack_notifications_enabled or similar) and use it instead of is_slack_connected, or alternatively extract a centralized notification helper used by both Slack and Google Chat; update the import/usage in task.py (replace is_slack_connected at the notification gating site and ensure _has_google_chat_connected is aligned or replaced) so both Slack and Google Chat use the same notification-check pattern.

server/services/actions/postmortem_action.py
```diff
@@ -0,0 +1,272 @@
...
+                    "summary": (row[3] or "")[:4000],
+                    "started_at": row[4].isoformat() if row[4] else None,
+                    "analyzed_at": row[5].isoformat() if row[5] else None,
+                    "resolved_at": row[6].isoformat() if row[6] else None,
+                    "source_type": row[7] or "unknown",
+                    "environment": row[8] or "unknown",
+                }
+    except Exception:
+        logger.exception("[PostmortemAction] Failed to load incident for postmortem")
+        return None
```
isiddharthsingh, line 200: The _load_incident_for_prompt function is using get_admin_connection() which bypasses RLS and defeats the set_rls_context call; change the connection call to use get_connection() so RLS is enforced when querying the incidents table, keep the set_rls_context(cur, conn, user_id, ...) call intact, and ensure the rest of the function behavior (field selection, row handling, and ISO formatting) remains unchanged.

server/services/actions/system_actions.py
```diff
@@ -0,0 +1,85 @@
...
+    },
+]
+
+
+def _get_default_instructions(system_key: str) -> str:
+    """Resolve the default instructions for a given system action."""
+    if system_key == "generate_postmortem":
+        from services.actions.postmortem_action import DEFAULT_POSTMORTEM_INSTRUCTIONS
+        return DEFAULT_POSTMORTEM_INSTRUCTIONS
+    raise ValueError(f"Unknown system action: {system_key}")
```
isiddharthsingh, line 35: currently performs a local import of DEFAULT_POSTMORTEM_INSTRUCTIONS from postmortem_action to avoid a suspected circular import; verify whether importing DEFAULT_POSTMORTEM_INSTRUCTIONS at module top level causes a circular dependency by attempting the module-level import and running tests/static analysis, and if it does not, move the import to the module-level for readability; if a circular import exists, either (a) document the local import in _get_default_instructions with a brief comment explaining it's intentional to avoid circular imports, or (b) refactor by extracting DEFAULT_POSTMORTEM_INSTRUCTIONS into a small shared constants module (e.g., an actions defaults module) that both postmortem_action and system_actions can import, then replace the local import with a module-level import of the constant.

server/chat/backend/agent/tools/cloud_tools.py
```diff
@@ -1305,6 +1305,29 @@ def cloud_exec_wrapper(provider: str, command: str, output_file: Optional[str] =
     _action_id = getattr(state_context, 'trigger_action_id', None) if state_context else None
     if _action_id:
         tool_functions.append((trigger_action, "trigger_action"))
+
+    # Postmortem tools (always available)
+    from .postmortem_tool import get_postmortem, save_postmortem
+    tool_functions.append((get_postmortem, "get_postmortem"))
+    tool_functions.append((save_postmortem, "save_postmortem"))
```
isiddharthsingh: The postmortem tool imports are not wrapped in a try/except block, unlike other optional integration tools (Slack, Bitbucket, Confluence, Notion, Jira, SharePoint). If the postmortem_tool module has import errors, it will break the entire get_cloud_tools() function and prevent registration of all other tools.

server/routes/actions/actions_routes.py
```diff
@@ -323,16 +355,49 @@ def delete_action(user_id, action_id):
...
+                    (action_id,),
+                )
+                row = cur.fetchone()
+                if not row:
+                    return jsonify({"error": _ERR_NOT_FOUND}), 404
+                if not row[0]:
+                    return jsonify({"error": "Only system actions can be restored to default"}), 400
+                cur.execute(
+                    "UPDATE actions SET instructions = %s, updated_at = CURRENT_TIMESTAMP WHERE id = %s",
+                    (row[1], action_id),
```
isiddharthsingh: Replace the SQL CURRENT_TIMESTAMP usage with an explicit Python UTC timestamp to match the rest of the file: import datetime/timezone if needed at the top, compute a timestamp with datetime.now(timezone.utc) where the UPDATE is executed (the cur.execute call that updates instructions using row[1] and action_id) and pass that timestamp as a parameter in the SQL (update the parameter tuple for the cur.execute in actions_routes.py) so all timestamp handling uses Python's timezone-aware datetime.

server/routes/incidents_routes.py
```diff
@@ -1090,20 +1090,20 @@ def update_incident(user_id, incident_id: str):
...
                             sanitize(incident_id),
                         )
-                    except Exception as pm_exc:
+                    except Exception as act_exc:
                         logger.warning(
-                            "[INCIDENTS] Failed to trigger postmortem generation for incident %s: %s",
+                            "[INCIDENTS] on-resolved actions failed for %s: %s",
                             sanitize(incident_id),
-                            pm_exc,
+                            act_exc,
```
isiddharthsingh: Consider sanitizing the exception object for consistency.

server/utils/db/db_utils.py
```diff
@@ -1873,7 +1890,8 @@ def initialize_tables():
                         notion_page_id TEXT,
                         notion_page_url TEXT,
                         notion_exported_at TIMESTAMP,
-                        notion_database_id TEXT
+                        notion_database_id TEXT,
+                        generation_session_id VARCHAR(255)
```
isiddharthsingh: On existing deployments the postmortems table already exists, so adding generation_session_id only inside CREATE TABLE IF NOT EXISTS does not add the column. The new GET and regeneration paths select/insert postmortems.generation_session_id, so upgraded environments will start returning database errors for postmortem reads or generation until the DB is rebuilt; add an ALTER TABLE postmortems ADD COLUMN IF NOT EXISTS generation_session_id ... migration.

server/routes/incidents_routes.py
```diff
@@ -1090,20 +1090,20 @@ def update_incident(user_id, incident_id: str):
...
                     _record_audit_event(org_id or "", user_id, f"incident_{event_type}", "incident", incident_id, {"from": previous_status, "to": data["status"]}, request)
 
-                # Trigger postmortem generation only on transition to resolved
+                # Trigger on_incident actions for the "resolved" event
                 if data.get("status") == "resolved" and previous_status != "resolved":
                     try:
-                        from chat.background.postmortem_generator import generate_postmortem
-                        generate_postmortem.delay(incident_id, user_id, org_id)
+                        from services.actions.executor import dispatch_on_incident_actions
+                        dispatch_on_incident_actions(user_id, incident_id, timing="resolved")
```
isiddharthsingh, line 1100: For orgs that have not opened the Actions page after this deploy, no generate_postmortem system action row exists because seeding only happens in list_actions. In that case resolving an incident now calls dispatch_on_incident_actions, the dispatcher finds no matching row, and automatic postmortem generation is silently skipped, whereas the deleted Celery task used to run unconditionally on resolve.

server/services/actions/postmortem_action.py
```diff
@@ -0,0 +1,272 @@
...
+
+                set_rls_context(cur, conn, user_id, log_prefix="[PostmortemAction:reserve]")
+
+                cur.execute(
+                    """INSERT INTO postmortems (incident_id, user_id, org_id, content, generation_session_id)
+                       VALUES (%s, %s, %s, NULL, %s)
+                       ON CONFLICT (incident_id)
+                       DO UPDATE SET content = NULL,
+                                     generation_session_id = EXCLUDED.generation_session_id,
+                                     updated_at = CURRENT_TIMESTAMP""",
```
isiddharthsingh, line 274: When regenerating an existing postmortem, this conflict handler immediately sets content = NULL and the code never snapshots the previous content despite the comment. If enqueueing, the agent, or save_postmortem fails, the prior reviewed postmortem is lost and the GET endpoint remains in the 202 generating state; it also prevents the agent's first get_postmortem call from using the old content as a baseline.

server/routes/postmortem_routes.py
```diff
@@ -156,34 +208,140 @@ def update_postmortem(user_id, incident_id):
...
+
+    restored_content = row[0]
+
+    cursor.execute(
+        """UPDATE postmortems
+           SET content = %s, updated_at = CURRENT_TIMESTAMP
+           WHERE id = %s""",
+        (restored_content, postmortem_id),
+    )
+    conn.commit()
```
isiddharthsingh, line 334: Restoring an older version updates the current postmortems.content but does not insert a new postmortem_versions row. After restoring v1 while v2 exists, the versions endpoint still sorts v2 first, so the UI labels v2 as Current and hides its restore button even though v1 is now the live content; create a restore-sourced version for the restored content.

server/services/actions/postmortem_action.py
```diff
@@ -0,0 +1,272 @@
...
+    run_id = None
+    if action_id:
+        from services.actions.executor import _create_run, _update_run
+        run_id = _create_run(
+            action_id=action_id,
+            org_id=org_id,
+            user_id=user_id,
+            incident_id=incident_id,
+            trigger_context=trigger_meta,
+            status="running",
```
isiddharthsingh, line 130: This special postmortem action path creates an action_runs row with status="running" before session creation, reservation, and Celery enqueue, but unlike dispatch_action it does not catch failures and update the run to error. If session creation or enqueueing fails, the Actions page will show a run stuck as running forever, so the setup steps need the same error handling used by the generic executor.

server/chat/backend/agent/tools/slack_tool.py
OlivierTrudeau: Done — changed to logger.info across all 3 functions.

client/src/app/incidents/components/PostmortemPanel.tsx
OlivierTrudeau: Done — replaced with Math.max(...versions.map(v => v.versionNumber)).

client/src/app/incidents/components/PostmortemPanel.tsx
OlivierTrudeau: Done — version-list refresh now in its own try/catch, isolated from restore error.

client/src/app/incidents/components/PostmortemPanel.tsx
OlivierTrudeau: Done — added postmortemService.getVersions(), handleLoadVersions uses it now.

client/src/app/incidents/components/PostmortemPanel.tsx
OlivierTrudeau: Done — added postmortemService.regeneratePostmortem(), removed inline fetch.

server/chat/backend/agent/tools/postmortem_tool.py
OlivierTrudeau: Done — atomic INSERT...SELECT COALESCE(MAX(version_number),0)+1...RETURNING.

server/chat/backend/agent/tools/postmortem_tool.py
OlivierTrudeau: Done — null check added before indexing fetchone() result.

server/utils/db/db_utils.py
OlivierTrudeau: Done — changed to VARCHAR(255) to match postmortems table.

server/utils/db/db_utils.py
OlivierTrudeau: Done — added to rls_tables.

server/routes/postmortem_routes.py
OlivierTrudeau: Done — atomic INSERT...SELECT with RETURNING version_number, now returns int.

server/routes/postmortem_routes.py
OlivierTrudeau: Done — snapshots current content as pre_restore version before overwriting.

server/chat/backend/agent/skills/integrations/postmortem/SKILL.md
OlivierTrudeau: Done — clarified: only latest version via tool, history is UI-only.

server/chat/background/task.py
OlivierTrudeau: Done — now imports get_slack_client_for_user directly from the connector, consistent with Google Chat pattern.

server/services/actions/postmortem_action.py
OlivierTrudeau: Done — changed to get_connection().

server/services/actions/system_actions.py
OlivierTrudeau: Done — moved to top-level import, no circular dependency.

server/chat/backend/agent/tools/cloud_tools.py
OlivierTrudeau: Done — wrapped in try/except ImportError like other optional tools.

server/routes/actions/actions_routes.py
OlivierTrudeau: Done — replaced with datetime.now(timezone.utc).

server/routes/incidents_routes.py
OlivierTrudeau: Done — removed exception object from log output.

server/utils/db/db_utils.py
OlivierTrudeau: Done — added ALTER TABLE postmortems ADD COLUMN IF NOT EXISTS generation_session_id migration.

server/routes/incidents_routes.py
OlivierTrudeau: Done — added seed_system_actions() call before dispatching on-resolved actions.

server/chat/backend/agent/tools/slack_tool.py
OlivierTrudeau: Extracted duplicated auth/error into constants but kept functions flat and sync — the callback pattern decreased readability for 3 tools with no practical benefit. Tool registration doesn't use async.

server/chat/backend/agent/tools/slack_tool.py
OlivierTrudeau: See above — kept flat, tool registration is sync.

server/chat/backend/agent/tools/slack_tool.py
OlivierTrudeau: See above — kept flat, tool registration is sync.

client/src/app/actions/page.tsx
OlivierTrudeau: Skipping — forwardRequest is a server-side utility for API routes. fetchR is the correct client-side wrapper (used throughout this file).

client/src/lib/services/incidents.ts
OlivierTrudeau: Kept fetch — apiRequest throws on non-OK which doesn't work here since we need to branch on 202/404 as non-error cases. Removed console.error for expected statuses.

server/services/actions/postmortem_action.py
OlivierTrudeau: Done — _reserve_postmortem_row now snapshots existing content as a pre_regenerate version before setting NULL.

server/routes/postmortem_routes.py
OlivierTrudeau: Done — restore now inserts a restore-sourced version row for the restored content, so the UI correctly reflects current state.

server/services/actions/postmortem_action.py
OlivierTrudeau: Done — enqueue wrapped in try/except, marks action_run as error on failure.

server/services/actions/postmortem_action.py
OlivierTrudeau: Update: reverted the pre_regenerate snapshot — previous content is already in postmortem_versions from the original save. A pre_regenerate row would duplicate the latest version and create confusing numbering (v1, v2=copy of v1, v3=actual new).

client/src/app/incidents/components/PostmortemPanel.tsx
```diff
@@ -33,40 +34,69 @@ export default function PostmortemPanel({ incidentId, incidentTitle, isVisible,
...
+      setEditContent(result.data.content);
+      setPostmortemNotFound(false);
+      setRegenerating(false);
+    } else if (result.generating) {
+      setPostmortemNotFound(false);
+      setRegenerating(true);
+    } else if (!regenerating) {
+      setPostmortemNotFound(true);
+    }
+  }, [incidentId, regenerating]);
```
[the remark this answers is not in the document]
OlivierTrudeau: Fixed. loadPostmortem now early returns on error so a transient 5xx doesn't wipe the displayed content.

client/src/components/tool-calls/tool-command-parser.ts
```diff
@@ -494,3 +494,37 @@ export function parseCloudflareCommand(toolName: string, toolInput: string): str
...
+  if (toolName === "get_thread_replies") {
+    const parts: string[] = []
+    if (args.channel_id) parts.push(`#${args.channel_id}`)
+    if (args.thread_ts) parts.push(`thread ${args.thread_ts}`)
+    const detail = parts.length ? ` (${parts.join(", ")})` : ""
+    return `Slack: Get thread replies${detail}`
+  }
+
+  return `Slack: ${toolName.replace(/slack_?/g, "").replace(/_/g, " ")}`
+}
```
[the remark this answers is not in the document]
OlivierTrudeau: Skipping this one. It's a flat if/else dispatch for 3 tool names, not genuinely complex. Refactoring into a map would make it harder to read for no real benefit.

client/src/lib/services/incidents.ts
```diff
@@ -563,6 +583,34 @@ export const postmortemService = {
...
+  async getVersions(incidentId: string): Promise<{ versions: PostmortemVersion[] }> {
+    try {
+      const res = await fetch(`/api/incidents/${incidentId}/postmortem/versions`, {
+        credentials: 'include',
+      });
+      const data = await res.json();
+      return { versions: data.versions ?? [] };
+    } catch {
+      return { versions: [] };
+    }
```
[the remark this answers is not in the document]
OlivierTrudeau: Good catch. getVersions now returns an error field so the UI can distinguish between empty history and a failed fetch.

server/chat/backend/agent/skills/integrations/slack/SKILL.md
```diff
@@ -0,0 +1,47 @@
...
+
+1. Call `list_slack_channels` — scan names/topics for the affected service or "incident"/"oncall" keywords
+2. Call `get_channel_history` on the most relevant channels, scoped to the incident time window
+3. Look for messages about: deployments, rollbacks, alerts firing, team handoffs, escalations
+4. If a message has `reply_count > 0` and looks relevant, call `get_thread_replies` for full context
+
+## Limitations
+- Read-only — cannot post messages
+- Bot must be a member of the channel to read it
+- No cross-channel search — must check channels individually by name
```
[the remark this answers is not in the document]
OlivierTrudeau: Will clean up the markdown formatting in a follow up, not blocking for this PR.

server/chat/backend/agent/tools/cloud_tools.py
```diff
@@ -1305,6 +1305,32 @@ def cloud_exec_wrapper(provider: str, command: str, output_file: Optional[str] =
...
         tool_functions.append((trigger_action, "trigger_action"))
+
+    # Postmortem tools (always available)
+    try:
+        from .postmortem_tool import get_postmortem, save_postmortem
+        tool_functions.append((get_postmortem, "get_postmortem"))
+        tool_functions.append((save_postmortem, "save_postmortem"))
+    except ImportError:
+        logger.debug("Postmortem tools not available")
+
```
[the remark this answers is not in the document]
OlivierTrudeau: Agreed, bumped to logger.warning since these are core tools for this feature and a silent failure would be confusing.

server/routes/actions/actions_routes.py
```diff
@@ -149,17 +150,32 @@ def list_actions(user_id):
...
+                if org_row and org_row[0]:
+                    cur.execute(
+                        "SELECT 1 FROM actions WHERE org_id = %s AND is_system = true LIMIT 1",
+                        (org_row[0],),
+                    )
+                    if not cur.fetchone():
+                        try:
+                            seed_system_actions(org_row[0], user_id)
+                        except Exception:
+                            logger.debug("Failed to lazy-seed system actions")
```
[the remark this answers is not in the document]
OlivierTrudeau: Fixed. The guard now checks whether all expected system_keys exist rather than just any one, so new built in actions get backfilled on partially seeded orgs.

server/routes/postmortem_routes.py
```diff
@@ -156,34 +207,149 @@ def update_postmortem(user_id, incident_id):
...
-    org_id = get_org_id_from_request()
+    _create_version(cursor, postmortem_id, org_id, user_id, content, source="manual")
+
+    cursor.execute(
+        """UPDATE postmortems
+           SET content = %s, updated_at = CURRENT_TIMESTAMP
+           WHERE id = %s""",
+        (content, postmortem_id),
+    )
+    conn.commit()
```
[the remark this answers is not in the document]
OlivierTrudeau: Good catch. update_postmortem now snapshots the current content as a pre_edit version before overwriting, so restore actually has the old state available.

server/services/actions/postmortem_action.py
```diff
@@ -0,0 +1,272 @@
...
+    with db_pool.get_connection() as conn:
+        with conn.cursor() as cur:
+            set_rls_context(cur, conn, user_id, log_prefix="[PostmortemAction]")
+            cur.execute(
+                """SELECT id FROM postmortems
+                   WHERE incident_id = %s AND content IS NULL""",
+                (incident_id,),
+            )
+            if cur.fetchone():
+                raise ValueError("Postmortem generation already running for this incident")
```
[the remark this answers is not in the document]
OlivierTrudeau: Acknowledged but leaving for now. Two concurrent regenerate calls by the same user within the same millisecond is extremely unlikely in practice, and adding advisory locks or SELECT FOR UPDATE here adds complexity for a near zero probability scenario. Happy to revisit if it ever surfaces.

server/services/actions/system_actions.py
```diff
@@ -0,0 +1,86 @@
...
+                            instructions,
+                            action_def["trigger_type"],
+                            json.dumps(action_def["trigger_config"]),
+                            action_def["mode"],
+                            key,
+                            instructions,
+                        ),
+                    )
+                    created += 1
+                    logger.info("[SystemActions] Seeded '%s' for org %s", key, org_id)
```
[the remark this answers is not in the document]
OlivierTrudeau: org_id here is always resolved from the DB (queried from the users table) or from the authenticated session, never from raw user input. False positive from the static analyzer.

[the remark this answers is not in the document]
OlivierTrudeau: False positive. org_id is queried from the users table internally, not user supplied input. No injection risk here.

server/services/actions/system_actions.py
```diff
@@ -0,0 +1,86 @@
...
+
+            conn.commit()
+    except Exception:
+        logger.exception("[SystemActions] Failed to seed actions for org %s", org_id)
```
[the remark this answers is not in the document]
OlivierTrudeau: Same as above, org_id is DB derived. Not user controlled.

server/routes/postmortem_routes.py
```diff
@@ -45,6 +46,74 @@ def _format_timestamp(ts) -> Optional[str]:
...
+           RETURNING id, version_number""",
+        (postmortem_id, org_id, user_id, content, postmortem_id, source),
+    )
+    row = cursor.fetchone()
+    version_id, version_number = row[0], row[1]
+    cursor.execute(
+        "UPDATE postmortems SET current_version_id = %s WHERE id = %s",
+        (str(version_id), postmortem_id),
+    )
+    return version_number
```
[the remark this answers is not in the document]
OlivierTrudeau: Good catch. Added a set_current parameter (defaults to True). The pre_edit snapshot now passes set_current=False so it doesn't advance the pointer.

server/chat/backend/agent/tools/postmortem_tool.py
```diff
@@ -0,0 +1,157 @@
...
+                    """INSERT INTO postmortems (incident_id, user_id, org_id, content,
+                                               generation_session_id, generated_at, updated_at)
+                       VALUES (%s, %s, %s, %s, %s, CURRENT_TIMESTAMP, CURRENT_TIMESTAMP)
+                       ON CONFLICT (incident_id)
+                       DO UPDATE SET content = EXCLUDED.content,
+                                     generation_session_id = EXCLUDED.generation_session_id,
+                                     updated_at = CURRENT_TIMESTAMP
+                       RETURNING id""",
+                    (incident_id, user_id, org_id, content, session_id),
+                )
```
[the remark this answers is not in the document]
OlivierTrudeau: Fixed. Now resolves org_id from the incident table under RLS instead of the user row, so a mismatched incident_id from another tenant will correctly return not found.

server/chat/backend/agent/tools/cloud_tools.py
```diff
@@ -1305,6 +1305,32 @@ def cloud_exec_wrapper(provider: str, command: str, output_file: Optional[str] =
...
+        )
+        if user_id and is_slack_connected(user_id):
+            tool_functions.append((list_slack_channels, "list_slack_channels"))
+            tool_functions.append((get_channel_history, "get_channel_history"))
+            tool_functions.append((get_thread_replies, "get_thread_replies"))
+            logging.info(f"Added Slack tools for user {user_id}")
+        else:
+            logging.debug(f"Slack tools not added - user {user_id} not connected")
+    except Exception as e:
+        logging.warning(f"Failed to add Slack tools: {e}")
```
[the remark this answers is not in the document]
OlivierTrudeau: get_cloud_tools() is invoked fresh per agent run, there's no persistent cross session cache. The tool list is rebuilt on every chat message, so connecting/disconnecting Slack takes effect on the next invocation without any invalidation needed.

server/services/actions/postmortem_action.py
```diff
@@ -0,0 +1,279 @@
...
+) -> str:
+    """Dispatch the postmortem generation as a background agent session.
+
+    Args:
+        user_id: The user who owns the incident
+        incident_id: The incident to generate a postmortem for
+        custom_instructions: Optional user-customized instructions (layered on top of defaults)
+
+    Returns:
+        run_id of the dispatched action
```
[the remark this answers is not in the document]
OlivierTrudeau: Intentional. The frontend polls by session_id (it subscribes to the chat session for streaming progress), not by action run_id. Renaming the field would be a breaking change for the existing polling logic. The run_id is tracked internally for action run history but the caller needs the session to connect to.

isiddharthsingh (approved):

PR #383, opened 2026-05-11
fix: prevent reasoning tokens from streaming as normal output

damianloch:
## Summary

- Google models (via OpenRouter and direct Gemini SDK) were leaking reasoning/thinking tokens into the frontend chat, appearing as regular assistant output
- `_ReasoningChatOpenAI` now parses `delta.reasoning_details` (array format used by Google models via OpenRouter) in addition to `delta.reasoning`
- For user-facing chats, Google models via OpenRouter use `reasoning.exclude=true` so reasoning is used internally but not returned in the stream
- `workflow.py` streaming path uses `include_thinking=False` and only forwards reasoning content in background RCA chats (for incident thought tracking)

## Test plan

- [ ] Send a message using a Google model (e.g. Gemini 3.1 Pro) via OpenRouter in a user-facing chat — verify no reasoning/thinking tokens appear in the response
- [ ] Trigger a background RCA investigation with a Google model — verify reasoning still flows into incident thoughts
- [ ] Test with direct Gemini provider (Google AI Studio) — verify thinking blocks don't leak into user-visible output
- [ ] Test OpenAI reasoning models (o-series) via OpenRouter — verify they still work correctly (no regression)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed streaming so internal AI reasoning no longer replaces or appears in regular bot messages; reasoning is shown only where appropriate.
  * Ensured reasoning is visible in streamed output only for background analysis sessions.

* **Improvements**
  * Improved separation of reasoning vs. user-facing text during streaming and fallbacks.
  * Tuned AI reasoning settings to produce higher-quality analysis when applicable.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Arvo-AI/aurora/pull/383)

### DIFF
--- a/server/chat/backend/agent/agent.py
+++ b/server/chat/backend/agent/agent.py
@@ -21,6 +21,29 @@ import asyncio
 _DIRECT_ONLY_PROVIDERS = frozenset({"vertex", "ollama"})
 
 
+def _extract_reasoning_from_delta(delta: dict) -> tuple:
+    """Extract reasoning text from an OpenRouter streaming delta.
+
+    Returns (reasoning_text, has_reasoning_details) where has_reasoning_details
+    indicates the delta carried a reasoning_details array regardless of whether
+    any text was extractable.
+    """
+    reasoning = delta.get("reasoning") or ""
+    reasoning_details = delta.get("reasoning_details")
+    has_reasoning_details = bool(reasoning_details) and isinstance(reasoning_details, list)
+
+    if has_reasoning_details:
+        parts = [
+            detail.get("text", "")
+            for detail in reasoning_details
+            if isinstance(detail, dict) and detail.get("text")
+        ]
+        if parts:
+            reasoning = (reasoning + "".join(parts)) if reasoning else "".join(parts)
+
+    return reasoning, has_reasoning_details
+
+
 class _ReasoningChatOpenAI(ChatOpenAI):
     """ChatOpenAI subclass that captures OpenRouter reasoning fields.
 
@@ -41,27 +64,18 @@ class _ReasoningChatOpenAI(ChatOpenAI):
         if result is None:
             return None
         choices = chunk.get("choices") or chunk.get("chunk", {}).get("choices") or []
-        if choices:
-            delta = choices[0].get("delta") or {}
+        if not choices:
+            return result
 
-            reasoning = delta.get("reasoning") or ""
+        delta = choices[0].get("delta") or {}
+        reasoning, has_reasoning_details = _extract_reasoning_from_delta(delta)
 
-            reasoning_details = delta.get("reasoning_details")
-            if reasoning_details and isinstance(reasoning_details, list):
-                parts = []
-                for detail in reasoning_details:
-                    if isinstance(detail, dict):
-                        text = detail.get("text", "")
-                        if text:
-                            parts.append(text)
-                if parts:
-                    reasoning = (reasoning + "".join(parts)) if reasoning else "".join(parts)
+        if reasoning and hasattr(result.message, "additional_kwargs"):
+            result.message.additional_kwargs["reasoning_content"] = reasoning
 
-            if reasoning and hasattr(result.message, "additional_kwargs"):
-                result.message.additional_kwargs["reasoning_content"] = reasoning
-                # Clear content so reasoning-only chunks don't appear as output
-                if not delta.get("content"):
-                    result.message.content = ""
+        if (reasoning or has_reasoning_details) and not delta.get("content"):
+            result.message.content = ""
+
         return result
 
 class Agent:
--- a/server/chat/backend/agent/agent.py
+++ b/server/chat/backend/agent/agent.py
@@ -528,6 +528,9 @@ class Agent:
                         openrouter_model_kwargs["extra_body"] = {"reasoning": {"effort": "high"}}
                         logging.info(f"Enabled reasoning effort=high for {openrouter_model_name} via OpenRouter")
                 elif detected_provider == "google":
+                    # OpenRouter path: exclude reasoning tokens from response for
+                    # user-facing chats. Direct Google SDK uses a different mechanism
+                    # (structured thinking blocks filtered by include_thinking=False).
                     reasoning_cfg = {"effort": "high"}
                     if not is_background:
                         reasoning_cfg["exclude"] = True
--- a/server/chat/backend/agent/agent.py
+++ b/server/chat/backend/agent/agent.py
@@ -528,6 +528,8 @@ class Agent:
                         openrouter_model_kwargs["extra_body"] = {"reasoning": {"effort": "high"}}
                         logging.info(f"Enabled reasoning effort=high for {openrouter_model_name} via OpenRouter")
                 elif detected_provider == "google":
+                    # OpenRouter path only. Direct Google SDK uses structured
+                    # thinking blocks filtered by include_thinking=False downstream.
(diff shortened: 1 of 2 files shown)


server/chat/backend/agent/agent.py
```diff
@@ -484,15 +506,19 @@ def on_llm_end(self, response, run_id=None, **kwargs):
...
                         openrouter_model_kwargs["extra_body"] = {"reasoning": {"effort": "high"}}
                         logging.info(f"Enabled reasoning effort=high for {openrouter_model_name} via OpenRouter")
                 elif detected_provider == "google":
-                    openrouter_model_kwargs["extra_body"] = {"reasoning": {"effort": "high"}}
-                    logging.info(f"Enabled reasoning effort=high for {openrouter_model_name} via OpenRouter")
+                    reasoning_cfg = {"effort": "high"}
+                    if not is_background:
+                        reasoning_cfg["exclude"] = True
+                    openrouter_model_kwargs["extra_body"] = {"reasoning": reasoning_cfg}
+                    logging.info(f"Enabled reasoning effort=high for {openrouter_model_name} via OpenRouter (exclude={not is_background})")
```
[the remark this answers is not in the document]
damianloch: Intentional — we want high reasoning effort for output quality even in user-facing chats. The model produces better answers with deeper reasoning regardless of whether the reasoning tokens are surfaced to the user. The cost trade-off is acceptable since it directly improves response quality. If cost becomes a concern we can revisit, but for now quality is the priority.

server/chat/backend/agent/agent.py
```diff
@@ -21,25 +21,61 @@
...
+        if not choices:
+            return result
+
+        delta = choices[0].get("delta") or {}
+        reasoning, has_reasoning_details = _extract_reasoning_from_delta(delta)
+
+        if reasoning and hasattr(result.message, "additional_kwargs"):
+            result.message.additional_kwargs["reasoning_content"] = reasoning
+
+        if (reasoning or has_reasoning_details) and not delta.get("content"):
```
beng360, line 76: This clears content for the streaming path, but `_convert_to_ui_messages()` in workflow.py (~line 1549) still calls `_extract_text_from_content` with `include_thinking=True` as a fallback for tool-call messages. For direct-mode Google (not OpenRouter), thinking blocks survive into the persisted `chat_sessions.messages` and show up on page reload even though they're stripped during live streaming.

server/chat/backend/agent/agent.py
```diff
@@ -484,15 +520,19 @@ def on_llm_end(self, response, run_id=None, **kwargs):
...
                     native = model_name.split("/", 1)[-1] if "/" in model_name else model_name
                     if OpenAIProvider._supports_reasoning(native):
                         openrouter_model_kwargs["extra_body"] = {"reasoning": {"effort": "high"}}
                         logging.info(f"Enabled reasoning effort=high for {openrouter_model_name} via OpenRouter")
                 elif detected_provider == "google":
-                    openrouter_model_kwargs["extra_body"] = {"reasoning": {"effort": "high"}}
-                    logging.info(f"Enabled reasoning effort=high for {openrouter_model_name} via OpenRouter")
+                    reasoning_cfg = {"effort": "high"}
+                    if not is_background:
+                        reasoning_cfg["exclude"] = True
```
beng360, line 535: This only covers the OpenRouter path. Direct Google SDK (vertex/native) doesn't get the exclude flag — reasoning comes back as content blocks with `type: "thinking"` instead and relies on `include_thinking=False` downstream. Two completely different reasoning formats handled by different mechanisms, worth a comment so nobody assumes this is the only defense.

server/chat/backend/agent/workflow.py
```diff
@@ -1074,17 +1074,24 @@ async def stream(self, input_state: State):
...
+                            reasoning = chunk_obj.additional_kwargs.get("reasoning_content", "")
+
+                        # Extract visible content. _ReasoningChatOpenAI clears
+                        # chunk_obj.content for reasoning-only chunks upstream, so
+                        # this is safe to call unconditionally.
                         if hasattr(chunk_obj, 'content') and chunk_obj.content:
-                            # Extract text content (handles Gemini thinking model list format)
-                            # include_thinking=True so reasoning flows to save_incident_thought() for RCA
-                            content = _extract_text_from_content(chunk_obj.content, include_thinking=True)
+                            content = _extract_text_from_content(chunk_obj.content, include_thinking=False)
```
beng360: This fixes the leak for all providers, not just Google — Anthropic thinking tokens were also streaming through before. Side effect: if a reasoning model (DeepSeek-R1 etc.) returns only reasoning with no content in a foreground chunk, the user now sees nothing for that chunk. Probably fine since those models generally produce both, but worth knowing about.

server/chat/backend/agent/agent.py
damianloch: Fixed in 2fa514ad. Both `_consolidate_message_chunks` (line 1367) and `_convert_to_ui_messages` (line 1549) now use `include_thinking=False`. Thinking blocks no longer survive into persisted messages.

server/chat/backend/agent/agent.py
damianloch: Added a clarifying comment in 2fa514ad. The exclude flag is OpenRouter-only; direct Google SDK relies on structured thinking blocks filtered by `include_thinking=False` downstream.

server/chat/backend/agent/workflow.py
damianloch: Acknowledged. Reasoning-only chunks in foreground producing no visible output is correct behavior — the actual answer arrives in subsequent content chunks. DeepSeek-R1 reasoning flows through `additional_kwargs["reasoning_content"]` which is intentionally suppressed in foreground.

beng360 (approved):

server/chat/backend/agent/agent.py
```diff
@@ -21,25 +21,61 @@
...
+    if has_reasoning_details:
+        parts = [
+            detail.get("text", "")
+            for detail in reasoning_details
+            if isinstance(detail, dict) and detail.get("text")
+        ]
+        if parts:
+            reasoning = (reasoning + "".join(parts)) if reasoning else "".join(parts)
+
+    return reasoning, has_reasoning_details
```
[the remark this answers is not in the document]
damianloch: Skipping — no test infrastructure exists yet per team guidelines. Noted as a good follow-up for when we add a test suite for the agent layer.

server/chat/backend/agent/workflow.py
```diff
@@ -1074,17 +1074,24 @@ async def stream(self, input_state: State):
...
-                            content = _extract_text_from_content(chunk_obj.content, include_thinking=True)
+                            content = _extract_text_from_content(chunk_obj.content, include_thinking=is_background)
 
-                        # Check for reasoning content (OpenRouter reasoning, DeepSeek-R1 etc.)
-                        # regardless of whether content was found above
-                        if not content and hasattr(chunk_obj, 'additional_kwargs'):
-                            content = chunk_obj.additional_kwargs.get("reasoning_content", "")
+                        # For background RCA chats, reasoning feeds into incident thoughts
+                        if not content and reasoning and is_background:
+                            content = reasoning
```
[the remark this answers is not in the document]
damianloch: Acknowledged. This preserves pre-existing behavior — before this PR, `include_thinking=True` was unconditional for all chats. The suggested refactor to split visible text from thinking is a valid improvement but out of scope for this bug fix. Filed as a follow-up.

PR #389, opened 2026-05-12
feat: Sentry connector: webhook ingestion + RCA tools

Zarlanx:
## Summary

Adds a Sentry connector for issue/error webhook ingestion and automated RCA. Internal Integration auth-token flow, HMAC-signed webhook receiver, agent tools for issue/event/breadcrumb retrieval, and skill registration.

- **Backend**: `routes/sentry/` (Flask blueprint + Celery task), `connectors/sentry_connector/` (read-only API client), `chat/backend/agent/tools/sentry_tool.py` (issue search, event/stacktrace fetch, Discover query), skill manifest at `agent/skills/integrations/sentry/SKILL.md`, registration in `connector_status` PROVIDER_CHECKERS, RCA prompt builder hook.
- **Frontend**: `/sentry/auth` two-step wizard (token entry → webhook verification), proxy routes under `/api/sentry/*`, connector registry entry, official Sentry mark from simple-icons.
- **Security**: HMAC-SHA256 signature verification on every webhook against the stored client_secret (rejects unsigned/invalid). Vault-backed credential storage via `store_tokens_in_db` / `get_token_data`. Read-only scopes only (`event:read`, `org:read`, `project:read`).

The final commit on this branch (\`fix(sentry): connector end-to-end after live webhook test\`) lands six issues found while running a real \`issue.created\` webhook through ngrok end-to-end — see commit message for details.

## Test plan

- [x] Connect via UI: paste auth token + webhook secret → org validated against Sentry API → credentials stored in Vault → status flips to Connected
- [x] Webhook URL surfaced in Step 1 (before connecting) so the user can configure the Sentry Internal Integration up front
- [x] Real Sentry \`issue.created\` webhook → ngrok → Aurora → 202 accepted
- [x] HMAC signature verified; tampering rejected with 401
- [x] Celery task \`sentry.process_event\` registered and dispatched
- [x] Incident row persisted; \`alert_metadata\` populated; \`source_alert_id\` = sentry_events PK
- [x] Background RCA session triggered with Sentry context
- [x] Incident summary generated
- [ ] CodeRabbit review pass
- [ ] Test against EU region (\`de.sentry.io\`)
- [ ] Verify webhook delivery on \`error\` events from a Business/Enterprise plan

## Summary by CodeRabbit

* **New Features**
  * Added Sentry connector: connect/disconnect UI, webhook URL display/copy, and connection status.
  * Platform can list Sentry projects, issues, and events; ingest Sentry webhooks to create or correlate incidents.
  * Sentry data accessible to the AI assistant for incident investigation and automated RCA workflows.

* **Documentation**
  * Added comprehensive Sentry connector setup and troubleshooting guides.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Arvo-AI/aurora/pull/389)


client/src/app/sentry/auth/page.tsx
```diff
@@ -0,0 +1,293 @@
...
+    if (result?.connected) {
+      setRegion(result.region || "us");
+      if (result.orgSlug) setOrgSlug(result.orgSlug);
+      await loadWebhookUrl();
+    } else if (typeof window !== 'undefined') {
+      localStorage.removeItem(CACHE_KEYS.STATUS);
+    }
+  };
+
+  const loadStatus = async (skipCache = false) => {
```
damianloch, line 74: The loadStatus function reads isInitialLoad state, but it's only called from the useEffect on mount. The isInitialLoad closure captures the initial value (true), so the branch at line ~92 (if (isInitialLoad)) always evaluates true on the first and only call. Then it sets isInitialLoad to false — but that state update is never observed because loadStatus is never called again. The whole isInitialLoad mechanism is effectively dead code that adds complexity without providing a real "stale-while-revalidate" benefit across multiple calls.

client/src/components/sentry/SentryConnectionStep.tsx
```diff
@@ -0,0 +1,194 @@
...
+  copied: boolean;
+  onCopyWebhook: () => void;
+}
+
+const REGION_HINTS = [
+  { value: "us", label: "US (sentry.io)" },
+  { value: "eu", label: "EU (de.sentry.io)" },
+];
+
+const SENTRY_PURPLE = "#362d59";
```
damianloch: This constant is also defined in page.tsx and SentryWebhookStep.tsx. If the brand color ever changes you'd need to update three files. Make a const to import

server/chat/backend/agent/skills/integrations/sentry/SKILL.md
```diff
@@ -0,0 +1,92 @@
+---
+name: sentry
+id: sentry
+description: "Sentry error tracking integration for searching issues, fetching stacktraces, listing projects, and running Discover event searches during RCA investigations"
+category: observability
+connection_check:
```
damianloch, line 6: The frontmatter declares a connection check method, but the skill was never registered in server/chat/backend/agent/skills/registry.py (not in this diff). Without the registry entry, the agent won't load this skill during RCA regardless of what's in the SKILL.md. Per the connector checklist this is required.

server/routes/connector_status.py
```diff
@@ -701,9 +729,10 @@ def _check_netdata(creds: Dict[str, Any]) -> Dict[str, Any]:
     "dynatrace": _check_dynatrace,
     "bigpanda": _check_bigpanda,
     "tailscale": _check_tailscale,
+    "sentry":        _check_sentry,
     # Credential-existence checks (no live API endpoint to validate against)
-    "netdata": _check_netdata,
-    "newrelic": _check_newrelic,
+    "netdata":       _check_netdata,
+    "newrelic":      _check_newrelic,
```
isiddharthsingh: remove the whitespaces

client/src/app/api/sentry/connect/route.ts
damianloch, line 1: File: All 6 route files

Every other mature connector in the codebase (incident-io, etc.) uses forwardRequest or forwardAuthenticatedGet which automatically attaches the internal API secret header (X-Internal-Secret) to backend calls. These sentry routes instead manually call getAuthenticatedUser() + raw fetch(), which means:

The X-Internal-Secret header is never sent — if the backend enforces it on these endpoints, requests will be rejected.

Error handling is inconsistent with the centralized proxy (which normalizes timeouts, retries, and error shapes).

More code to maintain per route.

server/routes/sentry/tasks.py
```diff
@@ -0,0 +1,420 @@
...
+                            logger.info(
+                                "[SENTRY][WEBHOOK] Triggered background RCA for session %s (task_id=%s)",
+                                session_id, task.id,
+                            )
+                    except Exception as e:
+                        logger.error("[SENTRY][WEBHOOK] Failed to trigger RCA: %s", e)
+
+    except Exception as exc:
+        logger.exception("[SENTRY][WEBHOOK] Failed to process webhook payload")
+        raise self.retry(exc=exc)
```
isiddharthsingh: The current except Exception as exc block indiscriminately retries all errors; change it to only retry for transient errors by replacing the broad except with logic that detects transient exceptions (e.g., DB OperationalError, connection/timeout errors, requests exceptions) and only calls self.retry(exc=exc) for those types; for non-transient errors (validation errors, integrity errors, etc.) log with logger.exception("[SENTRY][WEBHOOK] Failed to process webhook payload") and re-raise immediately without retrying. Locate the generic except block around the webhook processing (the except Exception as exc handler that calls self.retry) and implement an isinstance-based or helper function is_transient_error(exc) to decide when to call self.retry versus re-raising. Ensure self.retry remains used for transient cases and non-transient cases propagate the original exception.

client/src/app/api/sentry/connect/route.ts
```diff
@@ -0,0 +1,31 @@
...
+    const authResult = await getAuthenticatedUser();
+    if (authResult instanceof NextResponse) return authResult;
+    const { headers: authHeaders } = authResult;
+    const payload = await request.json();
+
+    const response = await fetch(`${API_BASE_URL}/sentry/connect`, {
+      method: 'POST',
+      headers: { ...authHeaders, 'Content-Type': 'application/json' },
+      body: JSON.stringify(payload),
+      credentials: 'include',
```
damianloch: credentials: 'include' sends browser cookies, but this is a Next.js server-side API route making a server-to-server call to Flask. There are no browser cookies in this context. This is dead code repeated across all 6 files. forwardRequest handles auth correctly without this.

client/src/app/api/sentry/connect/route.ts
```diff
@@ -0,0 +1,31 @@
...
+    });
+
+    if (!response.ok) {
+      const text = await response.text();
+      return NextResponse.json({ error: text || 'Failed to connect Sentry' }, { status: response.status });
+    }
```
damianloch: If the backend returns a JSON error body (which it does — {"error": "..."}), this wraps it as a string inside another JSON object: {"error": "{\"error\": \"...\"}"}. The frontend then has to double-parse. forwardRequest handles this correctly by streaming the backend response as-is.

In all 6 route.ts files

client/src/components/sentry/SentryConnectionStep.tsx
isiddharthsingh: Consider renaming constants to follow camelCase convention.

client/src/app/sentry/auth/page.tsx
```diff
@@ -0,0 +1,293 @@
...
+    updateLocalStorageConnection(result?.connected ?? false);
+
+    if (result?.connected) {
+      setRegion(result.region || "us");
+      if (result.orgSlug) setOrgSlug(result.orgSlug);
+      await loadWebhookUrl();
+    } else if (typeof window !== 'undefined') {
+      localStorage.removeItem(CACHE_KEYS.STATUS);
+    }
+  };
```
isiddharthsingh, line 72: Replace the manual fetchAndUpdateStatus flow with a query-client driven query that uses revalidateOnEvents:['sentryStateChanged'] from the query utility (import from "@/lib/query.ts") instead of calling sentryService.getStatus manually and managing useState/useEffect/localStorage synchronization; specifically remove the manual fetchAndUpdateStatus logic that sets status, updates CACHE_KEYS.STATUS in localStorage, calls updateLocalStorageConnection, sets region via setRegion, sets orgSlug via setOrgSlug, and calls loadWebhookUrl, and instead: create a query keyed for Sentry status that invokes sentryService.getStatus, configure revalidateOnEvents:['sentryStateChanged'], and move any localStorage/cache writes or side-effects into the query's onSuccess or listeners so the component consumes the query result directly and no longer manipulates CACHE_KEYS.STATUS or calls updateLocalStorageConnection manually.

client/src/app/sentry/auth/page.tsx
```diff
@@ -0,0 +1,293 @@
...
+      console.error('[sentry] Failed to load status', error);
+      toast({ title: 'Error', description: 'Unable to load Sentry status', variant: 'destructive' });
+    }
+  };
+
+  useEffect(() => {
+    loadStatus();
+    loadWebhookUrl();
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
isiddharthsingh, line 106: Consider stabilizing effect dependencies or documenting the intent.

The useEffect calls loadStatus() and loadWebhookUrl() but disables the exhaustive-deps lint rule. While this works for a mount-only effect, consider either:

1. Wrapping these functions with useCallback to stabilize them and include them in the dependency array
2. Adding a comment explaining why the effect should only run once on mount
This would make the intent clearer and prevent potential issues if the functions are later modified to depend on props or state.

server/routes/sentry/sentry_routes.py
```diff
@@ -0,0 +1,479 @@
...
+
+    metadata = {
+        "resource": resource,
+        "request_id": request_id,
+        "timestamp": request.headers.get("Sentry-Hook-Timestamp", ""),
+        "remote_addr": request.remote_addr,
+    }
+
+    from routes.sentry.tasks import process_sentry_event, extract_sentry_title
+    title = extract_sentry_title(payload, resource)
```
isiddharthsingh: Consider moving import to module level.

server/routes/sentry/sentry_routes.py
```diff
@@ -0,0 +1,479 @@
...
+    """List Sentry webhook events stored in the database."""
+    org_id = get_org_id_from_request()
+    limit = request.args.get("limit", default=50, type=int)
+    offset = request.args.get("offset", default=0, type=int)
+    resource_filter = request.args.get("resource")
+
+    try:
+        with db_pool.get_admin_connection() as conn:
+            cursor = conn.cursor()
+            set_rls_context(cursor, conn, user_id, log_prefix="[Sentry]")
```
isiddharthsingh: The cursor in the db_pool.get_admin_connection() block is created without a context manager; change the code to use a context manager for the cursor (i.e., use "with conn.cursor() as cursor:") so the cursor is closed automatically, then call set_rls_context(cursor, conn, user_id, log_prefix="[Sentry]") inside that with-block; locate the block with db_pool.get_admin_connection() and the set_rls_context call in sentry_routes.py and adjust indentation accordingly.

damianloch (changes requested):

client/src/app/api/sentry/webhook-url/route.ts
```diff
@@ -0,0 +1,30 @@
...
+    const authResult = await getAuthenticatedUser();
+    if (authResult instanceof NextResponse) return authResult;
+    const { headers: authHeaders } = authResult;
+
+    const response = await fetch(`${API_BASE_URL}/sentry/webhook-url`, {
+      method: 'GET',
+      headers: authHeaders,
+      credentials: 'include',
+      cache: 'no-store',
+    });
```
isiddharthsingh: Use forwardRequest from @/lib/backend-proxy, instead of direct fetch to Flask backend.

client/src/app/api/sentry/projects/route.ts
```diff
@@ -0,0 +1,30 @@
...
+    const authResult = await getAuthenticatedUser();
+    if (authResult instanceof NextResponse) return authResult;
+    const { headers: authHeaders } = authResult;
+
+    const response = await fetch(`${API_BASE_URL}/sentry/projects`, {
+      method: 'GET',
+      headers: authHeaders,
+      credentials: 'include',
+      cache: 'no-store',
+    });
```
isiddharthsingh: Use forwardRequest from @/lib/backend-proxy, instead of direct fetch to Flask backend.

client/src/app/api/sentry/issues/route.ts
```diff
@@ -0,0 +1,31 @@
...
+    if (authResult instanceof NextResponse) return authResult;
+    const { headers: authHeaders } = authResult;
+    const search = request.nextUrl.search || '';
+
+    const response = await fetch(`${API_BASE_URL}/sentry/issues${search}`, {
+      method: 'GET',
+      headers: authHeaders,
+      credentials: 'include',
+      cache: 'no-store',
+    });
```
isiddharthsingh: Use forwardRequest from @/lib/backend-proxy, instead of direct fetch to Flask backend.

client/src/app/api/sentry/events/ingested/route.ts
```diff
@@ -0,0 +1,31 @@
...
+    if (authResult instanceof NextResponse) return authResult;
+    const { headers: authHeaders } = authResult;
+    const search = request.nextUrl.search || '';
+
+    const response = await fetch(`${API_BASE_URL}/sentry/events/ingested${search}`, {
+      method: 'GET',
+      headers: authHeaders,
+      credentials: 'include',
+      cache: 'no-store',
+    });
```
isiddharthsingh: Use forwardRequest from @/lib/backend-proxy, instead of direct fetch to Flask backend.

client/src/lib/services/sentry.ts
```diff
@@ -0,0 +1,84 @@
...
+  },
+
+  async connect(payload: SentryConnectPayload): Promise<SentryStatus> {
+    const data = await apiRequest<UnknownRecord>(`${API_BASE}/connect`, {
+      method: 'POST',
+      body: JSON.stringify(payload),
+      cache: 'no-store',
+    });
+    return {
+      connected: Boolean(data?.success ?? true),
```
isiddharthsingh: Unsafe default: connected defaults to true when success field is missing.

If the backend response doesn't include a success field, the code defaults connected to true, which could mask connection failures or malformed responses. Consider defaulting to false or explicitly validating that a successful response contains the expected fields.

client/src/lib/services/sentry.ts
```diff
@@ -0,0 +1,84 @@
+'use client';
+
+import { apiRequest } from '@/lib/services/api-client';
```
isiddharthsingh, line 3: Use forwardRequest from @/lib/backend-proxy instead of apiRequest.

All client-to-backend HTTP requests must go through the Next.js API route proxy using forwardRequest from @/lib/backend-proxy. The current implementation uses apiRequest from @/lib/services/api-client, which violates the coding guidelines. As per coding guidelines: "All client-to-backend HTTP requests must go through Next.js API route proxy using forwardRequest from @/lib/backend-proxy; never call Flask backend directly from browser."

Then update all method calls accordingly (lines 43, 63, 80).

server/connectors/sentry_connector/README.md
```diff
@@ -0,0 +1,72 @@
...
+   - **Webhook URL**: paste the Aurora webhook URL shown after connecting (e.g. `https://your-aurora-domain/sentry/webhook/{user_id}`)
+   - **Permissions** (minimum, all read-only):
+     - Issue & Event: `Read`
+     - Project: `Read`
+     - Organization: `Read`
+     - Member: `Read` (optional, for richer context)
+   - **Webhooks** subscriptions:
+     - `issue` (created / resolved / assigned)
+     - `error` (created) — *Business/Enterprise plans only*
+     - `event_alert` (triggered)
```
isiddharthsingh: i dont see this on sentry

damianloch (dismissed):

damianloch (approved):

PR #395, opened 2026-05-13
MCP: token-lean hybrid tool surface with per-user connector gating

isiddharthsingh:
## Summary
- Replaces the 6-tool MCP surface with a tiered design: 9 always-on Tier-1, 6 connector-gated Tier-2, and 53 long-tail entries reachable via `search_tools`/`call_tool`.
- Per-user gating via `SkillRegistry.check_connection` — Tier-2 tools appear only when an enabling connector is connected.
- New `chat_with_aurora` tool routes external clients through Aurora's own agent (prompts, skills, RCA pipeline, citations) instead of having the external LLM drive raw API calls.
- Every path verified against the live Flask url_map; infra-write surfaces (Terraform apply, kubectl mutations, shell exec, Cloudflare WAF) excluded by a hard-coded allowlist with import-time assertions.

## Token surface (measured)
Tier-1 only: 1,365 tok · Tier-1+Tier-2: 2,113 tok · `search_tools` response: 600 tok.

## Test plan
- [ ] Generate an MCP token via Aurora UI → Org Settings → MCP API Tokens.
- [ ] Hook into Claude Desktop / Cursor / Windsurf using the provided JSON config.
- [ ] Verify `tools/list` returns ~9 Tier-1 tools with no connectors; reconnect Datadog/Jira/etc and confirm matching Tier-2 tools appear on the next request.
- [ ] `chat_with_aurora("why did <svc> page?")` returns a real RCA-shaped answer with citations; the session appears in Aurora's chat UI.
- [ ] Direct tool calls (`query_logs`, `query_jira`, etc.) return real data when the connector is connected, `not_connected` envelope otherwise.
- [ ] `call_tool("terraform_apply", {...})` is refused with `tool_not_allowlisted`.

## Summary by CodeRabbit

* **New Features**
  * Tiered MCP tool surface (always‑on, connector‑gated, searchable dispatch) with unified MCP server, prompts, and resource endpoints.
  * Chat bridge with polling/resume, message-level endpoints, and citation extraction.
  * New incident flows: free‑text RCA trigger API and guided incident prompts; connector‑backed tools for logs, metrics, alerts, Jira, Notion, Bitbucket.
  * Response size controls and pagination for tool outputs.
* **Bug Fixes**
  * Safer RCA error messages; deduplication of immediate user messages and streaming‑placeholder handling.
* **Documentation**
  * Updated MCP guide with tier model and allowlist guidance.
* **Tests**
  * Comprehensive MCP test coverage added (dispatch, gating, registry, bridge, response).


server/routes/chat_routes.py
```diff
@@ -491,6 +490,178 @@ def delete_chat_session(user_id, session_id):
...
+        for m in reversed(messages):
+            if m.get('sender') == 'aurora' and m.get('citations'):
+                citations = m['citations']
+                break
+
+        return jsonify({
+            'session_id': session_id,
+            'status': status,
+            'seq': total,
+            'messages': new_msgs,
```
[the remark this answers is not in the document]
isiddharthsingh: No functional bug — current scheme works end-to-end (smoke test passes, citations & chat session creation OK). Only caller is chat_bridge.py (verified by grep — no frontend usage). The proposal is a contract change for a stylistic inconsistency between POST (returns length-1) and GET (returns length); applying it requires coordinated changes across chat_routes.py, chat_bridge.py polling loop, and the max(0, after) clamp I added earlier (which would have to become max(-1, after)). Risk of regression outweighs the cosmetic gain.

server/aurora_mcp/registry.py
```diff
@@ -0,0 +1,770 @@
...
+    DispatchEntry(
+        name="github_list_branches",
+        description="List branches for a GitHub repository.",
+        category="code",
+        method="GET",
+        path="/github/user-branches/{repo_full_name}",
+        enabling_skills=("github",),
+        path_args=("repo_full_name",),
+    ),
+)
```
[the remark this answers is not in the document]
isiddharthsingh: Verified the concern is real (allowlist could drift from backend routes) but the proposed location is wrong — registry.py runs in the MCP container which deliberately does not import the Flask app. Adding from main_compute import app there would pull DB pools, Casbin init, rate limiter, and the MCP preloader into a process that exists specifically to avoid them.

Confirmed there's no current drift: ran app.url_map.bind("").match(path, method) from inside aurora-server against all 53 DISPATCH_ALLOWLIST entries with feature flags forced on → 0 mismatches.

Existing coverage: scripts/_live_check.sh exercises every dispatch entry with a connected provider, and commit a2b5f34a now logs every upstream non-2xx server-side, so any future drift surfaces immediately.

Will track the proper fix as a separate PR: a server/tests/architectural/test_dispatch_allowlist.py that imports main_compute with NEXT_PUBLIC_ENABLE_* flags forced on via conftest and validates every entry against app.url_map. ~30 LOC, but adds DB-dependent test setup which is out of scope for this triage.

server/aurora_mcp/tools_gated.py
```diff
@@ -0,0 +1,360 @@
...
+                body={"jql": jql, "maxResults": _clamp_limit(max_results)},
+            ),
+            tool_name="query_jira",
+        )
+    if action == "get_issue":
+        if not issue_key:
+            return {"error": "issue_key_required"}
+        return truncate_payload(
+            await api_call("GET", f"/jira/issue/{issue_key}"),
+            tool_name="query_jira",
```
beng360, line 209: issue_key, db_id, workspace, repo_slug are interpolated into URL paths without encoding. httpx resolves ../ segments, so ../../admin/secrets escapes the intended prefix. dispatch.py:65 already does quote(str(value), safe="") — apply the same here.

server/aurora_mcp/chat_bridge.py
```diff
@@ -0,0 +1,191 @@
...
+    interval = _POLL_INTERVAL_INITIAL
+    latest_partial: Optional[str] = None
+
+    while elapsed < _POLL_TOTAL_SECONDS:
+        await asyncio.sleep(interval)  # NOSONAR S7484: cross-process HTTP poll, no in-process signal to wait on.
+        elapsed += interval
+        interval = min(interval * 2, _POLL_INTERVAL_MAX)
+
+        page, msgs, status, last_seq = await _poll_once(api_call, sid, last_seq)
+        latest_partial = _latest_assistant_text(msgs, latest_partial)
```
beng360: _poll_once() here is outside the try/except that wraps earlier polls. If it raises TimeoutError, the caller never gets back the session_id and cannot retry. Return {status: "in_progress", session_id: sid} on timeout instead.

server/aurora_mcp/registry.py
```diff
@@ -0,0 +1,770 @@
...
+
+
+# ---------------------------------------------------------------------------
+# Hard-banned tool name fragments — defense-in-depth against accidental
+# inclusion of infra-write surfaces in DISPATCH_ALLOWLIST.
+# ---------------------------------------------------------------------------
+
+_BANNED_NAME_FRAGMENTS: Tuple[str, ...] = (
+    "terraform",
```
beng360, line 747: Ban list misses kubectl_scale, kubectl_patch, kubectl_rollout, shell_exec, shell_run. Consider banning the kubectl_ prefix entirely (except kubectl_get) instead of enumerating verbs.

beng360 (commented):
MCP review — code issues + live testing findings (8 inline comments)

server/aurora_mcp/registry.py
```diff
@@ -0,0 +1,770 @@
...
+        for frag in _BANNED_NAME_FRAGMENTS:
+            if frag in lname:
+                raise RuntimeError(
+                    f"DISPATCH_ALLOWLIST contains banned tool '{entry.name}' "
+                    f"(matched fragment '{frag}'). Infra-write tools must not "
+                    f"be reachable from MCP. See server/aurora_mcp/registry.py."
+                )
+        # Reject obviously dangerous methods + paths combinations.
+        if entry.method.upper() == "DELETE" and "/apply" in entry.path:
+            raise RuntimeError(
```
beng360: This checks DELETE + /apply. No REST API uses DELETE on an /apply endpoint — this assertion can never fire. Either check for POST/PUT on banned paths or remove it.

server/mcp_server.py
```diff
@@ -66,8 +70,13 @@ async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
...
 
-_pool: psycopg2.pool.ThreadedConnectionPool | None = None
-_last_used_cache: dict[str, float] = {}
+_pool: Optional[psycopg2.pool.ThreadedConnectionPool] = None
+_last_used_cache: Dict[str, float] = {}
+
+# token → (resolved_at_monotonic, user_id, org_id). Avoids a Postgres SELECT
+# on every list_tools / call_tool request.
+_TOKEN_READ_TTL = 60.0
+_token_read_cache: Dict[str, tuple] = {}
```
beng360, line 81: _token_read_cache and _last_used_cache are unbounded dicts. The 60s TTL controls re-reads, not eviction. Rotating tokens leak memory. Use cachetools.TTLCache or functools.lru_cache.

server/aurora_mcp/chat_bridge.py
```diff
@@ -0,0 +1,191 @@
...
+            last_seq = posted_seq
+
+    elapsed = 0.0
+    interval = _POLL_INTERVAL_INITIAL
+    latest_partial: Optional[str] = None
+
+    while elapsed < _POLL_TOTAL_SECONDS:
+        await asyncio.sleep(interval)  # NOSONAR S7484: cross-process HTTP poll, no in-process signal to wait on.
+        elapsed += interval
+        interval = min(interval * 2, _POLL_INTERVAL_MAX)
```
beng360: elapsed accumulates sleep durations but ignores HTTP request time. Each _poll_once can take up to 15s. Real wall time can far exceed the 45s budget. Use time.monotonic().

server/aurora_mcp/resources.py
```diff
@@ -0,0 +1,144 @@
...
+    except Exception:
+        logger.exception("catalog/skills failed")
+        return dict(_INTERNAL_ERROR)
+
+
+def _slim_incident_row(i: Dict[str, Any]) -> Dict[str, Any]:
+    return {
+        "id": i.get("id"),
+        "title": i.get("alert_title") or i.get("title"),
+        "status": i.get("aurora_status") or i.get("status"),
```
beng360: _slim_incident_row reads alert_title/alert_service but the API nests these under alert.title/alert.service_name. All 20 incidents come back with null title, service, and created_at. Needs to traverse the nested dict.

server/aurora_mcp/response.py
```diff
@@ -0,0 +1,80 @@
...
+def paginate(
+    items: Iterable[Any],
+    *,
+    cursor: Optional[str] = None,
+    limit: int = 20,
+    max_limit: int = 100,
+) -> Tuple[List[Any], Optional[str]]:
+    """Slice `items` by an integer-offset cursor and return (page, next_cursor)."""
+    limit = max(1, min(int(limit or 20), max_limit))
+    try:
```
beng360, line 62: max(1, ...) clamps limit=0 to 1, so callers get 1 result when they asked for 0. Change to max(0, ...) or return early.

server/aurora_mcp/registry.py
```diff
@@ -0,0 +1,770 @@
...
+
+def _entry_matches_search(
+    entry: "DispatchEntry", q: str, cat: str, conn: str, user_id: Optional[str],
+) -> bool:
+    if cat and entry.category.lower() != cat:
+        return False
+    if conn and conn not in {s.lower() for s in entry.enabling_skills}:
+        return False
+    if q and q not in entry.name.lower() and q not in entry.description.lower():
+        return False
```
beng360: Search is substring-only. "incident update" (space) doesn't match incident_update (underscore), so legitimate queries return empty. Normalize underscores to spaces before matching, or split into tokens.

server/aurora_mcp/chat_bridge.py
```diff
@@ -0,0 +1,191 @@
...
+        api_call: bound `_api` proxy from mcp_server.py (forwards user identity).
+        message: user message text. Ignored when poll_only=True.
+        session_id: continue an existing session, or None to start a new one.
+        mode: "chat" (default) or "rca" — passed to the backend agent.
+        poll_only: when True, skip create+post and just poll session_id for new
+            assistant messages. Use to resume a still-running session without
+            sending a new turn.
+    """
+    err = _validate_inputs(message, session_id, mode, poll_only)
+    if err is not None:
```
beng360, line 236: Validation errors return {error: "..."} without session_id or status. Invalid session_id throws a 404 exception. Docstring says those fields are always present. Three different error shapes — pick one.

server/aurora_mcp/tools_gated.py
```diff
@@ -0,0 +1,360 @@
...
+integration. Combined with per-request visibility filtering (in
+mcp_server.py), this implements the "rebuild on every request" gating
+model called out in the design.
+
+Every path here has been verified against the Flask url_map — only real
+endpoints are reachable. Tools whose backing connectors expose no REST
+data-query routes (Coroot, ThousandEyes, distributed tracing across all
+connectors, GitHub `rca`) have been dropped — the agent's Python tools
+for those are reachable through `chat_with_aurora` instead.
+"""
```
beng360, line 16: Sentry, Grafana, Slack, Jenkins, Spinnaker, Cloudflare reads are all absent from every MCP tier. Docstring says they're "reachable through chat_with_aurora" but clients can't discover that. Surface them in search_tools results with a hint at minimum.

server/mcp_server.py
```diff
@@ -140,28 +153,86 @@ def _get_token() -> str:
...
+    *,
+    params: Optional[Dict[str, Any]] = None,
+    body: Optional[Dict[str, Any]] = None,
+) -> Dict[str, Any]:
+    """Proxy a request to the Aurora Flask API with identity from the MCP token.
+
+    Default request timeout is 30s (configured on the shared httpx client).
+    Callers that need a different deadline wrap with `async with asyncio.timeout(N)`.
+    """
     if not path.startswith("/"):
```
beng360: path.startswith("/") doesn't prevent traversal. A path like "/../admin/secrets" passes the check and httpx resolves it. Validate that the resolved URL stays under API_BASE.

beng360 (commented):
Additional findings from live MCP testing:

1. No access to the service graph (`/graph/` routes). The `blast_radius_analysis` prompt references this but there's no tool to back it.
2. No aggregate LLM usage/cost endpoint. Per-incident `tokenUsage` works but there's no org-level summary.
3. DORA metrics (MTTR, MTTD, incident frequency) exist at `/metrics/` but aren't exposed.
4. Confluence only has fetch-by-URL in the dispatch — no search. The agent has CQL search internally.
5. No `whoami` resource to verify which user/org the token resolves to. Makes debugging "why can't I see X" harder than it needs to be.
6. AWS shows connected in the catalog but has zero tools in any MCP tier. Either add read-only dispatch entries or mark it agent-only in the catalog.

server/aurora_mcp/response.py
isiddharthsingh: i looked at the code and the symptom doesn't actually reproduce. int(limit or 20) evaluates 0 or 20 as 20, so limit=0 gets you 20 results, not 1. whether limit=0 should return zero results is a separate design call, happy to add an early return if you want that, but the off-by-one as described isn't there.

server/aurora_mcp/chat_bridge.py
isiddharthsingh: fair point on the inconsistency. pre-session-creation failures genuinely don't have a sid to attach yet, so adding one would be a placeholder. could normalize everything to {error, session_id: null, status: "error"} but it touches every error path for what's mostly cosmetic. leaving for a follow-up unless you feel strongly.

server/aurora_mcp/tools_gated.py
isiddharthsingh: they dont have much of tool calls of theier own, so i have kept them to be used directly from chat with aurora tool.

beng360 (approved):

PR #396, opened 2026-05-13
fix: resolve org-scoped credentials for all connector token lookups

Harrio-6:
get_user_token_data() only matched rows by user_id, so org members other than the original connector saw "connected" in the UI but got auth failures at runtime. Introduces _org_read_predicate() — a single helper that emits WHERE (user_id = %s OR org_id = %s) — and applies it consistently across all credential reads, writes, deletes, and GitHub repo queries. Modifies tests pinning the SQL shape, case normalization, and SQL-injection guards.

## Summary by CodeRabbit

* **New Features**
  * Organization-scoped access for shared credentials, repositories, manual VMs (responses now include isShared) and SSH keys; repo metadata/selection and webhook/service-account flows respect org-shared credentials.
* **Bug Fixes**
  * Corrected token owner resolution to prevent duplicate/ghost credential rows and improved org-aware preference handling.
* **Chore**
  * Database migration to enforce org-level uniqueness for provider credentials.
* **Tests**
  * Updated tests to cover org-scoped predicates and UUID validation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Arvo-AI/aurora/pull/396)

### DIFF
--- a/server/routes/github/github_repo_selection.py
+++ b/server/routes/github/github_repo_selection.py
@@ -5,6 +5,7 @@ Manages which repos a user has connected for RCA investigation.
 import logging
 import json
 from flask import Blueprint, jsonify, request
+from psycopg2 import sql as pgsql
 from utils.db.connection_pool import db_pool
 from utils.auth.rbac_decorators import require_permission
 from utils.auth.stateless_auth import set_rls_context
@@ -39,11 +40,11 @@ def get_repo_selections(user_id):
             with conn.cursor() as cur:
                 set_rls_context(cur, conn, user_id, log_prefix="[github_repo_selection:get_repo_selections]")
                 cur.execute(
-                    f"""SELECT DISTINCT ON (repo_full_name)
+                    pgsql.SQL("""SELECT DISTINCT ON (repo_full_name)
                               repo_full_name, repo_id, default_branch, is_private,
                               metadata_summary, metadata_status, repo_data, created_at
                        FROM github_connected_repos
-                       WHERE {predicate} ORDER BY repo_full_name, updated_at DESC""",
+                       WHERE {} ORDER BY repo_full_name, updated_at DESC""").format(predicate),
                     pred_params,
                 )
                 rows = cur.fetchall()
@@ -87,10 +88,10 @@ def save_repo_selections(user_id):
                 set_rls_context(cur, conn, user_id, log_prefix="[github_repo_selection:save_repo_selections]")
 
                 cur.execute(
-                    f"""SELECT DISTINCT ON (repo_full_name) repo_full_name, user_id
+                    pgsql.SQL("""SELECT DISTINCT ON (repo_full_name) repo_full_name, user_id
                         FROM github_connected_repos
-                        WHERE {predicate}
-                        ORDER BY repo_full_name, updated_at DESC""",
+                        WHERE {}
+                        ORDER BY repo_full_name, updated_at DESC""").format(predicate),
                     pred_params,
                 )
                 # {repo_full_name: owner_user_id} — we need the owner to delete the right row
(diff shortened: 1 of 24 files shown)


server/utils/secrets/secret_ref_utils.py
```diff
@@ -226,25 +248,23 @@ def get_user_token_data(self, user_id: str, provider: str) -> Optional[Dict[str,
...
                 f"""SELECT secret_ref, client_id, client_secret
                    FROM user_tokens
-                   WHERE user_id = %s
+                   WHERE {predicate}
                      AND provider = %s
                      AND secret_ref IS NOT NULL
                      AND is_active = TRUE
-                     {clause}
-                   ORDER BY CASE WHEN user_id = %s THEN 0 ELSE 1 END
                    LIMIT 1""",
```
beng360, line 223: The old `ORDER BY CASE WHEN user_id = %s THEN 0 ELSE 1 END` was dropped when you switched to `_org_read_predicate`, but the new predicate can match multiple users in the same org. There are real duplicates in prod — two Confluence rows in org `4fb825fc` owned by different users. Without an ORDER BY, `LIMIT 1` returns whichever row Postgres finds first, which is nondeterministic and can flip after a VACUUM or index rebuild.

Interestingly `get_token_owner_id` (added in this same PR) does have an `ORDER BY`, so the intent was clearly there. Should add something like `ORDER BY CASE WHEN user_id = %s THEN 0 ELSE 1 END` back here to prefer the caller's own row.

server/utils/secrets/secret_ref_utils.py
```diff
@@ -165,8 +188,8 @@ def update_user_token_with_secret_ref(self, user_id: str, provider: str, secret_
             set_rls_context(cursor, conn, user_id, log_prefix="[SecretRef:updateToken]")
             cursor.execute(
                 f"UPDATE user_tokens SET secret_ref = %s, is_active = TRUE "
-                f"WHERE user_id = %s AND provider = %s {clause}",
-                (secret_ref, user_id, provider) + params,
+                f"WHERE {predicate} AND provider = %s",
```
beng360, line 148: This UPDATE uses the org-wide predicate with no `user_id` narrowing and no LIMIT. When two users in the same org both have tokens for the same provider (which exists in prod — Confluence in org `4fb825fc`), this overwrites both rows' `secret_ref` to point at the same Vault path. The other user's original Vault secret becomes orphaned and their token row silently points to someone else's credential.

On main this was scoped with `WHERE user_id = %s AND provider = %s` so it only hit one row. Same issue applies to `migrate_token_to_secret_ref` (line ~351) and `_clear_secret_ref` (line ~386) which also use the broad predicate for UPDATEs.

server/routes/github/github_repo_selection.py
```diff
@@ -8,24 +8,12 @@
 from utils.db.connection_pool import db_pool
 from utils.auth.rbac_decorators import require_permission
 from utils.auth.stateless_auth import set_rls_context
+from utils.secrets.secret_ref_utils import _resolve_org, _org_read_predicate
```
beng360: `clear_repo_selections` (line 171) still uses `WHERE user_id = %s` — it wasn't updated to use `_org_read_predicate` like every other function in this file (`get_repo_selections`, `save_repo_selections`, `update_repo_metadata`, `trigger_metadata_generation`). In practice this means an org member who didn't originally connect the repos can hit Disconnect, get a 200 success back, but actually delete zero rows. Should use the same `_resolve_org` / `_org_read_predicate` pattern here.

beng360 (changes requested):
Three issues found — two are data-corruption bugs confirmed with live production data, one is a missed function.

beng360 (approved):

PR #402, opened 2026-05-14
Fix: clean up Memgraph discovery nodes on provider disconnect

Harrio-6:
Fixes stale infrastructure nodes persisting in Memgraph after a cloud provider is disconnected. Adds delete_services_for_provider, delete_services_for_cluster, and delete_services_for_aws_account methods to MemgraphClient and calls them from all disconnect endpoints (GCP, AWS, Azure, OVH, Scaleway, Tailscale, kubectl). Also adds a delete_stale_services safety net to the daily stale task that hard-deletes nodes flagged stale for 30+ days. Fixes a pre-existing bug where _query_connected_providers was missing the OR org_id = %s fallback, causing org-shared AWS connections to be silently skipped during discovery.

## Summary by CodeRabbit

* **New Features**
  * Organization-level connections now supported for discovery operations.

* **Improvements**
  * Disconnecting cloud providers and clusters now triggers automatic cleanup of discovered infrastructure entries (AWS, GCP, OVH, Scaleway, Tailscale, Kubernetes), with failures logged but not blocking disconnects.
  * Stale infrastructure data older than 30 days is now automatically removed to improve performance and hygiene.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Arvo-AI/aurora/pull/402)

### DIFF
--- a/server/routes/aws/onboarding.py
+++ b/server/routes/aws/onboarding.py
@@ -345,6 +345,13 @@ def workspace_cleanup(user_id, workspace_id):
             "You can now restart the onboarding flow from scratch."
         )
 
+        # Delete discovered infrastructure nodes from Memgraph
+        try:
+            from services.graph.memgraph_client import get_memgraph_client
+            get_memgraph_client().delete_services_for_provider(user_id, "aws")
+        except Exception as e:
+            logger.warning("Failed to delete Memgraph nodes for user=%s provider=aws: %s", user_id, e)
+
         return jsonify({"success": True, "message": message})
 
     except Exception as e:
@@ -492,6 +499,13 @@ def delete_aws_account(user_id, workspace_id, account_id):
         from utils.db.connection_utils import delete_connection_secret
         success = delete_connection_secret(user_id, "aws", account_id)
         if success:
+            # Delete discovered infrastructure nodes scoped to this account only.
+            # Nodes for other AWS accounts still connected to this user are preserved.
+            try:
+                from services.graph.memgraph_client import get_memgraph_client
+                get_memgraph_client().delete_services_for_aws_account(user_id, account_id)
+            except Exception as e:
+                logger.warning("Failed to delete Memgraph nodes for user=%s aws account=%s: %s", user_id, account_id, e)
             return jsonify({"success": True, "message": f"Account {account_id} disconnected."})
         else:
             return jsonify({"error": "Account not found or already disconnected"}), 404
--- a/server/routes/aws/onboarding.py
+++ b/server/routes/aws/onboarding.py
@@ -520,7 +520,7 @@ def delete_aws_account(user_id, workspace_id, account_id):
             return jsonify({"error": "Account not found or already disconnected"}), 404
 
     except Exception as e:
-        logger.error(
+        logger.exception(
             "Failed to delete AWS account %s for workspace %s: %s",
             sanitize(account_id),
             sanitize(workspace_id),
--- a/server/routes/aws/onboarding.py
+++ b/server/routes/aws/onboarding.py
@@ -348,7 +348,10 @@ def workspace_cleanup(user_id, workspace_id):
         # Delete discovered infrastructure nodes from Memgraph
         try:
             from services.graph.memgraph_client import get_memgraph_client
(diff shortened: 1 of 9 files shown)


server/services/graph/memgraph_client.py
```diff
@@ -773,6 +773,66 @@ def get_graph_stats(self, user_id):
...
+        """Delete all Service nodes (and their edges) for a user/provider pair.
+
+        Called immediately when a cloud provider is disconnected so that
+        stale infrastructure nodes are not left behind in the graph.
+        """
+        query = """
+        MATCH (s:Service {user_id: $user_id, provider: $provider})
+        WITH s, count(s) AS deleted
+        DETACH DELETE s
+        RETURN deleted;
```
beng360: The `WITH s, count(s) AS deleted` pattern doesn't aggregate here because `s` is in the grouping key — it produces one row per node, each with `deleted = 1`. The `sum()` in Python compensates so the final count is correct, but the query is doing unnecessary work returning N rows.

The existing `delete_service` method (line ~289) already has the right pattern:
```cypher
MATCH (s:Service {user_id: $user_id, provider: $provider})
DETACH DELETE s
RETURN count(s) AS deleted;
```
Then the result handling becomes `results[0]["deleted"] if results else 0` (single row).

Same applies to `delete_services_for_cluster`, `delete_services_for_aws_account`, and `delete_stale_services` — all four use the same `WITH s, count(s)` pattern.

server/services/discovery/tasks.py
```diff
@@ -18,19 +18,24 @@ def _query_connected_providers(cur, user_id=None, conn=None):
...
             SELECT DISTINCT provider FROM (
                 SELECT provider FROM user_connections
-                WHERE user_id = %s AND status = 'active' AND provider IN %s
+                WHERE (user_id = %s OR org_id = %s) AND status = 'active' AND provider IN %s
                 UNION
                 SELECT provider FROM user_tokens
                 WHERE user_id = %s AND is_active = true AND provider IN %s
             ) AS connected
-        """, (user_id, SUPPORTED_PROVIDERS, user_id, SUPPORTED_PROVIDERS))
+        """, (user_id, org_id, SUPPORTED_PROVIDERS, user_id, SUPPORTED_PROVIDERS))
```
beng360: The `org_id` fallback was added to `user_connections` but the `user_tokens` subquery right below was left as `WHERE user_id = %s`. The `user_tokens` table also has an `org_id` column, and the existing `get_connected_providers()` in `stateless_auth.py` already uses `(user_id = %s OR org_id = %s)` on **both** tables.

Without this, org-shared OAuth tokens (GCP, Azure, OVH stored in `user_tokens`) will still be silently skipped during discovery. Same issue in the `else` branch at line ~60.

Should be:
```sql
SELECT provider FROM user_tokens
WHERE (user_id = %s OR org_id = %s) AND is_active = true AND provider IN %s
```
And add `org_id` to the parameter tuple for both branches.

server/routes/aws/onboarding.py
```diff
@@ -345,6 +345,17 @@ def workspace_cleanup(user_id, workspace_id):
             "You can now restart the onboarding flow from scratch."
         )
 
+        # Delete discovered infrastructure nodes from Memgraph
+        try:
+            from services.graph.memgraph_client import get_memgraph_client
```
beng360, line 350: `workspace_cleanup` only disconnects a single AWS account's credentials (via `get_user_aws_connection` → `delete_connection_secret`), but this calls `delete_services_for_provider(user_id, "aws")` which nukes **all** AWS nodes across every account.

If a user has multiple AWS accounts and cleans up one workspace, they lose Memgraph nodes for accounts that are still connected.

Since `account_id` is already available a few lines above (line 330), this should use the account-scoped method instead:
```python
get_memgraph_client().delete_services_for_aws_account(user_id, account_id)
```
Same pattern you already use in `delete_aws_account` below.

beng360 (changes requested):
Nice work overall — the disconnect cleanup and stale-deletion safety net are solid additions. Found 3 issues, one of which affects correctness.

beng360 (approved):

PR #413, opened 2026-05-16
Fix: discovery credential failure tracking and enrichment pipeline reliability

Harrio-6:
Adds credential-failure tracking — consecutive auth errors (AWS/GCP/Azure) auto-disconnect the provider after a configurable threshold and notify the user
Distinguishes true credential failures from transient/billing errors (e.g. GCP SCC premium tier returning UNAUTHENTICATED) so those don't accumulate toward auto-disconnect
Fixes GCP isolated env setup where a missing properties file caused gke-gcloud-auth-plugin to fail on every kubectl invocation even with a valid token
Tightens org-level connection deduplication in the cross-org discovery loop to avoid redundant per-user runs
Improves Kubernetes and serverless enrichment resilience

## Summary by CodeRabbit

* **New Features**
  * Per-provider error tracking in discovery results (including unknown-provider attribution)
  * Isolated per-provider subprocess environments for cloud CLIs and ephemeral gcloud configs with automatic cleanup
  * Org-level discovery runs with multi-account AWS/GCP routing and persisted connected project IDs
  * Labeled CLI command logging for clearer diagnostics

* **Bug Fixes**
  * Detects and reports stale Kubernetes clusters
  * Graceful handling of Cloud Asset SCC premium cases
  * Auto-disconnect after repeated credential failures tracked via Redis
* **Other**
  * Disconnect now clears additional GCP preference keys

[Review Change Stack](https://app.coderabbit.ai/change-stack/Arvo-AI/aurora/pull/413?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/server/chat/backend/agent/tools/cloud_exec_tool.py
+++ b/server/chat/backend/agent/tools/cloud_exec_tool.py
@@ -531,6 +531,10 @@ def setup_gcp_environment_isolated(user_id: str, selected_project_id: str | None
                 "GCP isolated env: could not write CLOUDSDK_CONFIG properties (error_type=%s) — falling back to shared path",
                 type(mkdir_err).__name__,
             )
+            # Point CLOUDSDK_CONFIG at the system default (empty string = unset)
+            # so gcloud uses its normal config directory rather than the empty
+            # tmpdir that has no properties file.
+            cloudsdk_config_dir = ""
 
         isolated_env = {
             "PATH": os.environ.get("PATH", ""),
@@ -541,6 +545,7 @@ def setup_gcp_environment_isolated(user_id: str, selected_project_id: str | None
             "GOOGLE_CLOUD_PROJECT": project_id,
             "CLOUDSDK_CONFIG": cloudsdk_config_dir,
             # Expose the temp dir path so callers can clean up after use.
+            # Empty string means the OSError fallback fired and there is nothing to clean up.
             "_gcloud_tmpdir": cloudsdk_config_dir,
         }
         if is_sa_mode:
@@ -560,6 +565,13 @@ def setup_gcp_environment_isolated(user_id: str, selected_project_id: str | None
 
         logger.info("GCP isolated environment configured (%s)", auth_method)
         logger.info("TIME: setup_gcp_environment_isolated completed in %.2fs", time.perf_counter() - fn_start)
+
+        # Safety-net: register an atexit handler so the tmpdir is always
+        # removed even if the caller forgets to clean it up.  This covers
+        # every call-site, not just discovery_service.py.
+        if cloudsdk_config_dir:
+            import atexit, shutil as _shutil
+            atexit.register(_shutil.rmtree, cloudsdk_config_dir, True)
 
         return True, project_id, auth_method, isolated_env
 
--- a/server/chat/backend/agent/tools/cloud_exec_tool.py
+++ b/server/chat/backend/agent/tools/cloud_exec_tool.py
@@ -513,6 +513,7 @@ def setup_gcp_environment_isolated(user_id: str, selected_project_id: str | None
         # on-disk gcloud state persists across invocations.  Each call gets its
         # own directory; the OS reclaims it when the process exits (or earlier
         # when the caller explicitly calls shutil.rmtree on _gcloud_tmpdir).
+        import atexit, shutil as _shutil
         cloudsdk_config_dir = tempfile.mkdtemp(prefix=f".gcloud-{user_id}-")
         try:
             # Write a minimal properties file so gcloud sees an active account.
@@ -528,13 +529,11 @@ def setup_gcp_environment_isolated(user_id: str, selected_project_id: str | None
                 _pf.write(f"[core]\naccount = {account_identity}\n")
         except OSError as mkdir_err:
             logger.warning(
-                "GCP isolated env: could not write CLOUDSDK_CONFIG properties (error_type=%s) — falling back to shared path",
+                "GCP isolated env: could not write CLOUDSDK_CONFIG properties (error_type=%s)",
                 type(mkdir_err).__name__,
             )
-            # Point CLOUDSDK_CONFIG at the system default (empty string = unset)
-            # so gcloud uses its normal config directory rather than the empty
-            # tmpdir that has no properties file.
-            cloudsdk_config_dir = ""
+            _shutil.rmtree(cloudsdk_config_dir, ignore_errors=True)
+            return False, None, None, None
 
         isolated_env = {
             "PATH": os.environ.get("PATH", ""),
@@ -545,7 +544,6 @@ def setup_gcp_environment_isolated(user_id: str, selected_project_id: str | None
             "GOOGLE_CLOUD_PROJECT": project_id,
             "CLOUDSDK_CONFIG": cloudsdk_config_dir,
             # Expose the temp dir path so callers can clean up after use.
-            # Empty string means the OSError fallback fired and there is nothing to clean up.
             "_gcloud_tmpdir": cloudsdk_config_dir,
         }
         if is_sa_mode:
@@ -569,9 +567,7 @@ def setup_gcp_environment_isolated(user_id: str, selected_project_id: str | None
         # Safety-net: register an atexit handler so the tmpdir is always
         # removed even if the caller forgets to clean it up.  This covers
         # every call-site, not just discovery_service.py.
-        if cloudsdk_config_dir:
-            import atexit, shutil as _shutil
-            atexit.register(_shutil.rmtree, cloudsdk_config_dir, True)
+        atexit.register(_shutil.rmtree, cloudsdk_config_dir, True)
 
         return True, project_id, auth_method, isolated_env
 
--- a/server/chat/backend/agent/tools/cloud_exec_tool.py
+++ b/server/chat/backend/agent/tools/cloud_exec_tool.py
@@ -260,7 +260,11 @@ def setup_aws_environment_isolated(user_id: str, selected_region: str | None = N
                 "aws_regions": [selected_region or "us-east-1"],
             }
 
-            logger.info("Assumed workspace role for %s, obtained temporary credentials (expires %s)", hash_for_log(str(ws["id"])), sts_creds["expiration"])
+            logger.info(
+                "Assumed workspace role for %s in region %s",
+                hash_for_log(str(ws["id"])),
+                selected_region or "us-east-1",
+            )
 
         except Exception as e:
             logger.error("Failed to obtain AWS workspace credentials: %s", e)
--- a/server/chat/backend/agent/tools/cloud_exec_tool.py
+++ b/server/chat/backend/agent/tools/cloud_exec_tool.py
@@ -204,7 +204,7 @@ def setup_aws_environment_isolated(user_id: str, selected_region: str | None = N
             else:
                 aws_conn = get_user_aws_connection(user_id)
             if not aws_conn or not aws_conn.get('role_arn'):
-                logger.error("User %s does not have an active AWS connection", user_id)
+                logger.error("User %s does not have an active AWS connection", hash_for_log(user_id))
                 return False, None, None, None
 
             conn_region = aws_conn.get("region")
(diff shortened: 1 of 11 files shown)


server/chat/backend/agent/tools/cloud_exec_tool.py
```diff
@@ -509,19 +509,28 @@ def setup_gcp_environment_isolated(user_id: str, selected_project_id: str | None
...
+            # gke-gcloud-auth-plugin on every kubectl invocation) exits with
+            # "You do not currently have an active account selected" even when
+            # GOOGLE_OAUTH_ACCESS_TOKEN is set, because gcloud still validates
+            # that core/account is populated before resolving the token.
+            properties_path = os.path.join(cloudsdk_config_dir, "properties")
+            account_identity = sa_email if sa_email else "aurora-discovery@local"
+            fd = os.open(properties_path, os.O_WRONLY | os.O_CREAT | os.O_TRUNC, 0o600)
+            with os.fdopen(fd, "w") as _pf:
+                _pf.write(f"[core]\naccount = {account_identity}\n")
         except OSError as mkdir_err:
```
beng360, line 534: The OSError handler logs "falling back to shared path" but no longer reassigns `cloudsdk_config_dir`. After this except block, `isolated_env["CLOUDSDK_CONFIG"]` still points to the mkdtemp directory (which exists but has no properties file). gcloud will see an empty config dir and fail with confusing errors. Either set a fallback path or re-raise.

server/chat/backend/agent/tools/cloud_exec_tool.py
```diff
@@ -531,6 +540,8 @@ def setup_gcp_environment_isolated(user_id: str, selected_project_id: str | None
             "CLOUDSDK_AUTH_ACCESS_TOKEN": access_token,
             "GOOGLE_CLOUD_PROJECT": project_id,
             "CLOUDSDK_CONFIG": cloudsdk_config_dir,
+            # Expose the temp dir path so callers can clean up after use.
+            "_gcloud_tmpdir": cloudsdk_config_dir,
```
beng360, line 551: Cleanup of `_gcloud_tmpdir` is only wired in `discovery_service.py`. Other callers of `setup_gcp_environment_isolated` (the agent cloud_exec chat tool path) will leak one tmpdir per invocation indefinitely. Consider `atexit.register` or documenting the cleanup contract at the function level.

server/services/discovery/enrichment/serverless_enrichment.py
```diff
@@ -305,7 +366,84 @@ def _fetch_cloud_function_env_vars(function_name):
...
+
+    if provider == "gcp" and sub_type == "cloud_function":
+        return _fetch_cloud_function_env_vars(name, project, gcp_env)
+
+    if provider == "aws" and sub_type == "lambda":
+        if _any_aws_env is None:
+            raise ValueError(f"No AWS credentials for Lambda function {name}")
+        node_arn = node.get("cloud_resource_id", "")
+        acct_id = _extract_account_id_from_arn(node_arn)
+        aws_env = aws_account_envs.get(acct_id) or _any_aws_env
```
beng360, line 392: When `acct_id` is non-None but missing from `aws_account_envs`, this falls back to a different account's credentials. The subsequent AccessDenied will surface as an error, but a logger.warning here when `acct_id and acct_id not in aws_account_envs` would make the mismatch immediately obvious in logs rather than requiring someone to trace why the API call failed.

server/services/discovery/providers/gcp_asset_discovery.py
```diff
@@ -46,6 +46,12 @@ def _run_command(args, timeout=600, env=None):
                     "Cloud Asset API is not enabled for this project. "
                     "Enable it with: gcloud services enable cloudasset.googleapis.com"
                 )
+            # SCC premium required for relationship data — not a credential
+            # problem, just a billing tier limitation.  Raise a specific
+            # exception so callers can downgrade to a warning.
+            if "Relationship is only supported for SCC premium customers" in stderr or \
+               ("UNAUTHENTICATED" in stderr and "premium customers" in stderr):
+                raise RuntimeError("__SCC_PREMIUM_REQUIRED__")
```
beng360: Nit: both the raise and the catch are in the same file, so practical risk is low. But a custom exception subclass (`class SCCPremiumRequiredError(RuntimeError): pass`) would be more idiomatic and make the control flow explicit without relying on string matching.

server/services/discovery/tasks.py
```diff
@@ -74,6 +76,173 @@ def _clear_discovery_lock(user_id):
...
+    try:
+        conn = connect_to_db_as_admin()
+        with conn.cursor() as cur:
+            set_rls_context(cur, conn, user_id, log_prefix="[Discovery:AutoDisconnect]")
+            now = datetime.now(timezone.utc)
+            cur.execute(
+                """
+                UPDATE user_connections
+                SET status = 'inactive', last_verified_at = %s
+                WHERE (user_id = %s OR org_id = %s) AND provider = %s AND status = 'active'
```
beng360, line 195: Minor: when `org_id` is None, `org_id = %s` becomes `org_id = NULL` which is never true in SQL. The user-scoped rows still match via `user_id = %s` so it's not a functional failure. In the current call paths (beat-scheduled discovery), `org_id` is always non-None by the time this runs, so this is unreachable today. Just flagging for awareness if this function gets called from other paths in the future.

server/services/discovery/discovery_service.py
```diff
@@ -264,6 +273,33 @@ def run_discovery_for_user(user_id, connected_providers):
                 summary["phase2_relationships"] += write_dependencies(user_id, k8s_rels)
             if k8s_result.get("errors"):
                 summary["errors"].extend(k8s_result["errors"])
+                # Attribute each error to its provider via the structured "provider"
+                # field when available; fall back to substring matching only as a
+                # last resort so misattribution is minimised.
+                for err in k8s_result["errors"]:
+                    attributed = False
+                    if isinstance(err, dict):
```
beng360: `isinstance(err, dict)` branch is unreachable. `kubernetes_enrichment.enrich()` only returns plain strings in its `errors` list. The structured `err.get("provider")` path will never fire. Either update the enrichment module to return dicts, or remove this dead code.

server/services/discovery/discovery_service.py
```diff
@@ -264,6 +273,33 @@ def run_discovery_for_user(user_id, connected_providers):
...
+                        pname = err.get("provider")
+                        if pname and pname in connected_providers:
+                            summary["provider_errors"].setdefault(pname, []).append(
+                                err.get("message", err)
+                            )
+                            attributed = True
+                    if not attributed:
+                        err_str = err if isinstance(err, str) else str(err)
+                        for pname in ("gcp", "aws", "azure"):
+                            if pname in connected_providers and pname in err_str.lower():
```
beng360: Nit: `pname in err_str.lower()` does substring matching, so cluster names containing "gcp" or "aws" would match. In practice this is usually correct attribution (a cluster named "aws-prod" IS on AWS), but worth noting the assumption.

server/services/discovery/discovery_service.py
```diff
@@ -333,4 +393,15 @@ def run_discovery_for_user(user_id, connected_providers):
...
         f"{len(summary['errors'])} errors, {elapsed:.1f}s"
     )
+
+    # Clean up any ephemeral gcloud temp directories created during this run.
+    import shutil
+    for _pname, (env, _) in provider_envs.items():
+        tmpdir = env.get("_gcloud_tmpdir") if isinstance(env, dict) else None
+        if tmpdir:
+            try:
+                shutil.rmtree(tmpdir, ignore_errors=True)
```
beng360: `shutil.rmtree(tmpdir, ignore_errors=True)` swallows all exceptions internally, making the enclosing `except Exception` unreachable. Either drop `ignore_errors=True` (and rely on the try/except for the debug log) or remove the try/except wrapper.

server/services/discovery/enrichment/kubernetes_enrichment.py
```diff
@@ -405,35 +442,52 @@ def _build_relationships(ingress_backends, service_nodes, workload_nodes, cluste
...
 
 
-def _discover_cluster(cluster, provider_credentials):
-    """Discover internal resources for a single Kubernetes cluster.
+_STALE_CLUSTER_FRAGMENTS = (
+    # EKS cluster deleted from AWS
+    "ResourceNotFoundException",
+    "No cluster found for name",
+    # GKE cluster deleted or project removed
+    "NOT_FOUND",
```
beng360, line 460: Minor: `_is_stale_cluster_error` is only called on `_get_cluster_credentials` output (not kubectl fetch errors), so the scope is appropriately limited. However `"NOT_FOUND"` and `"cluster not found"` could overlap with other error contexts if you later expand the call sites. Consider a comment noting this function is only valid for credential-command stderr.

server/services/discovery/enrichment/kubernetes_enrichment.py
```diff
@@ -405,35 +442,52 @@ def _build_relationships(ingress_backends, service_nodes, workload_nodes, cluste
...
-    logger.info(f"K8s enrichment: getting credentials for cluster {cluster_name}")
-    cred_error = _get_cluster_credentials(cluster, provider_credentials)
-    if cred_error:
-        error_msg = f"Failed to get credentials for cluster {cluster_name}: {cred_error}"
-        logger.warning(error_msg)
-        errors.append(error_msg)
-        return nodes, relationships, errors
 
-    # Step 2: Fetch all resource types via kubectl
+def _resolve_kubectl_env(cluster, provider_envs):
```
beng360, line 481: Edge case: when `provider_envs.get(provider)` is None, this returns `{"PATH": ...}` without HOME. Since `subprocess.run` with an explicit `env` replaces the entire environment, kubectl won't find the kubeconfig written by `get-credentials` (which used the full env). In practice this fallback rarely triggers for connected providers, but if it does, kubectl calls will fail silently. Consider including HOME in the fallback.

beng360 (changes requested):
Credential failure tracking + org-level dedup are solid. The enrichment pipeline isolation is a needed fix. A few correctness and safety issues to address.

server/chat/backend/agent/tools/cloud_exec_tool.py
Harrio-6: Done. Now cleans up temp directory, returns early, never builds isolated_env

server/chat/backend/agent/tools/cloud_exec_tool.py
Harrio-6: Fixed, line 574 has gurantee backstop for all call sites.

server/services/discovery/enrichment/serverless_enrichment.py
Harrio-6: Fixed, added the warning

server/services/discovery/tasks.py
Harrio-6: Fixed, added logger.warning in case None org ids happen

server/services/discovery/enrichment/kubernetes_enrichment.py
Harrio-6: Fixed, added comment at lines 469-476

server/services/discovery/enrichment/kubernetes_enrichment.py
Harrio-6: Fixed, added Home in fallback at line 496

server/services/discovery/providers/gcp_asset_discovery.py
Harrio-6: fixed, added exception subclass

server/services/discovery/discovery_service.py
Harrio-6: fixed, removed

server/services/discovery/discovery_service.py
Harrio-6: acknowledged, but ignored. Usually sth named aws-prod will be in aws for example

server/services/discovery/discovery_service.py
Harrio-6: Fixed, removed redundant try except wrapper

beng360 (approved):
All major issues addressed. One minor nit: _build_env_for_account and _build_single_account_env in serverless_enrichment.py still lack HOME in the minimal env. The kubernetes_enrichment side was fixed — this is the same pattern. Low-risk since AWS CLI works without HOME when creds are in env vars, but some CLI versions try to write to ~/.aws/cli/cache.

PR #424, opened 2026-05-19
Fix: scope kubernetes to org instead of user

Harrio-6:
Switch all kubectl token/connection queries from user_id-scoped to org_id-scoped so any member of an org can see and use clusters registered by teammates
Replace get_org_id_from_request() with resolve_org_id(user_id) in kubectl_token_routes.py to correctly resolve org context outside request headers
Fix cluster authorization in agent_ws_handler.py to verify org membership rather than exact user ownership
Fix the discovery pipeline (tasks.py) to query kubectl_agent_tokens / active_kubectl_connections directly (since kubectl connections never appear in user_connections) and include kubectl-only orgs in the full discovery loop
Remove kubectl from SUPPORTED_PROVIDERS to prevent the standard provider query from incorrectly handling it
Add idx_kubectl_tokens_org index on kubectl_agent_tokens(org_id) with a safe migration savepoint for existing deployments

## Summary by CodeRabbit

* **New Features**
  * On-premises Kubernetes connections are now scoped at the organization level, allowing multiple organization members to access shared cluster credentials and connections.

* **Updates**
  * Kubernetes connection discovery and authorization logic now operates at the organization level rather than the individual user level.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Arvo-AI/aurora/pull/424?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/server/services/discovery/tasks.py
+++ b/server/services/discovery/tasks.py
@@ -9,7 +9,11 @@ from utils.auth.stateless_auth import set_rls_context, get_org_id_for_user
 
 logger = logging.getLogger(__name__)
 
-SUPPORTED_PROVIDERS = ('gcp', 'aws', 'azure', 'ovh', 'scaleway', 'tailscale', 'kubectl')
+# Providers whose connections are stored in user_connections / user_tokens.
+# kubectl is intentionally excluded: its connections live in the separate
+# kubectl_agent_tokens / active_kubectl_connections tables and are picked up
+# by the explicit kubectl queries in run_full_discovery / run_user_discovery.
+SUPPORTED_PROVIDERS = ('gcp', 'aws', 'azure', 'ovh', 'scaleway', 'tailscale')
 
 
 def _query_connected_providers(cur, user_id=None, conn=None):
@@ -165,7 +169,7 @@ def _reset_provider_failure(user_id: str, provider: str) -> None:
 def _mark_provider_inactive(user_id: str, provider: str) -> None:
     """Mark all active connections for (user_id, provider) as inactive.
 
-    Covers both user_connections (AWS/kubectl/OVH/Scaleway/Tailscale) and
+    Covers both user_connections (AWS/OVH/Scaleway/Tailscale) and
     user_tokens (GCP/Azure and other OAuth-backed providers).  Deactivates
     both user-scoped and org-scoped rows so org-shared connections are also
     cleaned up.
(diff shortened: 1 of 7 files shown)


server/services/discovery/tasks.py
```diff
@@ -9,7 +9,38 @@
...
+            """SELECT 1 FROM kubectl_agent_tokens t
+               JOIN active_kubectl_connections c ON t.token = c.token
+               WHERE (t.user_id = %s OR t.org_id = %s)
+                 AND t.status = 'active' AND c.status = 'active'
+               LIMIT 1""",
+            (uid, org_id),
+        )
+        if cur.fetchone():
+            result.append((org_id, uid))
+    return result
```
damianloch: This function fetches ALL users in the system, then executes a query per user to check if they have kubectl clusters. On a platform with hundreds/thousands of users this will be slow and hammers the DB on every hourly discovery run. Rewrite as a single query that doesn't need per-user RLS iteration

server/services/discovery/tasks.py
```diff
@@ -472,6 +506,41 @@ def _process_org(org_id, org_info, get_owner, run_discovery):
...
+                    "[Discovery Task] Found %d active kubectl cluster(s) for org=%s",
+                    len(clusters), org_id,
+                )
+            else:
+                del providers["kubectl"]
+        except Exception as e:
+            logger.warning("[Discovery Task] kubectl cluster query failed for org=%s: %s", org_id, e)
+            del providers["kubectl"]
+        finally:
+            if conn:
```
damianloch: _query_kubectl_orgs already found which orgs have kubectl clusters, but _process_org opens a new admin connection to query the exact same data again.

server/services/discovery/tasks.py
```diff
@@ -9,7 +9,38 @@
...
+        cur.execute(
+            """SELECT 1 FROM kubectl_agent_tokens t
+               JOIN active_kubectl_connections c ON t.token = c.token
+               WHERE (t.user_id = %s OR t.org_id = %s)
+                 AND t.status = 'active' AND c.status = 'active'
+               LIMIT 1""",
+            (uid, org_id),
+        )
+        if cur.fetchone():
+            result.append((org_id, uid))
```
damianloch: The query is SELECT DISTINCT id, org_id FROM users WHERE org_id IS NOT NULL. If an org has 5 users, you'll iterate 5 times and potentially append the same (org_id, uid) multiple times (whichever user first matches).

damianloch (changes requested):

damianloch (approved):

PR #425, opened 2026-05-19
fix(mcp): connector status shows callable_now: false despite dashboard showing Connected

beng360:
## Summary

- The MCP server runs as a standalone process (no Flask context), so `SkillRegistry.check_connection()` silently fails when querying RLS-protected tables (`user_tokens`, `user_connections`) — all gated tools return `callable_now: false`
- The dashboard's `/api/connectors/status` endpoint works correctly because it runs within Flask with proper RLS context
- Fix: the MCP now fetches connector status from the backend API and caches it (30s TTL), using that as the authoritative source for skill connectivity checks

## Test plan

- [ ] Connect GitHub/GCP in the dashboard, verify `search_tools` returns `callable_now: true` for gated tools
- [ ] Verify `call_tool` works for connected integrations
- [ ] Verify disconnected integrations still show `callable_now: false`
- [ ] Verify cache TTL — disconnect a provider, wait 30s, confirm MCP reflects the change

## Summary by CodeRabbit

* **New Features**
  * Redesigned Actions list into full-width actionable cards with an improved empty state and "Create Action" button.

* **Bug Fixes**
  * Updated postmortem handling and success messaging after incident resolution to reflect background generation and reduce false "no postmortem" states.
  * Improved connector-status detection to make tool availability more accurate.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Arvo-AI/aurora/pull/425?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/client/src/app/actions/page.tsx
+++ b/client/src/app/actions/page.tsx
@@ -176,16 +176,14 @@ function ActionsListView({ actions, onSelect, onCreate }: {
       </div>
 
       {actions.length === 0 ? (
-        <Panel title="Configured Actions" subtitle="Background agent tasks that follow your instructions">
-          <div className="flex flex-col items-center justify-center py-16 text-center">
-            <Workflow className="h-8 w-8 text-zinc-700 mb-3" />
-            <p className="text-sm text-zinc-500">No actions yet</p>
-            <p className="text-xs text-zinc-600 mt-1 mb-4">Create your first action to automate SRE workflows</p>
-            <Button variant="outline" size="sm" onClick={onCreate}>
-              <Plus className="h-3.5 w-3.5 mr-1.5" /> Create Action
-            </Button>
-          </div>
-        </Panel>
+        <div className="flex flex-col items-center justify-center py-16 text-center border border-zinc-800/80 rounded-xl bg-zinc-900/60">
+          <Workflow className="h-8 w-8 text-zinc-700 mb-3" />
+          <p className="text-sm text-zinc-500">No actions yet</p>
+          <p className="text-xs text-zinc-600 mt-1 mb-4">Create your first action to automate SRE workflows</p>
+          <Button variant="outline" size="sm" onClick={onCreate}>
+            <Plus className="h-3.5 w-3.5 mr-1.5" /> Create Action
+          </Button>
+        </div>
       ) : (
         <div className="space-y-3">
           {actions.map((action) => (
--- a/client/src/app/actions/page.tsx
+++ b/client/src/app/actions/page.tsx
@@ -175,8 +175,8 @@ function ActionsListView({ actions, onSelect, onCreate }: {
         <StatCard label="Total Runs" value={String(totalRuns)} icon={Play} />
       </div>
 
-      <Panel title="Configured Actions" subtitle="Background agent tasks that follow your instructions">
-        {actions.length === 0 ? (
+      {actions.length === 0 ? (
+        <Panel title="Configured Actions" subtitle="Background agent tasks that follow your instructions">
           <div className="flex flex-col items-center justify-center py-16 text-center">
             <Workflow className="h-8 w-8 text-zinc-700 mb-3" />
             <p className="text-sm text-zinc-500">No actions yet</p>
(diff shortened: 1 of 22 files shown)


server/aurora_mcp/registry.py
```diff
@@ -15,12 +15,39 @@
...
+
+def _get_cached_connector_status(user_id: str) -> Optional[Dict[str, bool]]:
+    """Return cached connector status if fresh, else None."""
+    entry = _connector_cache.get(user_id)
+    if entry is None:
+        return None
+    ts, status = entry
+    if time.monotonic() - ts > _CONNECTOR_CACHE_TTL:
+        return None
+    return status
```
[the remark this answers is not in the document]
beng360: Connector cache code was reverted from this PR — no longer applicable.

server/mcp_server.py
```diff
@@ -346,6 +367,10 @@ async def _filtered_list_tools():  # type: ignore[no-redef]
         )
         return [t for t in all_tools if t.name not in _GATED_NAMES]
 
+    # Populate connector cache from the backend API so that gated_tool_visible
+    # (and downstream search_tools/call_tool) use the authoritative status.
+    await _refresh_connector_cache(user_id)
```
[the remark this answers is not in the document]
beng360: Connector cache code was reverted from this PR — no longer applicable.

server/aurora_mcp/registry.py
```diff
@@ -15,12 +15,39 @@
...
 logger = logging.getLogger(__name__)
 
 
+# ---------------------------------------------------------------------------
+# Connector-status cache — populated via the Flask backend API by the MCP
+# server so skill-connectivity checks don't rely on local DB access (which
+# can fail due to missing Flask/RLS context in the standalone MCP process).
+# ---------------------------------------------------------------------------
+
+_connector_cache: Dict[str, Tuple[float, Dict[str, bool]]] = {}
```
Harrio-6, line 31: connector_cache does not have a way to delete entries, the list keeps on growing. Either add an upper limit or a way to remove entires.

server/aurora_mcp/dispatch.py
```diff
@@ -177,6 +179,22 @@ def _user_id() -> str:
...
+            data = await api_call("GET", "/api/connectors/status")
+            connectors = data.get("connectors", {})
+            status = {
+                provider.lower(): bool(info.get("connected"))
+                for provider, info in connectors.items()
+                if isinstance(info, dict)
+            }
+            set_connector_status(user_id, status)
+        except Exception:
+            logger.debug("connector cache refresh failed in dispatch", exc_info=True)
```
Harrio-6: I think that this should throw an error instead of a debug, because if not it will go back to the old behaviour and the user would not know that the issue is not with their setup

server/mcp_server.py
```diff
@@ -325,12 +325,33 @@ async def _api(
...
+        data = await _api("GET", "/api/connectors/status")
+        connectors = data.get("connectors", {})
+        status = {
+            provider.lower(): bool(info.get("connected"))
+            for provider, info in connectors.items()
+            if isinstance(info, dict)
+        }
+        set_connector_status(user_id, status)
+    except Exception:
+        logger.debug("connector cache refresh failed for user=%s", user_id, exc_info=True)
```
Harrio-6: I think its the same issue here, should be an error instead of a debug

server/aurora_mcp/dispatch.py
Harrio-6: There is two different functions doing the same thing (this and refresh_connector_cache in mcp_server.py). Maybe create a helper

client/src/app/actions/page.tsx
```diff
@@ -175,8 +175,8 @@ function ActionsListView({ actions, onSelect, onCreate }: {
         <StatCard label="Total Runs" value={String(totalRuns)} icon={Play} />
       </div>
 
-      <Panel title="Configured Actions" subtitle="Background agent tasks that follow your instructions">
-        {actions.length === 0 ? (
+      {actions.length === 0 ? (
+        <Panel title="Configured Actions" subtitle="Background agent tasks that follow your instructions">
```
Harrio-6: I'm not sure what "Configured Actions" is for, because right now it only shows when there are no conigured actions

Image

server/main_chatbot.py
```diff
@@ -1550,6 +1557,7 @@ async def confirmation_sender(data):
             # Fetch org tool permissions for gate bypass
             _permitted_tools = None
             try:
+                from utils.db.connection_pool import db_pool
```
Harrio-6: move this import to top of page

server/main_chatbot.py
```diff
@@ -174,6 +174,13 @@ async def _ws_reject(websocket, text: str, *, close_reason: str = ""):
 
 def _warm_user_caches(user_id: str):
     """Kick off background tasks that pre-warm per-user caches."""
+    import uuid as _uuid
```
Harrio-6: move import to top

server/main_chatbot.py
```diff
@@ -174,6 +174,13 @@ async def _ws_reject(websocket, text: str, *, close_reason: str = ""):
...
+    try:
+        _uuid.UUID(user_id)
+    except (ValueError, TypeError, AttributeError):
```
Harrio-6: I do not believe AttributeError is possible from uuid.UUID(user_id)

Harrio-6 (changes requested):

server/aurora_mcp/registry.py
beng360: Fixed. Stale entries are now evicted with `_connector_cache.pop(user_id, None)` when TTL expires. Also replaced the raw dict with `parse_and_cache_connector_status` shared helper so parsing logic lives in one place.

server/aurora_mcp/dispatch.py
beng360: Fixed. Changed to `logger.error` so failures are visible without debug-level logging enabled.

server/mcp_server.py
beng360: Fixed. Changed to `logger.error` here as well.

server/aurora_mcp/dispatch.py
beng360: Fixed. Extracted `parse_and_cache_connector_status(user_id, api_response)` into registry.py. Both mcp_server.py and dispatch.py now call it instead of duplicating the connector-status parsing logic.

client/src/app/actions/page.tsx
beng360: Fixed. Removed the Panel wrapper with "Configured Actions" title from the empty state. Now shows a simple centered empty-state without the confusing heading.

server/main_chatbot.py
beng360: Fixed. Removed the local import — `db_pool` is already imported at the top of the file (line 47).

server/main_chatbot.py
beng360: Fixed. Removed the local `import uuid as _uuid` — the module-level `import uuid` (line 19) is used directly now.

server/main_chatbot.py
beng360: Fixed. Removed `AttributeError` from the except clause — `uuid.UUID()` only raises `ValueError` or `TypeError`.

client/src/app/incidents/components/PostmortemPanel.tsx
```diff
@@ -53,13 +56,20 @@ export default function PostmortemPanel({ incidentId, incidentTitle, isVisible,
...
-      setPostmortemNotFound(true);
+      notFoundCountRef.current += 1;
+      if (justResolved && notFoundCountRef.current < 5) {
+        setRegenerating(true);
+      } else {
+        setPostmortemNotFound(true);
+      }
     }
-  }, [incidentId, regenerating]);
+  }, [incidentId, regenerating, justResolved]);
```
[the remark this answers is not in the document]
beng360: Fixed. Removed the `if (regenerating) return` guard so `notFoundCountRef` always increments when backend reports no data and not generating. This ensures the 5-attempt give-up threshold is respected even during polling, rather than running for the full 40 attempts.

client/src/app/incidents/components/PostmortemPanel.tsx
```diff
@@ -42,24 +103,37 @@ export default function PostmortemPanel({ incidentId, incidentTitle, isVisible,
...
-      setRegenerating(false);
-    } else if (result.generating) {
-      setPostmortemNotFound(false);
-      setRegenerating(true);
-    } else if (!regenerating) {
-      setPostmortemNotFound(true);
-    }
-  }, [incidentId, regenerating]);
+    handlePostmortemLoaded(result.data, result.generating);
+  }, [incidentId, handlePostmortemLoaded]);
```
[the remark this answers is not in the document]
beng360: Fixed. Added useEffect to reset notFoundCountRef.current to 0 when incidentId changes.

Harrio-6 (approved):

damianloch (approved):

PR #426, opened 2026-05-19
Fix: gate save_postmortem to postmortem action sessions only

Harrio-6:
Remove save_postmortem from the RCA agent's tool list — it is now only injected when the session is the dedicated "Generate Postmortem" action
Add is_postmortem_action flag to State, set in task.py by inspecting trigger_metadata.source and a lazy DB lookup on actions.system_key
Update tool cache key to include the new flag so RCA and postmortem sessions never share a cached tool set
Add UUID validation in postmortem_tool.py as a defense-in-depth guard against ULIDs or hallucinated strings reaching PostgreSQL
Update SKILL.md to reflect that save_postmortem is not available during RCA

### DIFF
--- a/server/chat/backend/agent/skills/integrations/postmortem/SKILL.md
+++ b/server/chat/backend/agent/skills/integrations/postmortem/SKILL.md
@@ -7,7 +7,6 @@ connection_check:
   method: always
 tools:
   - get_postmortem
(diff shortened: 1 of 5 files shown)


beng360 (approved):
Tested locally — all paths work correctly:

PR #427, opened 2026-05-20
fix(rca/github): resolve relation rename, snippet wipe, missing incident_start_time

isiddharthsingh:
## Summary
- **GitHub tool auth failure** — auth router/webhook/disconnect SQL still queried the legacy `github_connected_repos`; live DB only has `connected_repos`. Every `github_*` tool call returned `relation does not exist`.
- **PR diff wiped the file** — LLM was passing only the changed snippet as `suggested_content`; `push_files` overwrote the whole file. Added a length-ratio guard against the fetched original and tightened the tool/field descriptions.
- **`github_rca` failed-closed inside sub-agents** — dispatcher's `Send` payload never forwarded `incident_start_time`, and `sub_agent_node` rebuilt a fresh `State` without it, clobbering the parent's contextvar via `agentic_tool_flow → set_user_context(state=sub_state)`. Now forwarded end-to-end.
- **Incident-chat textbox also missed the timestamp** — `main_chatbot.py` built State with `incident_id` but never fetched `incidents.started_at`; added the same fetch used by the background-RCA path.

## Test plan
- [ ] Rebuild + roll `server`, `chatbot`, `celery_worker`
- [ ] Auto-RCA on a fresh incident → sub-agents call `github_rca` successfully (no `pinned timestamp` error)
- [ ] Type a follow-up in the incident chat → `github_rca` from chat works
- [ ] Approve a fix suggestion that touches a real multi-hundred-line file → PR diff shows only the targeted change, not whole-file deletion (or the LLM gets the new "too short" error and retries)
- [ ] Connect/disconnect GitHub App → no `relation "github_connected_repos" does not exist` errors in logs

## Summary by CodeRabbit

* **New Features**
  * Propagate incident start time through agent orchestration and sub-agent flows.
  * New anchored search-and-replace fix workflow: submit edits (old/new/replace_all) and have the server apply and save suggested fixes.
  * GitHub integration now targets the consolidated connected_repos records for repo-linkage operations.

* **Bug Fixes**
  * Reject ambiguous, empty, no-op, overlapping, or mixed-encoding edits to prevent unsafe or ineffective fixes.

* **Documentation**
  * Updated guidance for crafting and submitting code-fix edits and tool usage.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Arvo-AI/aurora/pull/427?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/server/chat/backend/agent/tools/github_fix_tool.py
+++ b/server/chat/backend/agent/tools/github_fix_tool.py
@@ -1,13 +1,17 @@
 """
 GitHub Fix Tool - Suggest code fixes during RCA.
 
-This tool allows the RCA agent to suggest code fixes when it identifies
-a code issue that caused an incident. The fix is stored as a suggestion
-for user review before being applied.
+The LLM proposes one or more anchored search-and-replace edits. The server
+fetches the current file from GitHub, applies the edits, stores the resulting
+full file body as the suggestion, and the user reviews + creates the PR from
+the UI.
+
+Replacer chain (exact → fuzzier) is ported from opencode/cline/gemini-cli
+edit tools to absorb common LLM whitespace/indent drift.
 """
 
 import logging
-from typing import Optional
+from typing import Optional, Callable, Generator
 from pydantic import BaseModel, Field
 
 from .github_mcp_utils import (
@@ -19,10 +23,29 @@ from .github_mcp_utils import (
 
 logger = logging.getLogger(__name__)
 
-# Reject suggested_content much smaller than the original — likely a snippet
-# rather than the full file body push_files will overwrite with.
-MIN_FILE_SIZE_FOR_TRUNCATION_CHECK = 500
-MIN_CONTENT_RATIO_THRESHOLD = 0.4
+
+# ---------------------------------------------------------------------------
+# Args schema
+# ---------------------------------------------------------------------------
+
+
+class FixEdit(BaseModel):
+    """A single anchored search-and-replace edit."""
+    old_string: str = Field(
+        description=(
+            "Exact text to match in the current file. Include enough surrounding "
+            "context (typically 1–3 lines above and below the change) to make the "
+            "match unique. Whitespace counts; if you copy from get_file_contents "
+            "the indentation will be right."
+        )
+    )
+    new_string: str = Field(
+        description="Replacement text. Indentation must match what belongs at that location."
+    )
+    replace_all: bool = Field(
+        default=False,
+        description="Replace every occurrence of old_string. Default False requires exactly one match.",
+    )
 
 
 class GitHubFixArgs(BaseModel):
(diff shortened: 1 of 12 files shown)


server/chat/backend/agent/tools/github_fix_tool.py
damianloch, line 1: this creates a non scalable tool.  This is already a solved problem and string replacement is the solution. Refert to https://github.com/anomalyco/opencode/blob/dev/packages/opencode/src/tool/edit.ts for example. Instead of makign the LLM rewrite an entier file, use a string replace command to replace as needed. This way you can edit 1 line in a 20k line package-lock.json without need for rewriting whole thing, which is likely the reason the LLM had resorted to only outputting diff

server/chat/backend/agent/tools/github_fix_tool.py
```diff
@@ -183,6 +195,31 @@ def github_fix(
...
     if original_content is None:
         logger.warning(f"Could not fetch original content for {file_path}, proceeding without it")
+    else:
+        orig_lines = original_content.count("\n") + 1
+        sugg_lines = suggested_content.count("\n") + 1
+        orig_len = len(original_content)
+        sugg_len = len(suggested_content)
+        if orig_len > MIN_FILE_SIZE_FOR_TRUNCATION_CHECK and (
+            sugg_len < MIN_CONTENT_RATIO_THRESHOLD * orig_len
+            or sugg_lines < MIN_CONTENT_RATIO_THRESHOLD * orig_lines
```
damianloch: non deterministic LLM slop. there are plenty of times where this can be a valid code change.

damianloch (changes requested):

server/chat/backend/agent/tools/github_fix_tool.py
isiddharthsingh: thanks, implemented it for our usecase

server/main_chatbot.py
```diff
@@ -1548,6 +1548,29 @@ async def confirmation_sender(data):
...
+                            )
+                            _row = _cur.fetchone()
+                            if _row and _row[0]:
+                                _started_at = _row[0]
+                                if _started_at.tzinfo is None:
+                                    _started_at = _started_at.replace(tzinfo=_tz.utc)
+                                _incident_start_time = _started_at.astimezone(_tz.utc).strftime("%Y-%m-%dT%H:%M:%SZ")
+                except Exception as _e:
+                    logger.warning("[Chatbot] Could not fetch incident started_at for %s: %s", _incident_id, _e)
+
```
damianloch: this is fine but please place in another helper file and import since main_chatbot is getting huge

server/chat/backend/agent/tools/github_fix_tool.py
```diff
@@ -20,53 +24,523 @@
...
+        end = start
+        for k in range(len(search_lines)):
+            end += len(original_lines[i + k])
+            if k < len(search_lines) - 1:
+                end += 1  # newline
+        yield content[start:end]
+
+
+_SINGLE_CANDIDATE_SIMILARITY = 0.5
+_MULTIPLE_CANDIDATES_SIMILARITY = 0.5
```
damianloch: these are 0.0 and 0.3 for opencode, not sure if you chose to change the numbers for a reason but they probably have better battle tested thresholds

server/chat/backend/agent/tools/github_fix_tool.py
damianloch, line 1: make sure to cite properly they have MIT license but you still need to comment at top of this file I think (if you built off of opencode)

server/chat/backend/agent/tools/github_fix_tool.py
```diff
@@ -20,53 +24,523 @@
...
+
+            last_idx = content.rfind(candidate)
+            if idx != last_idx:
+                # multiple matches via this replacer — keep trying.
+                continue
+
+            # If the escape-normalized replacer is what made this match, the
+            # LLM's new_string is probably escaped the same way — unescape it
+            # too so the file gets real newlines/tabs instead of backslash-n.
+            effective_new = _unescape_string(new) if name == "escape_normalized" else new
```
damianloch: when the escape-normalized replacer matches, new_string is also run through _unescape_string. This means if the replacement text intentionally contains literal backslash sequences (regex patterns, log format strings, JSON templates), they'll be silently converted to control characters. OpenCode's equivalent (EscapeNormalizedReplacer) only unescapes the find/old_string for matching purposes — new_string is always substituted verbatim.

server/chat/backend/agent/tools/github_fix_tool.py
```diff
@@ -20,53 +24,523 @@
...
+                f"file ({len(old_lf)} of {orig_lf_len} chars). An anchored edit "
+                "should target a narrow window around the change — usually 1–3 "
+                "lines of context above and below. Break this into smaller "
+                "targeted edits instead of regenerating the whole file."
+            )
+
+        new_content, err = _replace_with_chain(content, old_lf, new_lf, replace_all)
+        if err is not None or new_content is None:
+            return None, f"edit {i}: {err}"
+        content = new_content
```
damianloch: _apply_edits applies edits sequentially against the running content (line 497). If edit 2's old_string happens to match text that edit 1 just inserted, it will silently replace the newly-inserted code. There's no guard or warning for this — the LLM could accidentally craft edits that interfere with each other. OpenCode avoids this by design since it only processes one edit per tool call.

damianloch (changes requested):

damianloch (approved):

PR #432, opened 2026-05-20
feat: add infrastructure context tool for prediscovery + MCP

damianloch:
The prediscovery agent now produces a consolidated infrastructure context document as its primary deliverable (while still saving individual findings for backward-compatible RCA search). The document is exposed as a tool to both internal agents and external MCP clients.

Changes:
- New infrastructure_context table (org_id PK, content, updated_at)
- New infra_context_tool.py with save/get functions
- Registered save_infrastructure_context (prediscovery mode only)
- Registered get_infrastructure_context (all authenticated users)
- Rewritten prediscovery prompt targeting consolidated doc output
- GET /api/graph/infrastructure/context API endpoint
- MCP dispatch entry (Tier 3, always-on, category=topology)
- Knowledge base skill hint for internal agent discoverability

## Summary by CodeRabbit

* **New Features**
  * Add tools to save and retrieve a consolidated infrastructure context (topology, services, deployments, dependencies); exposed to authenticated users and used during prediscovery.
  * Prediscovery agents now produce and save a single synthesized infrastructure-context document.

* **API**
  * New authenticated endpoint returns stored infrastructure context with a last-updated timestamp.

* **Tests**
  * Coverage updated to exclude the infrastructure context store from RLS checks.

* **Chores**
  * Added infrastructure_context storage table to initialization scripts.

* **Documentation**
  * Updated agent guidance to consult and generate infrastructure context during investigations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Arvo-AI/aurora/pull/432?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


server/tests/architectural/test_rls_coverage.py
```diff
@@ -19,6 +19,7 @@
     "org_invitations",            # queried during invite/join flows before org context
     "knowledge_base_documents",   # Celery cleanup_stale_documents runs cross-org sweeps
     "knowledge_base_memory",      # same Celery task dependency as knowledge_base_documents
+    "infrastructure_context",     # org_id is the PK; single row per org, queried directly
```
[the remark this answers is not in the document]
damianloch: The table uses `org_id` as its primary key (one row per org). Every query is an exact PK match (`WHERE org_id = %s`), so RLS would be redundant. Additionally, the prediscovery Celery task writes via `get_admin_connection` without RLS context -- adding RLS would break the write path.

server/chat/backend/agent/tools/infra_context_tool.py
```diff
@@ -0,0 +1,114 @@
...
+
+def save_infrastructure_context(
+    content: str,
+    user_id: str | None = None,
+    **kwargs,
+) -> str:
+    """Upsert the infrastructure context document for the user's org."""
+    if not user_id:
+        return "Error: User authentication required."
+    if not content or len(content.strip()) < 100:
```
beng360, line 50: no max limit here, we should put a limiter on the TEXT column a bit like the postmortem tool (100k).

server/routes/graph_routes.py
```diff
@@ -128,6 +128,39 @@ def delete_dependency(user_id, dep_id):
...
+
+        with db_pool.get_admin_connection() as conn:
+            with conn.cursor() as cur:
+                cur.execute(
+                    "SELECT content, updated_at FROM infrastructure_context WHERE org_id = %s",
+                    (org_id,),
+                )
+                row = cur.fetchone()
+    except Exception as e:
+        logger.exception(f"Failed to fetch infrastructure context: {e}")
```
beng360: here the exception string prints twice

server/chat/background/prediscovery_task.py
```diff
@@ -75,22 +75,44 @@ def build_prediscovery_prompt(user_id: str, providers: List[str], integrations:
...
+These are indexed separately for semantic search during incident response.
+
 ## CRITICAL RULES
 
 - The local filesystem is Aurora's own code -- NEVER use terminal_exec to read local files (ls, cat, find, grep, env). There is nothing useful locally.
 - terminal_exec is ONLY allowed for SSH into manual VMs (e.g. ssh -i ~/.ssh/id_key user@ip).
-- Each finding must be a detailed paragraph describing real infrastructure you discovered by querying external APIs. Not a summary of how Aurora works.
+- Each finding must describe real infrastructure you discovered by querying external APIs.
 - Call save_discovery_finding after EVERY interconnection chain you discover.
+- Call save_infrastructure_context ONCE at the very end with the complete consolidated document.
```
beng360, line 115: what happens here if mid investigation it calls it again, it would upsert right? should we check for this or not

server/chat/backend/agent/skills/core/knowledge_base.md
```diff
@@ -16,3 +17,8 @@ knowledge_base_search(query, limit) - Search user's uploaded documentation:
   - 'escalation process database'
 - IMPORTANT: Reference knowledge base findings with source citations in your analysis
 - If a runbook exists for the issue, FOLLOW the documented steps
+
+INFRASTRUCTURE CONTEXT:
+get_infrastructure_context() - Retrieve the full infrastructure context document:
+- Returns a comprehensive document covering environments, services, dependencies, CI/CD, and monitoring. Gives a big picture of overall infrastructure context that can be helpful to understand an orgs setup.
```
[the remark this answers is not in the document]
damianloch: Fixed -- added the possessive apostrophe.

server/chat/backend/agent/tools/infra_context_tool.py
damianloch: Added a 100k char cap, same pattern as the postmortem tool.

server/routes/graph_routes.py
damianloch: Fixed -- removed the redundant `{e}` from the f-string since `logger.exception` already appends the traceback.

server/chat/background/prediscovery_task.py
damianloch: It upserts (`ON CONFLICT DO UPDATE`), so multiple calls are harmless -- later calls just overwrite earlier ones. The prompt instructs the agent to call it once at the end, and if it calls it mid-investigation by mistake, the final call still wins. No guard needed.

beng360 (approved):

PR #433, opened 2026-05-21
fix: github diff tool showing wrong diff

Harrio-6:
The LLM was passing only changed lines instead of the complete file, causing the diff viewer to show everything as deleted. Strengthened the tool description to explicitly require the full file contents.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the GitHub fix tool to require complete file contents in fix suggestions instead of partial snippets, preventing UI and diff display breakage.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Arvo-AI/aurora/pull/433?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


damianloch:
see #427

beng360:
Yep — #427 fixed the code-level bug (relation rename + missing incident_start_time). This one hardens the prompt so the LLM doesn't pass partial file content in the first place (which was the other half of the broken-diff problem).

PR #435, opened 2026-05-21
Add email notifications on action completion

beng360:
## Summary
- Sends email when Aurora Actions complete (success or failure)
- New user preference toggle `action_email_notifications` in Settings > Notifications
- Reuses existing email recipient system (primary email + verified additional recipients)
- Hooks into all action completion paths: success, guardrail block, timeout, and general errors

## Context
DEV-1170 — Users (e.g. Mario) want to know when actions finish without having to watch the UI.

## Test plan
- [ ] Enable `action_email_notifications` toggle in settings
- [ ] Trigger an action manually, verify email arrives on completion
- [ ] Trigger a failing action (e.g. guardrail block), verify error email arrives
- [ ] Verify toggle OFF = no emails sent
- [ ] Verify additional verified email recipients also receive the notification

## Summary by CodeRabbit

* **New Features**
  * New settings toggle to opt into "Action Complete" email notifications (off by default).
  * Action-completion emails sent to primary and verified additional recipients, include status, duration, session link, and optional error details.

* **Improvements**
  * Email settings UI reorganized into separate notification cards and streamlined save flow.
  * Toast messages and error handling in email add/verify/resend/toggle/remove flows updated.

* **Bug Fixes**
  * Timeouts, guardrail blocks, and failures now trigger action-completion emails with error information.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Arvo-AI/aurora/pull/435?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/client/src/components/RCASettings.tsx
+++ b/client/src/components/RCASettings.tsx
@@ -1,12 +1,12 @@
 "use client";
 
-import { useState, useEffect } from "react";
+import { useState, useEffect, useCallback } from "react";
 import { Card, CardContent, CardDescription, CardHeader, CardTitle } from "@/components/ui/card";
 import { Button } from "@/components/ui/button";
 import { Switch } from "@/components/ui/switch";
 import { Input } from "@/components/ui/input";
 import { useToast } from "@/hooks/use-toast";
-import { Bell, Plus, Trash2, Check, Clock, RefreshCw, Mail } from "lucide-react";
+import { Bell, Plus, Trash2, Check, Clock, RefreshCw, Mail, Zap } from "lucide-react";
 import { useAuth, useUser } from "@/hooks/useAuthHooks";
 import {
   Dialog,
@@ -37,16 +37,16 @@ export function RCASettings() {
   });
   const [savingPreferences, setSavingPreferences] = useState<Record<string, boolean>>({});
   const [isLoadingNotificationPref, setIsLoadingNotificationPref] = useState(true);
-  
+
   const [primaryEmail, setPrimaryEmail] = useState<string>("");
   const [additionalEmails, setAdditionalEmails] = useState<RCAEmail[]>([]);
   const [isLoadingEmails, setIsLoadingEmails] = useState(true);
-  
+
   // Add email state
   const [isAddDialogOpen, setIsAddDialogOpen] = useState(false);
   const [newEmail, setNewEmail] = useState("");
   const [isAddingEmail, setIsAddingEmail] = useState(false);
-  
+
   // Verification dialog state
   const [isVerifyDialogOpen, setIsVerifyDialogOpen] = useState(false);
   const [verifyingEmail, setVerifyingEmail] = useState("");
@@ -54,16 +54,15 @@ export function RCASettings() {
   const [isVerifying, setIsVerifying] = useState(false);
   const [isResending, setIsResending] = useState(false);
   const [resendCooldown, setResendCooldown] = useState(0);
-  
+
   const { toast } = useToast();
   const { userId } = useAuth();
   const { user } = useUser();
 
-  // Load notification preferences
   useEffect(() => {
     const loadNotificationPreferences = async () => {
       if (!userId) return;
-      
+
       try {
         const keys = ['rca_email_notifications', 'rca_email_start_notifications', 'action_email_notifications'];
         const loaded: Record<string, boolean> = {};
@@ -93,19 +92,16 @@ export function RCASettings() {
     loadNotificationPreferences();
   }, [userId]);
 
-  // Load email list
   useEffect(() => {
     loadEmails();
   }, [userId]);
 
-  // Set primary email from authenticated user
   useEffect(() => {
     if (user?.emailAddresses?.[0]?.emailAddress) {
       setPrimaryEmail(user.emailAddresses[0].emailAddress);
     }
   }, [user]);
 
-  // Resend cooldown timer
   useEffect(() => {
     if (resendCooldown > 0) {
       const timer = setTimeout(() => setResendCooldown(resendCooldown - 1), 1000);
@@ -113,9 +109,9 @@ export function RCASettings() {
     }
   }, [resendCooldown]);
 
-  const loadEmails = async () => {
+  const loadEmails = useCallback(async () => {
     if (!userId) return;
-    
+
     try {
       setIsLoadingEmails(true);
       const data = await listRCAEmails();
@@ -131,12 +127,11 @@ export function RCASettings() {
     } finally {
       setIsLoadingEmails(false);
     }
-  };
+  }, [userId, toast]);
 
-  const handlePreferenceChange = async (key: string, enabled: boolean, label: string) => {
+  const handlePreferenceChange = async (key: string, enabled: boolean) => {
     if (!userId) return;
-    
-    // Optimistic update for smoother UI
+
     setPreferences(prev => ({ ...prev, [key]: enabled }));
     setSavingPreferences(prev => ({ ...prev, [key]: true }));
 
@@ -145,13 +140,8 @@ export function RCASettings() {
         `/api/proxy/user-preferences`,
         {
           method: 'POST',
-          headers: {
-            'Content-Type': 'application/json',
-          },
-          body: JSON.stringify({
-            key,
-            value: enabled,
-          }),
+          headers: { 'Content-Type': 'application/json' },
+          body: JSON.stringify({ key, value: enabled }),
         }
       );
 
@@ -160,7 +150,6 @@ export function RCASettings() {
       }
     } catch (error) {
       console.error(`Error saving ${key} preference:`, error);
-      // Revert on error
       setPreferences(prev => ({ ...prev, [key]: !enabled }));
       toast({
         title: "Error",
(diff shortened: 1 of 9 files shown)


client/src/components/RCASettings.tsx
damianloch, line 1: I think we need some more effort into the frontend aspect of this. The frontend says RCA notifications section but we have actions notifications under it. Are the same people alerted? What about primary and secondary what happens with those? What if I want different people notified for actions vs RCA...

server/chat/background/task.py
```diff
@@ -701,8 +701,10 @@ def run_background_chat(
                         run_id=trigger_metadata['run_id'], status='error',
                         user_id=user_id, error_message='Action blocked by safety guardrails',
                     )
+                    _send_action_completion_notification(user_id, trigger_metadata, session_id, status='error', error_message='Action blocked by safety guardrails')
```
damianloch: _send_action_completion_notification is inside the same try block as update_action_run_status. If the notification raises (e.g. SMTP timeout), the except at line 708 logs "Failed to update action run status", which is misleading since the status update actually succeeded; it was the email that failed.

server/chat/background/task.py
```diff
@@ -1864,6 +1868,101 @@ def _get_incident_data(incident_id: str, user_id: str) -> Optional[Dict[str, Any
...
+                email_service.send_action_completed_email(recipient, action_data)
+            except Exception as e:
+                logger.warning("[ActionNotification] Failed to send to %s: %s", recipient, e)
+
+        logger.info("[ActionNotification] Sent %s notification for action '%s' to %d recipient(s)", status, action_name, len(all_emails))
+    except Exception as e:
+        logger.error("[ActionNotification] Error sending action completion notification: %s", e)
+
+
+def _is_action_email_notification_enabled(user_id: str) -> bool:
```
damianloch: I think this is just get_user_preference, no need to be a seperate function

damianloch (changes requested):

server/chat/background/task.py
damianloch, line 1: this needs to work in an org context (all users in org share these email settings).

damianloch, line 1: I created a new account and it failed to find my email because i had no connnected oauth providers. 

See cursor explanation: get_user_email only checks the user_tokens table, not the users table. Your email is in users.email but since you have no connected providers yet, there's no row in user_tokens with your email.

damianloch (changes requested):

server/utils/notifications/email_service.py
```diff
@@ -516,6 +516,102 @@ def send_verification_code_email(
...
+            else:
+                duration_str = f'{minutes} minutes'
+
+        is_success = status == 'success'
+        status_label = 'Completed Successfully' if is_success else 'Failed'
+        status_color = '#16a34a' if is_success else '#dc2626'
+
+        subject = f"[Aurora] Action {status_label} - {action_name}"
+
+        session_url = f"{self.frontend_url}/actions?session={session_id}" if session_id else self.frontend_url
```
damianloch: This page doesn't exist its either base actions page or the chat session id where the action ran

damianloch (changes requested):

server/chat/background/task.py
```diff
@@ -1864,6 +1868,101 @@ def _get_incident_data(incident_id: str, user_id: str) -> Optional[Dict[str, Any
...
+    try:
+        if not _is_action_email_notification_enabled(user_id):
+            return
+
+        user_email = get_user_email(user_id)
+        if not user_email:
+            logger.warning("[ActionNotification] No email found for user %s", user_id)
+            return
+
+        action_id = trigger_metadata.get('action_id')
```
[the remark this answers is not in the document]
beng360: Fixed — removed the unused `action_id` assignment.

server/chat/background/task.py
```diff
@@ -1864,6 +1868,101 @@ def _get_incident_data(incident_id: str, user_id: str) -> Optional[Dict[str, Any
...
+        try:
+            with db_pool.get_admin_connection() as conn:
+                with conn.cursor() as cur:
+                    set_rls_context(cur, conn, user_id, log_prefix="[ActionNotification:Emails]")
+                    cur.execute(
+                        "SELECT email FROM rca_notification_emails WHERE user_id = %s AND is_verified = TRUE AND is_enabled = TRUE",
+                        (user_id,),
+                    )
+                    all_emails += [r[0] for r in cur.fetchall()]
+        except Exception:
```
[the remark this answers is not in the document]
beng360: Fixed — now logs a warning with user_id and error details instead of silently passing.

server/chat/background/task.py
```diff
@@ -1864,6 +1868,101 @@ def _get_incident_data(incident_id: str, user_id: str) -> Optional[Dict[str, Any
...
+            pass
```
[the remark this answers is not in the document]
beng360: Fixed — same as above, the except block now logs a warning with user_id and the exception.

server/utils/notifications/email_service.py
```diff
@@ -516,6 +516,102 @@ def send_verification_code_email(
...
+
+                            {error_section}
+
+                            <!-- CTA Button -->
+                            {self._cta_button_html(session_url, "View Action Details")}
+                        </td>
+                    </tr>
+{self._email_footer_html()}"""
+
+        return self._send_email(to_email, subject, html_body, text_body)
```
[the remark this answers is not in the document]
beng360: Fixed — added `import html` and escape `action_name` and `error_msg` with `html.escape()` before inserting into the HTML template. Also fixed the session URL to use the correct `/chat?sessionId=` pattern (and fallback to `/actions`).

client/src/components/RCASettings.tsx
beng360: good call — split it into 3 separate cards now. recipients at the top (shared across all types), then investigation notifications, then action notifications as its own section. way cleaner. if we need separate recipient lists per type later thats easy to add but for now shared makes sense

server/chat/background/task.py
beng360: yep good catch — pulled the notification call out into its own try/except so if smtp times out it doesnt pollute the status update error log

server/chat/background/task.py
beng360: done — just using get_user_preference directly now, killed the wrapper

server/chat/background/task.py
beng360: fixed — switched to get_user_preference which already handles org context through RLS. so all users in the org share the setting now

beng360: ah yeah — get_user_email was only checking user_tokens. fixed it to check users.email first which always has it for credential accounts, then falls back to user_tokens for oauth

server/utils/notifications/email_service.py
beng360: fixed — now links to /chat?sessionId= which is the actual route, falls back to /actions if theres no session

server/utils/auth/stateless_auth.py
```diff
@@ -575,20 +575,26 @@ def get_user_email(user_id: str) -> Optional[str]:
...
+                )
+                result = cursor.fetchone()
+                if result and result[0]:
+                    return result[0]
+
+                # Fallback to user_tokens (OAuth providers store email here)
                 cursor.execute(
                     "SELECT email FROM user_tokens WHERE user_id = %s AND email IS NOT NULL LIMIT 1",
                     (user_id,)
                 )
```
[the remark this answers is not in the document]
beng360: fixed — added set_rls_context before the user_tokens query so it works from celery. the primary path (users table) doesnt need RLS since its not protected, but the fallback does.

server/utils/notifications/email_service.py
```diff
@@ -516,6 +517,102 @@ def send_verification_code_email(
...
+                - started_at: When the action started
+                - completed_at: When the action finished
+                - session_id: Chat session ID (for link)
+
+        Returns:
+            True if email sent successfully, False otherwise
+        """
+        action_name = html.escape(action_data.get('action_name', 'Unknown Action'))
+        status = action_data.get('status', 'success')
+        error_msg = html.escape(action_data.get('error')) if action_data.get('error') else None
```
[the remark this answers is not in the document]
beng360: good catch — split into raw values for subject/plaintext and separate _html vars for the template. no more escaped entities showing up in email subjects.

server/chat/background/task.py
```diff
@@ -1864,6 +1877,84 @@ def _get_incident_data(incident_id: str, user_id: str) -> Optional[Dict[str, Any
...
+        }
+
+        # Send to primary + verified additional emails
+        all_emails = [user_email]
+        try:
+            with db_pool.get_admin_connection() as conn:
+                with conn.cursor() as cur:
+                    set_rls_context(cur, conn, user_id, log_prefix="[ActionNotification:Emails]")
+                    cur.execute(
+                        "SELECT email FROM rca_notification_emails WHERE user_id = %s AND is_verified = TRUE AND is_enabled = TRUE",
```
damianloch: The preference is org-scoped but the recipient resolution is user-scoped. So if User A configured the org's notification recipients, those emails only fire when User A's actions run, not when User B's actions run. We just need it to be a single org state

damianloch (changes requested):

damianloch (approved):

OlivierTrudeau (approved):

PR #439, opened 2026-05-23
fix(rca): prevent worker death from marking completed RCAs as failed

damianloch:
## Summary

- **RCA/prediscovery/action sessions incorrectly marked as "failed"** despite completing successfully. The Celery worker child process exits (exitcode 0) during `asyncio.run()` event loop teardown after `_execute_background_chat` returns. Critical state transitions (`completed_successfully = True`, session status update, incident advancement) were placed 150+ lines after `asyncio.run()` and never reached before process death. This affects all background task types, not only MCP-tool-using sessions — production evidence confirms the same crash on sessions using only `cloud_exec` (no MCP subprocess involved).
- **30-60 seconds of dead wait time** after every RCA that used MCP tools (`mcp_get_commit`, `mcp_get_file_contents`). The `_wait_for_ongoing_tool_calls` function polled for a `completed` flag that MCP tools never set (they only call `send_tool_completion` for WebSocket notifications, not `capture_tool_end`).
- **Health check causing thread starvation** — the `/health` endpoint performed a full LLM WebSocket round-trip, blocking Gunicorn threads and triggering full agent workflows for a fake `health-check` user every ~15 seconds.
- **MCP file contents not reaching the LLM** — `mcp_get_file_contents` response parsing didn't handle `resource`-type content items (only checked for top-level `text` field).
- **LLM citing section headers instead of numeric evidence** — the summarization prompt's "AGENT REASONING" section name was being used as a citation target, silently dropping all real citations.

## Changes

| File | Fix |
|------|-----|
| `server/chat/background/task.py` | (1) Move `completed_successfully = True` to immediately after `asyncio.run()` returns. (2) Add full finalization inside `_execute_background_chat` before teardown: mark session 'completed', clear `rca_celery_task_id`, advance incident to 'summarizing', enqueue post-RCA summarization. (3) Add idempotent guards in outer function: `WHERE rca_celery_task_id IS NOT NULL`, check `aurora_status = 'running'` before advancing. |
| `server/chat/backend/agent/workflow.py` | Reduce `_wait_for_ongoing_tool_calls` timeout from 30s to 5s, force-mark stale MCP tool entries as completed after timeout (MCP tools never call `capture_tool_end`) |
| `server/chat/backend/agent/tools/mcp_tools.py` | Handle `resource.text` and `resource.blob` content types in MCP tool responses (aggregate all content items, not just first) |
| `server/chat/background/summarization.py` | Rename "AGENT REASONING" section to "INVESTIGATOR NOTES" with explicit "do NOT cite" instruction, add CITATION RULES block requiring numeric-only citations |
| `server/routes/health_routes.py` | Replace full chatbot WebSocket LLM call with TCP port check to prevent Gunicorn thread starvation |

## Evidence

From production logs — 3 incidents in 4 days (May 22-23) with the same pattern:
```
19:43:56,740 - Workflow completed normally for session e63252f2
19:43:56,770 - Workflow execution completed - all streams and tool calls finished
19:43:56,792 - Extracted 23 tool calls for visualization
19:43:56,899 - WorkerLostError: exitcode 0 Job: 2623
```

The async function completes, but `asyncio.run()` teardown kills the child process before execution returns to the outer function. With the fix, session status and incident advancement are committed inside the async function before teardown begins.

Note: This is NOT specific to MCP subprocess cleanup. Production crash on May 23 19:43 (session e63252f2) used only `cloud_exec` and `save_discovery_finding` — no MCP tools. The process death is a general `asyncio.run()` teardown issue in billiard (Celery) child processes.

## Test plan

- [ ] Trigger incident → RCA completes → session status is "completed" (not "failed")
- [ ] Check celery logs: no `WorkerLostError` on the RCA task
- [ ] Verify "Timed out after 5s, force-marked N tool calls" appears (instead of 30s dead wait)
- [ ] Health endpoint responds quickly without blocking threads or triggering agent workflows
- [ ] Prediscovery task completes → session marked "completed" (same fix path)
- [ ] Action task completes → session marked "completed", action_run marked "success"
- [ ] Incident summary uses numeric citations only (no [Agent Reasoning] or [Investigator Notes])

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed tool output handling to aggregate all content items, extract nested text, and gracefully handle binary/encoded blobs.
  * Improved tool execution completion detection with faster polling and reliable timeout handling that finalizes stalled calls.

* **Improvements**
  * More robust background task completion and incident lifecycle handling to reduce false-running states and ensure post-run updates.
  * Simplified chatbot health check to a lightweight TCP probe.
  * Refined incident-summary prompts to better separate investigator notes from evidence.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Arvo-AI/aurora/pull/439?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/server/chat/background/task.py
+++ b/server/chat/background/task.py
@@ -662,22 +662,28 @@ def run_background_chat(
                 logger.error(f"[BackgroundChat] Failed to determine severity: {e}")
             
             # Regenerate incident summary (skip if already enqueued by inner function)
+            should_enqueue = False
             try:
                 with db_pool.get_admin_connection() as conn:
                     with conn.cursor() as cursor:
                         set_rls_context(cursor, conn, user_id, log_prefix="[BackgroundChat:SumCheck]")
                         cursor.execute("SELECT aurora_status FROM incidents WHERE id = %s", (incident_id,))
                         row = cursor.fetchone()
-                        if row and row[0] != 'summarizing':
-                            from chat.background.summarization import generate_incident_summary_from_chat
-                            generate_incident_summary_from_chat.delay(
-                                incident_id=incident_id,
-                                user_id=user_id,
-                                session_id=session_id,
-                            )
-            except Exception as e:
-                logger.error(f"[BackgroundChat] Failed to enqueue post-RCA summarization for incident {incident_id}: {e}")
-                _update_incident_aurora_status(incident_id, "complete", user_id=user_id)
+                        should_enqueue = bool(row and row[0] != 'summarizing')
+            except Exception:
+                logger.warning("[BackgroundChat] Failed to check summarization state for incident %s", incident_id)
+
+            if should_enqueue:
+                try:
+                    from chat.background.summarization import generate_incident_summary_from_chat
+                    generate_incident_summary_from_chat.delay(
+                        incident_id=incident_id,
+                        user_id=user_id,
+                        session_id=session_id,
+                    )
+                except Exception:
+                    logger.exception("[BackgroundChat] Failed to enqueue post-RCA summarization for incident %s", incident_id)
+                    _update_incident_aurora_status(incident_id, "complete", user_id=user_id)
             
             # Generate final complete visualization
             try:
--- a/server/chat/background/task.py
+++ b/server/chat/background/task.py
@@ -661,14 +661,20 @@ def run_background_chat(
             except Exception as e:
                 logger.error(f"[BackgroundChat] Failed to determine severity: {e}")
             
-            # Regenerate incident summary now that RCA chat has completed
+            # Regenerate incident summary (skip if already enqueued by inner function)
             try:
-                from chat.background.summarization import generate_incident_summary_from_chat
-                generate_incident_summary_from_chat.delay(
-                    incident_id=incident_id,
-                    user_id=user_id,
-                    session_id=session_id,
-                )
+                with db_pool.get_admin_connection() as conn:
+                    with conn.cursor() as cursor:
+                        set_rls_context(cursor, conn, user_id, log_prefix="[BackgroundChat:SumCheck]")
+                        cursor.execute("SELECT aurora_status FROM incidents WHERE id = %s", (incident_id,))
+                        row = cursor.fetchone()
+                        if row and row[0] != 'summarizing':
+                            from chat.background.summarization import generate_incident_summary_from_chat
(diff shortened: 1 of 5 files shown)


server/chat/background/task.py
```diff
@@ -1319,6 +1335,40 @@ async def _execute_background_chat(
...
+                        set_rls_context(cursor, conn, user_id, log_prefix="[BackgroundChat:Finalize]")
+                        cursor.execute(
+                            "UPDATE incidents SET rca_celery_task_id = NULL WHERE id = %s",
+                            (incident_id,)
+                        )
+                        conn.commit()
+            except Exception as e:
+                logger.warning(f"[BackgroundChat] Failed to clear task ID: {e}")
+
+            _update_incident_status_and_aurora(incident_id, "analyzed", "summarizing", user_id=user_id)
```
OlivierTrudeau, line 1365: Why do we have like this many fallbacks. Isn't this like the 3rd fallback to make sure we mark sessions as done?

OlivierTrudeau (dismissed):

server/chat/background/task.py
damianloch: Its only first fallback, the finally block is separate and it only marks things as failed if status is still in_progress. The 2 cases are inside the async function (what we added in this PR) and after async.io returns as a safety net.

server/chat/background/task.py
```diff
@@ -650,14 +661,20 @@ def run_background_chat(
...
+                        if row and row[0] != 'summarizing':
+                            from chat.background.summarization import generate_incident_summary_from_chat
+                            generate_incident_summary_from_chat.delay(
+                                incident_id=incident_id,
+                                user_id=user_id,
+                                session_id=session_id,
+                            )
             except Exception as e:
                 logger.error(f"[BackgroundChat] Failed to enqueue post-RCA summarization for incident {incident_id}: {e}")
                 _update_incident_aurora_status(incident_id, "complete", user_id=user_id)
```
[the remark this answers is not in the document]
damianloch: Good catch — split the try blocks so a transient DB read failure no longer triggers the "mark complete" fallback. Only an actual enqueue failure does now. Fixed in c03eb6fc.

OlivierTrudeau (approved):

PR #455, opened 2026-05-28
fix(gcp): SA-mode project disable now persists and gates agent access

isiddharthsingh:
## Summary

Fixes the GCP connector's broken project enable/disable in service-account mode, plus the follow-on issues that surfaced while testing it end-to-end.

### What was broken
- SA-mode project toggle didn't persist: POST was a no-op, GET hardcoded `enabled: true`. Toggling visibly reverted on reload.
- After disabling a project, the chat agent still queried it because `gcloud` talks to GCP's API directly.
- The SA-mode project picker always defaulted to `default_project_id` from the uploaded SA key. If the user disabled that specific project, every gcloud command got `--project=<disabled-default>` injected and was hard-blocked.
- Clicking **Set as Root** updated the backend but the modal kept showing the old root because `loadProjects()` hit the stale `ProjectCache`.

### What this PR does
- **Persist SA enable/disable** as `gcp_sa_disabled_projects` user pref. GET reflects saved state; discovery skips disabled projects (`server/services/discovery/tasks.py`).
- **Agent enforcement, two layers** so disabling actually keeps the agent out of a project:
  - Prompt: `provider_rules.build_provider_context_segment` appends a *"these GCP projects are off-limits"* block when the user is in SA mode and has a non-empty disabled list.
  - Wrapper guard: `cloud_exec` refuses any GCP command whose final form mentions a disabled project ID (whole-token regex; only runs the DB check when `auth_method == 'service_account'`).
- **Root project drives the active project in SA mode**: `setup_gcp_environment_isolated` now reads `gcp_root_project` (set via the existing "Set as Root" button) and uses it ahead of the SA key's `default_project_id`. This is what makes the button actually fix the auto-injection deadlock when the SA's bundled default is disabled.
- **Frontend force-refresh** after Set as Root in `GcpProviderIntegration` so the modal reflects the new root immediately instead of waiting for a manual Refresh click.
- **Misc safety fixes** picked up during review:
  - Hash `user_id` in the cloud_exec block log line.
  - Validate the `projects` POST payload is a list of dicts (returns 400 instead of 500 on malformed input).
  - `get_gcp_disabled_projects` now logs exceptions via `logger.exception` instead of swallowing silently; intentionally still fails open so a transient DB blip doesn't crash chat (cloud_exec guard remains the hard-enforcement layer).
  - Gate the prompt's disabled-projects block on `_is_gcp_service_account_mode(user_id)` so a stale pref left over from a SA→OAuth switch doesn't falsely restrict OAuth-mode access.
  - Removed three function-local `import re` statements in `cloud_exec` that were shadowing the module-level import and produced `UnboundLocalError: cannot access free variable 're'` once the new guard started using `re`.

### Commits (chronological)
1. `94d5cef8` — persist SA-mode enable/disable + apply in discovery
2. `88267097` — block disabled projects in agent prompt and cloud_exec guard
3. `8bcadfda` — restore Set as Root in SA mode + force-refresh after selection
4. `24e89626` — honor `gcp_root_project` pref when picking active project in SA mode
5. `67ecd225` — review-comment sweep (SA-mode gate on prompt block, hashed user_id, payload validation, `logger.exception`)

(`bf24b335`, the initial "hide root controls in SA mode" commit, is superseded by `8bcadfda` — keeping it in the history rather than rewriting since the PR is already pushed.)

## Test plan
- [ ] Disable a project, reload the modal — toggle stays off.
- [ ] Re-enable, reload — toggle stays on.
- [ ] Disable a project, ask the agent about it — agent refuses with `GCP_PROJECT_DISABLED`; re-enable, agent answers normally.
- [ ] Connect via SA key whose default project you then disable; click Set as Root on a different enabled project; run a gcloud command — agent targets the new root, no `--project=<disabled>` injected.
- [ ] Click Set as Root → modal updates the root badge immediately (no manual Refresh needed).
- [ ] OAuth-mode user: existing gcloud flows unchanged; the wrapper guard short-circuits on `auth_method`.
- [ ] Malformed POST to `/api/gcp/sa-project-access` returns 400, not 500.

## Summary by CodeRabbit

* **New Features**
  * Added root project/subscription selection across GCP, Azure, OVH, and Scaleway cloud providers
  * Enhanced project management with unified root-selection interface

* **Improvements**
  * Updated GCP service account setup documentation with detailed IAM role guidance
  * Improved in-app documentation for auto-discovering projects and configuring cloud resource access

* **Changes**
  * Service account project toggles now conditionally hidden based on authentication type

[Review Change Stack](https://app.coderabbit.ai/change-stack/Arvo-AI/aurora/pull/455?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/client/src/components/cloud-provider/projects/useSetAsRoot.ts
+++ b/client/src/components/cloud-provider/projects/useSetAsRoot.ts
@@ -1,0 +1,46 @@
+import { useState, useCallback } from 'react';
+import { useToast } from '@/hooks/use-toast';
+import { setRootProject } from './projectUtils';
+
+interface UseSetAsRootResult {
+  setAsRoot: (providerId: string, projectId: string) => Promise<void>;
+  isSaving: boolean;
+}
+
+export function useSetAsRoot(
+  userId: string | null,
+  refresh: () => Promise<void>,
+  resourceLabel: string = 'project',
+): UseSetAsRootResult {
+  const [isSaving, setIsSaving] = useState(false);
+  const { toast } = useToast();
+
+  const setAsRoot = useCallback(
+    async (providerId: string, projectId: string) => {
+      if (!userId) return;
+      setIsSaving(true);
+      try {
+        await setRootProject(providerId, projectId);
+        await refresh();
+        toast({
+          title: 'Success',
+          description: `Root ${resourceLabel} updated successfully`,
+        });
+      } catch (error: unknown) {
+        const message =
+          error instanceof Error ? error.message : `Failed to set root ${resourceLabel}`;
+        console.error(`Error setting root ${resourceLabel}:`, error);
+        toast({
+          title: 'Error',
+          description: message,
+          variant: 'destructive',
+        });
+      } finally {
+        setIsSaving(false);
+      }
+    },
+    [userId, refresh, toast, resourceLabel],
+  );
+
+  return { setAsRoot, isSaving };
+}
--- a/client/src/components/connectors/GcpServiceAccountForm.tsx
+++ b/client/src/components/connectors/GcpServiceAccountForm.tsx
@@ -276,12 +276,24 @@ export function GcpServiceAccountForm({ onSuccess }: GcpServiceAccountFormProps)
 
       <Alert>
         <AlertTitle className="text-sm">Auto-discover all your projects</AlertTitle>
-        <AlertDescription className="text-xs text-muted-foreground">
-          To auto-discover all your projects with one key, grant the service
-          account <code className="font-mono">roles/browser</code> (or{" "}
-          <code className="font-mono">roles/viewer</code>) at the organization
-          or folder level in GCP IAM. Otherwise only the project this key was
-          created in will be connected.
+        <AlertDescription className="text-xs text-muted-foreground space-y-2">
+          <p>
+            Grant the service account{" "}
+            <code className="font-mono">roles/browser</code> at the{" "}
+            <strong>organization</strong> or <strong>folder</strong> level so
+            Aurora can list every project in that scope. Without this, only the
+            project the key was created in shows up.
+          </p>
+          <p>
+            <code className="font-mono">roles/browser</code> only lets Aurora{" "}
+            <em>see</em> the projects. To actually investigate one, also grant
+            viewer-tier roles on that project —{" "}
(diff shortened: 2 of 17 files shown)


server/chat/backend/agent/prompt/provider_rules.py
damianloch, line 1: We spent lots of time pruning the prompt from noise like this, this cannot go in prompt.

client/src/components/cloud-provider/projects/useSetAsRoot.ts
damianloch, line 1: why is it that there can be a root project but doesn't need to be. The UI can have no root project set and then our backend will fall back to one, but then why not just show what will be fallen back to as root project then. I feel as if our logic (existing and new) is too lenient and error prone

damianloch (changes requested):

client/src/components/cloud-provider/projects/useSetAsRoot.ts
isiddharthsingh: We can do it in backend but sometime it can create an issue when it starts fallback on a different project and gets picked most of the time.

client/src/components/gcp-provider-integration.tsx
```diff
@@ -253,7 +219,7 @@ export default function GcpProviderIntegration({ onDisconnect }: GcpProviderInte
                 isLoading={togglingProjectId === project.projectId}
                 onToggle={handleToggle}
                 onSetAsRoot={handleSetAsRoot}
-                showToggle={true}
+                showToggle={authType !== 'service_account'}
               />
```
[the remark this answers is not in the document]
isiddharthsingh: yes that was the main behavior we wanted.

server/chat/backend/agent/prompt/provider_rules.py
isiddharthsingh: removed

client/src/components/connectors/GcpServiceAccountForm.tsx
damianloch, line 1: Actually one small ui thing here, can you make service account the default and have it say "recommended" when a user click on "connect" for gcp provider .That way they are more likely to go thru service account instead of manually switching to oauth mode.

damianloch (changes requested):

damianloch (approved):

PR #464, opened 2026-06-02
fly.io connector

OlivierTrudeau:
Fly.io connector which connects to fly.io through cli and api for metrics

## Summary by CodeRabbit

* **New Features**
  * Fly.io connector: connect/disconnect with API token + org slug; view live status, permission tier, and app list
  * Agent integrations: run Fly.io CLI via agent, new Fly.io agent skill, and a structured Prometheus metrics query tool
  * Frontend: Fly.io auth page, connector registration, proxy routing, and Fly.io command/logo display

* **Chores**
  * Server: flyctl installed, new Fly.io routes, CORS and secret/provider lists, and policy updates for Fly.io command allow-rules


client/src/app/flyio/auth/page.tsx
```diff
@@ -0,0 +1,339 @@
...
+  const [error, setError] = useState<string | null>(null);
+  const [apiToken, setApiToken] = useState("");
+  const [orgSlug, setOrgSlug] = useState("");
+  const [status, setStatus] = useState<FlyioStatus | null>(null);
+  const { toast } = useToast();
+
+  const checkStatus = useCallback(async () => {
+    setIsCheckingStatus(true);
+    try {
+      const response = await fetch(`/api/proxy/flyio/status?validate=true`);
```
damianloch: checkStatus unconditionally passes ?validate=true, which forces a live Fly.io API call every time the page mounts. If the Fly.io API is slow or down, the page hangs on the spinner. Other connectors (Cloudflare) only do this on demand.

server/utils/auth/command_policy.py
damianloch, line 1: migration needed

damianloch (changes requested):
Only did partial review until cloud_exec_tool.py. Tested connector on ui

client/src/app/flyio/auth/page.tsx
```diff
@@ -0,0 +1,339 @@
...
+          app_names: data.app_names,
+        });
+      } else {
+        setStatus({ connected: false });
+      }
+    } catch {
+      setStatus({ connected: false });
+    } finally {
+      setIsCheckingStatus(false);
+    }
```
isiddharthsingh: The status-check logic (in the block that calls fetch and uses setStatus/setIsCheckingStatus) must also sync localStorage.isFlyioConnected: when the response shows connected:true set localStorage.isFlyioConnected = "true", and when response is not ok or catches/returns connected:false ensure localStorage.isFlyioConnected is removed or set to "false" so other UI won't read a stale "true"; update this in the same branches where setStatus({ connected: ... }) is called (the response.ok branch, the else branch, and the catch/finally handling).

server/connectors/flyio_connector/api_client.py
```diff
@@ -0,0 +1,88 @@
...
+            response = requests.post(
+                f"{FLYIO_MACHINES_API}/apps",
+                headers=self._headers,
+                json={"app_name": "", "org_slug": self.org_slug},
+                timeout=10,
+            )
+            return response.status_code in (400, 422)
+        except Exception as e:
+            logger.warning(f"Fly.io write access probe failed: {e}")
+            return False
```
isiddharthsingh, line 61: The current has_write_access method uses a fragile status-code heuristic (treating 400/422 as "write") which can misclassify tokens; change it to a definitive probe: either call a Fly endpoint that explicitly exposes token scopes (if available) or attempt a minimal valid write operation that will fail with 403 for readonly tokens (e.g., POST to f"{FLYIO_MACHINES_API}/apps" using a syntactically valid random app_name and org_slug in self._headers) and treat 201/202 as write, 403 as readonly, and 4xx/5xx/timeouts as inconclusive (return None or raise a specific ProbeError instead of assuming readonly); update has_write_access to use FLyIO_MACHINES_API, self.org_slug and self._headers and add retry/backoff for transient 5xx/timeouts.

server/routes/flyio/__init__.py
```diff
@@ -0,0 +1,17 @@
...
+
+flyio_bp = Blueprint('flyio', __name__)
+
+
+@flyio_bp.before_request
+def handle_options_request():
+    """Handle CORS preflight OPTIONS requests for all Fly.io routes."""
+    if request.method == 'OPTIONS':
+        return create_cors_response()
+    return None
```
isiddharthsingh: should not add manual OPTIONS handling because the RBAC decorator handles it. The RBAC test passes, but this will likely create review churn and should be removed unless there is a specific preflight bug.

server/connectors/flyio_connector/api_client.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+class FlyioClient:
+    """REST client for Fly.io Machines API and Prometheus federation endpoint."""
+
+    def __init__(self, api_token: str, org_slug: str):
+        self.api_token = api_token
+        self.org_slug = org_slug
+
+        auth_value = api_token if api_token.startswith("FlyV1") else f"Bearer {api_token}"
```
isiddharthsingh, line 25: sends raw FlyV1... tokens without the Bearer prefix. Fly’s Machines API docs show requests use Authorization: Bearer ${FLY_API_TOKEN} ([source](https://fly.io/docs/machines/api/working-with-machines-api/)). This makes normal Fly tokens fail connect, live status, and Prometheus queries.

client/public/flyio.svg
isiddharthsingh, line 1: this is not the correct one

isiddharthsingh:
it does not show connected like gcp, aws,...
also on connecting i get 
image

isiddharthsingh:
the app number can be misleading. There are 3 states on fly, so maybe we can, if there is any ability to use those and show it here. 

image
image

client/public/flyio.svg
OlivierTrudeau: It's from their docs🥲

server/routes/flyio/flyio_routes.py
```diff
@@ -0,0 +1,147 @@
...
+        app_names = token_data.get("app_names", [])
+
+        if request.args.get("validate", "").lower() == "true":
+            api_token = token_data.get("api_token")
+            if not api_token or not org_slug:
+                return jsonify({"connected": False}), 200
+
+            apps = FlyioClient(api_token, org_slug).list_apps()
+            if apps is None:
+                return jsonify({"connected": False, "reason": "token_invalid"}), 200
```
isiddharthsingh, line 117: When validate=true and the token no longer works, the route returns connected:false but leaves the persisted connection status as connected (and the stored secret intact). Subsequent non-validated status calls will keep reporting connected:true, diverging from reality. Consider marking the connection disconnected (or otherwise flagging it) when apps is None here.

server/utils/auth/command_policy.py
```diff
@@ -749,6 +752,9 @@ def get_policy_templates() -> List[dict]:
                 # Tailscale
                 {"priority": 140, "pattern": r"^tailscale\s+(status|device\s+(list|get)|dns|acl\s+(get|show)|routes|settings|auth-key\s+list)\b",
                  "description": "Tailscale read-only operations"},
+                # Fly.io
+                {"priority": 139, "pattern": r"^(fly|flyctl)\s+(apps\s+list|status|machines?\s+(list|status)|logs|checks|releases|certs\s+list|ips\s+list|volumes?\s+list|scale\s+show|regions|platform)\b",
+                 "description": "Fly.io read-only operations"},
```
isiddharthsingh: The Standard Operations Fly.io rule (the dict with "priority": 139 and pattern r"^(fly|flyctl)\s+(apps\s+list|status|machines?\s+(list|status)|logs|checks|releases|certs\s+list|ips\s+list|volumes?\s+list|scale\s+show|regions|platform)\b") currently duplicates the Observability Only pattern; update that pattern in command_policy.py so Standard Operations permits additional commands (e.g., include alternatives for ssh, proxy, configure, auth and other management verbs) by expanding the regex to allow tokens like "ssh", "proxy", "configure", "auth" (and any other management commands we want to allow) after the fly/flyctl prefix while preserving existing allowed verbs and the same "priority": 139 and "description": "Fly.io read-only operations" (adjust description if needed).

server/Dockerfile
```diff
@@ -206,6 +206,9 @@ RUN ARCH=$(uname -m) && \
 # Install Tailscale CLI (for local dev SSH to tailnet devices)
 RUN curl -fsSL https://tailscale.com/install.sh | sh
 
+# Install Fly.io CLI (flyctl) for Fly.io connector
+RUN curl -fsSL "https://fly.io/install.sh" | FLYCTL_INSTALL="/usr/local" sh
```
isiddharthsingh: Consider pinning the Fly.io CLI version for reproducible builds.

client/public/flyio.svg
isiddharthsingh: Oh yes, you are right. I referred to the favicon on their page, and it was a bit different, so that's why I wrote it.

server/connectors/flyio_connector/api_client.py
OlivierTrudeau: Yeah but tested afterwards and didn't work with the bearer prefix...

server/Dockerfile
OlivierTrudeau: Done

client/src/app/flyio/auth/page.tsx
OlivierTrudeau: Feel free to push back here, but I feel like this is good. Like for the list of all connectors, it doesn't check. however, when you are on the explicit fly.io page, it verifies that your connector actually works. Isn't that good?

server/routes/flyio/flyio_routes.py
OlivierTrudeau: Done, now deletes the credentials

server/utils/auth/command_policy.py
OlivierTrudeau: Done

server/routes/flyio/__init__.py
OlivierTrudeau: Okok removed it

server/utils/auth/command_policy.py
OlivierTrudeau: Made sure the error message is much clearer instead. I don't think our existing users care about fly.io and for the new orgs, they will have the template with fly.io included

client/src/app/flyio/auth/page.tsx
OlivierTrudeau: done

server/connectors/flyio_connector/api_client.py
OlivierTrudeau: Honestly kind of valid to do it my way. The thing is you can't even create a read only on the ui, only through cli. But anyways this works and changing it would mean we would make changes to their infra temporarily which is sort of worst. Anyways, not really that crucial to know if read only or not

OlivierTrudeau:
> the app number can be misleading. There are 3 states on fly, so maybe we can, if there is any ability to use those and show it here.
> 
> image image

Done

isiddharthsingh (approved):

damianloch (approved):

PR #466, opened 2026-06-02
Improve Aurora MCP tool discovery & steering

isiddharthsingh:
## Why

External agents (Cursor, Claude Desktop/Code, Codex, Windsurf) connecting to Aurora's self-hosted MCP server consistently **over-used `chat_with_aurora` / `ask_incident`** for questions a cheap direct read could answer, and **failed to discover obvious read-only tools** (infrastructure context, service graph, RCA findings). The steering text was chat-maximalist in three layers, the high-value reads were buried in `search_tools`-only Tier 3, and the per-sub-agent RCA findings weren't on the MCP surface at all. Result: wrong tool choices and ~40s latency where a ~1s direct read would do.

The tiered architecture is correct (keeps the upfront surface under client tool-count caps). This keeps it and fixes the problem with **(a)** re-steering the prompt text toward direct reads, **(b)** promoting a curated set of reads to the always-visible surface, and **(c)** fixing the `search_tools` matcher so the long tail is findable.

## What changed

**1. Re-steered the steering text (chat-first → direct-read-first)**
- `mcp_server.py` `instructions` — routing guide naming direct-read categories; critical sentence front-loaded; phantom `github_rca` scrubbed.
- `chat_with_aurora` docstring — dropped "Default tool for any question", added a "NOT a structured-data lookup" scope note, trimmed under the 2KB cap (was being truncated). `ask_incident` points at the direct readers first.
- Tier-2 `query_logs`/`query_metrics` descriptions softened; `search_tools` docstring + hint made loud about the long tail.
- `prompts.py` reordered to direct reads first; `blast_radius_analysis` repointed to the new `service_impact` tool.

**2. Promoted 6 curated reads to first-class Tier-1 tools**
`get_infrastructure_context`, `list_services`, `service_impact`, `incident_findings`, `incident_finding_detail`, `incident_list_alerts`. The 3 graph entries were removed from `DISPATCH_ALLOWLIST` (no double-exposure); findings are new. `service_impact` URL-encodes the service name itself.

**3. Fixed the `search_tools` matcher (two bugs)**
- Tokenized the query so multi-word searches match (the old whole-string matcher returned nothing for e.g. `"rca tools steps"` and forced a chat fallback); ranks by match count.
- Added a query-only fallback when a category/connector filter zeroes out a real token match (an LLM guessing `category="metrics"` for an LLM-usage question whose tools live under `category="usage"` previously got zero results and gave up).

**4. Expanded connector coverage** — added skill-gated read entries for Jenkins/CloudBees/Spinnaker (CI/CD), Sentry, and Grafana. Removed the phantom `github_list_repos` entry (`/github/repos` 404s).

**Docs** — updated `website/docs/integrations/mcp.md` (tier tables, routing note, per-client setup caveats) and added `docs/design/mcp-agent-steering.md`.

## Testing
- **MCP unit tests**: 54 pass (`docker exec aurora-server python -m pytest tests/mcp/ -q`).
- **Live MCP end-to-end** against the running server: all 6 promoted tools present + returning real data; `search_tools`/`call_tool` discovery, connector-gating, and the allowlist boundary all behave correctly; the wrong-category fallback resolves the reported empty-result case.
- **Tool-selection hit-rate eval**: factual prompts route to direct tools (0 chat-funnels); only open-ended prompts escalate to chat. Holds both with `instructions` and descriptions-only.
- **Latency**: a factual lookup that took ~31s via chat is ~0.01-0.7s via the direct tool.

## Note for reviewers
The CLAUDE.md New Connector Checklist and the architectural RBAC test do **not** apply here — these are MCP tools wrapping already-RBAC-decorated REST routes, not new connectors/routes.

## Summary by CodeRabbit

* **New Features**
  * Promoted several infrastructure and incident-analysis reads to always-available tools.
  * Added CI/CD and monitoring integrations (Jenkins, CloudBees, Spinnaker, Sentry, Grafana).
  * Improved search discovery: tokenized ranking, visible/callable entries prioritized, and graceful fallback when filters return no results.
  * Expanded tool hints to clarify callable-now requirements.

* **Bug Fixes**
  * Treat upstream no-content (204) responses as successful proxy calls.

* **Documentation**
  * Clarified when to use direct tools vs. assistant-driven investigations; updated examples.

* **Tests**
  * Expanded coverage for registry, search ranking, routing, and tier-1 tool behavior.


server/tests/mcp/test_mcp_registry.py
```diff
@@ -37,6 +40,190 @@ def test_search_dispatch_entries_query_filter():
...
+def test_action_reads_promoted_not_in_allowlist():
+    """The action reads are first-class Tier-1 tools, not dispatch entries."""
+    names = {e.name for e in registry.DISPATCH_ALLOWLIST}
+    for read in ("actions_list", "action_get", "action_list_runs"):
+        assert read not in names, f"{read} should be promoted, not in allowlist"
+    fake = FakeMCP()
+    api_call, _ = make_captured_api_call()
+    register_tier1_tools(fake, api_call)
+    for tool in ("list_actions", "get_action", "list_action_runs"):
+        assert tool in fake.tools, tool
```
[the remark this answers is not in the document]
isiddharthsingh: These names are intentional — not a mistake. The two assertion sets in this test guard different things:

### (1) the pre-promotion DISPATCH names must stay out of the allowlist
for read in ("actions_list", "action_get", "action_list_runs"):
    assert read not in names
### (2) the Tier-1 tool names must be registered as first-class tools
for tool in ("list_actions", "get_action", "list_action_runs"):
    assert tool in fake.tools

actions_list / action_get / action_list_runs are the dispatch-convention names these reads carried before they were promoted to Tier-1 (matching siblings like postmortem_list, incident_list_recent_unlinked). Asserting they're absent from DISPATCH_ALLOWLIST is the meaningful regression guard: it catches the realistic mistake of re-adding an actions read to the allowlist under that naming convention, which would double-expose it.

Switching set (1) to the Tier-1 names (list_actions, etc.) would make the assertion vacuously true — those names were never dispatch entries, so they can't leak into the allowlist. That guard is already covered by set (2), which verifies the Tier-1 names are registered.

So the suggested change would weaken the test (turn a real double-exposure guard into a no-op) rather than fix it. Leaving as-is.

website/docs/integrations/mcp.md
damianloch, line 1: these docs are supposed to be for a user setting up aurora but it sounds like this message is to a bot on how to use MCP

server/aurora_mcp/tools_always_on.py
```diff
@@ -281,3 +371,99 @@ async def search_runbooks(query: str, limit: int = 5) -> Dict[str, Any]:
...
+        """Get one Aurora action's full config plus its 20 most recent runs.
+
+        Args:
+          action_id: the action UUID (from list_actions).
+        """
+        return await _do_get_action(api_call, action_id)
+
+    @mcp.tool()
+    async def list_action_runs(
+        action_id: str, limit: int = 50, offset: int = 0,
```
damianloch, line 463: The docstring says limit: max runs (1-200, default 50) but no clamping is done so a caller could pass limit=999999. The damage is bounded by the backend API (which presumably has its own limits), but the contract is misleading.

damianloch (changes requested):

damianloch (approved):

PR #475, opened 2026-06-04
fix: exclude github_fix tool from action-triggered background chats

damianloch:
## Summary
- `github_fix` saves fix suggestions to `incident_suggestions` for user review in the incident card UI
- In Actions (which run autonomously after RCA completes), no user is actively viewing the incident card — suggestions dead-end silently
- This PR gates `github_fix` to only be available during actual RCA sessions (`source != 'action'`), so actions will instead use the GitHub MCP tools (`create_pull_request`, `push_files`) that produce visible artifacts like PRs

## Context
The "Mute noisy alerts" action was calling `github_fix` repeatedly, generating code suggestions that no one ever sees or acts on. With `github_fix` excluded from actions, the agent will fall through to the MCP tools which can create branches and PRs directly.

## Test plan
- [x] Trigger a background RCA (via Datadog/PagerDuty webhook) — verify `github_fix` is still available
- [x] Trigger an action (on_incident after_rca) — verify `github_fix` is NOT in the tool list and the agent uses MCP tools instead
- [x] Verify cache key differentiates action vs RCA contexts (no stale tool list served)
- [x] Chat-triggered RCA (trigger_rca tool) — verify `github_fix` is available in the spawned background session

## Summary by CodeRabbit

* **Improvements**
  * Tool caching now considers investigation context type to ensure appropriate tool sets are available.
  * GitHub integration tool availability is now contextually determined based on the current investigation mode, improving tool accuracy and preventing incompatible tool combinations.

### DIFF
--- a/server/chat/backend/agent/tools/cloud_tools.py
+++ b/server/chat/backend/agent/tools/cloud_tools.py
@@ -978,6 +978,25 @@ from .mcp_tools import (
     LANGCHAIN_TOOLS_CACHE_DURATION
 )
 
+
+_NON_RCA_SOURCES = frozenset(('action', 'prediscovery'))
+
+
+def _is_background_rca(state_context, is_background: bool) -> bool:
+    """True when the session is a background RCA investigating a real incident.
+
+    Used to gate tools like github_fix that only make sense during incident
+    investigation (where the user will see the incident card afterwards).
+    """
+    if not state_context or not is_background:
+        return False
+    incident_id = getattr(state_context, 'incident_id', None)
+    if not incident_id:
+        return False
+    rca_source = ((getattr(state_context, 'rca_context', None) or {}).get('source', '') or '').lower()
+    return bool(rca_source) and rca_source not in _NON_RCA_SOURCES
+
+
 def get_cloud_tools():
     """Get all cloud management tools including both Aurora native tools and REAL MCP tools."""
     # Import required classes at function start to avoid scope issues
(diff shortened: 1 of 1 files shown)


beng360 (commented):
Minor: `_rca_source_for_cache` and `_incident_for_cache` duplicate the same computation done at line 1331. Consider extracting `is_rca_context` once and reusing in both the cache key and the tool gate.

server/chat/backend/agent/tools/cloud_tools.py
```diff
@@ -1009,10 +1009,13 @@ def get_cloud_tools():
     rca_flag = getattr(state_context, 'trigger_rca_requested', False) if state_context else False
     is_background = getattr(state_context, 'is_background', False) if state_context else False
     is_postmortem_action = getattr(state_context, 'is_postmortem_action', False) if state_context else False
+    _rca_source_for_cache = (getattr(state_context, 'rca_context', None) or {}).get('source', '') if state_context else ''
```
beng360: Minor: `_rca_source_for_cache` and `_incident_for_cache` duplicate the same computation done at line 1331. Consider extracting `is_rca_context` once and reusing in both the cache key and the tool gate.

server/chat/backend/agent/tools/cloud_tools.py
```diff
@@ -1009,10 +1009,13 @@ def get_cloud_tools():
...
+    _incident_for_cache = getattr(state_context, 'incident_id', None) if state_context else None
+    is_rca_context = is_background and bool(_incident_for_cache) and _rca_source_for_cache not in ('action', 'prediscovery', '')
```
beng360: The empty-string exclusion is a safe guard but can't be triggered in practice since `_build_rca_context` validates against `_RCA_SOURCES`. But if `_RCA_SOURCES` ever gets out of sync or isn't updated when a new source is added, this would silently exclude `github_fix` from that source's RCAs with no error.

server/chat/backend/agent/tools/cloud_tools.py
damianloch: Done — extracted `_is_background_rca()` helper that both the cache key and tool registration now call.

server/chat/backend/agent/tools/cloud_tools.py
beng360: 👀

damianloch: Good catch. Flipped the logic: now uses `bool(rca_source) and rca_source not in _NON_RCA_SOURCES`. New sources added to `_RCA_SOURCES` will automatically get `github_fix` without touching this code.

server/chat/backend/agent/tools/cloud_tools.py
```diff
@@ -1009,10 +1009,13 @@ def get_cloud_tools():
...
     if tool_capture is None:
-        cache_key = f"{user_id}:nocapture:{mode_suffix}:background={is_background}:rca={rca_flag}:postmortem={is_postmortem_action}"
+        cache_key = f"{user_id}:nocapture:{mode_suffix}:background={is_background}:rca={rca_flag}:postmortem={is_postmortem_action}:is_rca_ctx={is_rca_context}"
     else:
-        cache_key = f"{user_id}:capture:{id(tool_capture)}:{mode_suffix}:background={is_background}:rca={rca_flag}:postmortem={is_postmortem_action}"
+        cache_key = f"{user_id}:capture:{id(tool_capture)}:{mode_suffix}:background={is_background}:rca={rca_flag}:postmortem={is_postmortem_action}:is_rca_ctx={is_rca_context}"
```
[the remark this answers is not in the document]
damianloch: Done in 9e44768 — extracted `_is_background_rca()` helper with a `_NON_RCA_SOURCES` frozenset. Single predicate used in both cache key and tool registration.

server/chat/backend/agent/tools/cloud_tools.py
beng360: 👀 🏇

beng360 (approved):

PR #479, opened 2026-06-05
Artifacts: persistent agent-maintained documents in Monitor

isiddharthsingh:
## Summary

Adds **Artifacts** — living markdown documents Aurora maintains across runs (findings lists, cost reports, runbooks), surfaced in a new **Monitor → Artifacts** tab. Scheduled Actions and chat create/update them through agent tools; users can view, edit, and browse version history in the UI.

## Backend

- `artifacts` + `artifact_versions` tables with per-org RLS policies
- Shared, Flask-free store helper for title-based upsert + versioning (reused by REST routes and agent tools so they never drift)
- REST routes: CRUD, version history, and restore — RBAC-gated (`artifacts` read/write)
- Agent tools (list / read / write) on the cloud toolbelt, plus MCP dispatch exposure
- Discoverability is via tool descriptions only — nothing added to any system prompt

## Frontend

- Artifacts tab: master/detail with View / Edit / History modes, plus create and delete
- Expandable per-version dropdown to read exactly what each version wrote
- Detail and history reads cached via `useQuery` — re-opening an artifact and re-expanding a version are served from cache instead of refetching

## Testing

- REST CRUD full cycle; cross-org access returns 403
- Agent tools end-to-end, including same-title re-write producing a version bump (proves cross-run persistence)
- Architectural RBAC tests pass; `tsc` and ESLint clean for the changed frontend

## Summary by CodeRabbit

* **New Features**
  * Persistent markdown artifacts: create, view, edit, delete, per-version viewing, and restore from history.
  * Artifacts tab in Monitor with creation UI, edit/history modes, title-collision blocking, and per-item delete confirmation.
  * Client-side service and API routes for listing, fetching, creating, updating, deleting, versioning, and restoring artifacts.
  * Agent-accessible artifact tools and registry integration for programmatic list/read/write operations.

* **Chores**
  * Database schema, row-level security, and RBAC updates to support tenant-scoped artifact storage and versioning.

### DIFF
--- a/client/src/lib/services/artifacts.ts
+++ b/client/src/lib/services/artifacts.ts
@@ -1,0 +1,131 @@
+'use client';
+
+import { apiGet, apiPost, apiRequest, type ApiError } from '@/lib/services/api-client';
+
+// ============================================================================
+// Types
+// ============================================================================
+
+export type ArtifactEditor = 'agent' | 'user';
+
+export interface ArtifactSummary {
+  id: string;
+  title: string;
+  version: number;
+  lastEditedBy: ArtifactEditor;
+  updatedAt: string | null;
+  createdAt: string | null;
+}
+
+export interface ArtifactData extends ArtifactSummary {
+  content: string;
+}
+
+export interface ArtifactVersion {
+  id: string;
+  versionNumber: number;
+  source: string;
+  userId: string;
+  createdAt: string | null;
+  generationSessionId: string | null;
+}
+
+export interface ArtifactVersionDetail extends ArtifactVersion {
+  content: string;
+}
+
+// ============================================================================
+// Service
+// ============================================================================
+
+export const artifactsService = {
+  async listArtifacts(): Promise<ArtifactSummary[]> {
+    try {
+      const data = await apiGet<{ artifacts: ArtifactSummary[] }>('/api/artifacts');
+      return data.artifacts ?? [];
+    } catch (error) {
+      console.error('Error fetching artifacts:', error);
+      return [];
+    }
+  },
+
+  async getArtifact(id: string): Promise<ArtifactData | null> {
+    try {
+      const data = await apiGet<{ artifact: ArtifactData }>(`/api/artifacts/${id}`);
+      return data.artifact ?? null;
+    } catch (error) {
+      if ((error as ApiError).status === 404) return null;
+      console.error('Error fetching artifact:', error);
+      return null;
+    }
+  },
+
+  async createArtifact(title: string, content: string): Promise<{ success: boolean; id?: string; error?: string }> {
+    try {
+      const data = await apiPost<{ id: string; version: number }>('/api/artifacts', { title, content });
+      return { success: true, id: data.id };
+    } catch (error) {
+      const apiErr = error as ApiError;
+      return { success: false, error: apiErr.message || 'Failed to create artifact' };
+    }
+  },
+
+  async updateArtifact(id: string, content: string): Promise<{ success: boolean; error?: string }> {
+    try {
+      await apiRequest(`/api/artifacts/${id}`, {
+        method: 'PATCH',
+        body: JSON.stringify({ content }),
+      });
+      return { success: true };
+    } catch (error) {
+      const apiErr = error as ApiError;
+      return { success: false, error: apiErr.message || 'Failed to update artifact' };
+    }
+  },
+
+  async deleteArtifact(id: string): Promise<{ success: boolean; error?: string }> {
+    try {
+      await apiRequest(`/api/artifacts/${id}`, { method: 'DELETE' });
+      return { success: true };
+    } catch (error) {
+      const apiErr = error as ApiError;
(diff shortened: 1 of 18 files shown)


server/chat/backend/agent/tools/artifact_tool.py
```diff
@@ -0,0 +1,177 @@
...
+                "updated_at": row[2].isoformat() if row[2] else None,
+                "version": row[3],
+            }
+            for row in rows
+        ]
+        return json.dumps({"status": "ok", "artifacts": artifacts})
+
+    except Exception:
+        logger.exception("[ArtifactTool] Failed to list artifacts")
+        return json.dumps({"error": "Failed to list artifacts."})
```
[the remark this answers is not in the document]
isiddharthsingh: Intentionally not addressing this one in this PR. RLS (set_rls_context) already scopes every artifact tool query to the caller's org, so cross-tenant access is blocked. Casbin RBAC is enforced at the REST layer (@require_permission), but no agent tool in the codebase enforces it — the agent toolbelt uniformly relies on RLS for isolation and connection-status for gating. Adding a permission check to the artifact tools alone would be inconsistent with every other tool and wouldn't close the gap meaningfully. Per-role RBAC on the agent surface is a cross-cutting change we'll handle separately rather than as an artifacts-specific patch.

client/src/lib/services/artifacts.ts
```diff
@@ -0,0 +1,132 @@
...
+export const artifactsService = {
+  async listArtifacts(): Promise<ArtifactSummary[]> {
+    try {
+      const data = await apiGet<{ artifacts: ArtifactSummary[] }>('/api/artifacts');
+      return data.artifacts ?? [];
+    } catch (error) {
+      console.error('Error fetching artifacts:', error);
+      return [];
+    }
+  },
```
damianloch: dead code

client/src/lib/services/artifacts.ts
```diff
@@ -0,0 +1,132 @@
...
+  async getArtifact(id: string): Promise<ArtifactData | null> {
+    try {
+      const data = await apiGet<{ artifact: ArtifactData }>(`/api/artifacts/${id}`);
+      return data.artifact ?? null;
+    } catch (error) {
+      // 404 means "doesn't exist" (a valid empty state); surface everything else
+      // so the UI can show a real error instead of a misleading "not found".
+      if ((error as ApiError).status === 404) return null;
+      throw error;
+    }
```
damianloch, line 50: getArtifact and getVersion treat 404 as a valid empty state (return null) and re-throw everything else — good, lets useQuery populate its error field. But getVersions catches everything and embeds the error in the return value ({ versions: [], error: '...' }). This means consumers handle errors differently depending on which method they call.

client/src/lib/services/artifacts.ts
```diff
@@ -0,0 +1,132 @@
...
+      const data = await apiPost<{ success: boolean; content: string }>(
+        `/api/artifacts/${id}/versions/${versionId}/restore`,
+      );
+      return { success: true, content: data.content };
+    } catch (error) {
+      const apiErr = error as ApiError;
+      return { success: false, error: apiErr.message || 'Failed to restore version' };
+    }
+  },
+};
```
damianloch: The restoreVersion method parses content from the response and returns it, but in artifacts-tab.tsx L374-382 the caller only checks result.success and then calls reloadArtifact() to refetch. The returned content is wasted. Harmless but the type could be simplified to just { success: boolean; error?: string }

server/chat/backend/agent/tools/artifact_tool.py
damianloch, line 1: One thing that is left out is some way for the agent to see edits that happened to the Canvas. Users want the ability to edit a canvas and have the LLM understand that edit (for example maybe there is an artifact of security issues, and a user deletes issues 5 and 7 because they solved them, then next updates shouldn't repopulate these issues since user explicitly removed them...). This was a requested ability for this feature. We need some way to expose them changes

damianloch (changes requested):

server/services/actions/executor.py
```diff
@@ -172,6 +172,26 @@ def build_action_prompt(
...
     ]
 
+    # Auto-maintain a living artifact for every action so results accumulate
+    # across runs without the user having to ask for it. Skip when the user's
+    # own instructions already steer a persistent doc — their steering wins.
+    # ("artifact" also matches read_artifact / write_artifact mentions.)
+    instructions_lc = action["instructions"].lower()
+    user_steers_doc = any(
+        kw in instructions_lc
+        for kw in ("artifact", "living document", "living doc", "persist across runs")
```
damianloch: Please no more substring matching in our code its so flaky. I could prompt action to "check my artifact registry for versions" and we wouldn't append the prompt just 1 example. I don't think its even necessary to conditionally apply a prompt here, might as well always do it.

damianloch (changes requested):

damianloch (approved):

PR #482, opened 2026-06-05
feat: kubeconfig upload connector for on-prem kubectl

damianloch:
## Summary

Adds a new Kubernetes authentication method: direct kubeconfig file uploads. This enables kubectl access for air-gapped or restricted environments where deploying the WebSocket agent (Helm chart) is not feasible — e.g. clusters only reachable through jump VMs.

Users can now choose between two auth methods on the `/kubectl/auth` page:
1. **Deploy Agent** (recommended) — install the Helm-based WebSocket agent into the target cluster
2. **Upload Kubeconfig** — upload one or more kubeconfig YAML files directly from the browser

Uploaded kubeconfigs are stored securely in Vault and used as a fallback execution path when no live agent is connected for a given cluster.

### Backend
- **New `kubeconfig_clusters` table** with RLS, storing metadata (context, cluster name, server URL, vault reference) per uploaded kubeconfig
- **New `/api/kubeconfig/` routes** — upload (POST), list (GET), delete (DELETE) with RBAC decorators and rate limiting
- **Vault integration** — full kubeconfig YAML stored as a Vault secret; referenced by `vault_provider` in DB
- **Chatbot execution fallback** — `_handle_kubectl_execute` in `api_handler.py` now checks for uploaded kubeconfigs when no WebSocket agent is found, writes a temp file with secure permissions, and runs `kubectl` directly
- **Unified cluster discovery** — `get_connected_clusters` and `is_kubectl_onprem_connected` tools now `UNION` results from both `active_kubectl_connections` and `kubeconfig_clusters`
- **Connector status** — `_check_kubectl` in `connector_status.py` updated to account for kubeconfig-sourced clusters

### Frontend
- **Tabbed auth UI** on `/kubectl/auth` — "Deploy Agent" tab (existing) and "Upload Kubeconfig" tab (new) with drag-and-drop file selection, client-side YAML validation, and batch upload
- **Manage page** — displays both agent-connected and kubeconfig-uploaded clusters with distinct badges ("Active"/"Stale" for agents, "Uploaded" for kubeconfigs)
- **Next.js API proxy routes** — `/api/kubeconfig/*` forwarding to backend

### Security
- Kubeconfig content never stored in the database — only in Vault
- Temporary kubeconfig files written to a private directory with `0o700`/`0o600` permissions
- `exec`-based auth supported (for GKE/EKS kubeconfigs that use `gke-gcloud-auth-plugin` or `aws` CLI)
- Org-scoped: all users in the same org see the same uploaded clusters; viewers cannot modify

## Test Plan
- [x] Upload single and multiple kubeconfig files via UI
- [x] Verify clusters appear on manage page with "Uploaded" badge
- [x] Run kubectl commands against uploaded kubeconfig cluster via chatbot
- [x] Delete uploaded cluster and confirm Vault cleanup
- [x] Verify agent-based clusters still work alongside kubeconfig clusters
- [x] CI passing (pre-checks, CodeQL, SonarCloud)

## Summary by CodeRabbit

* **New Features**
  * Upload kubeconfig files to register clusters; list and remove kubeconfig-backed clusters; UI adds tabbed flows for agent enrollment and kubeconfig uploads with validation and copy-to-clipboard feedback.
  * Unified cluster listing showing clusters from both agents and kubeconfigs; internal execution can run kubectl via agent or uploaded kubeconfig.

* **Bug Fixes**
  * Clearer fallback/error responses when agents are unavailable.

* **Documentation**
  * Updated kubectl integration docs to cover both connection modes.

* **Tests**
  * Adjusted tests for kubeconfig fallback behavior.

* **Chores**
  * Added persistent storage and discovery support for kubeconfig-backed clusters.


client/src/app/api/kubeconfig/[id]/route.ts
Comment (OlivierTrudeau) [inline client/src/app/api/kubeconfig/[id]/route.ts:1]: You can consolidate all the routes in one file: import { NextRequest } from 'next/server';
import { forwardRequest } from '@/lib/backend-proxy';

async function handler(
  request: NextRequest,
  { params }: { params: Promise<{ path: string[] }> },
) {
  const { path } = await params;
  const backendPath = '/bitbucket/' + path.join('/');
  return forwardRequest(request, request.method, backendPath, 'bitbucket');
}

export { handler as GET, handler as POST, handler as PUT, handler as DELETE };

server/utils/secrets/secret_ref_utils.py
```diff
@@ -188,7 +189,7 @@ def has_user_credentials(self, user_id: str, provider: str) -> bool:
                      AND secret_ref IS NOT NULL
                      AND is_active = TRUE
                    LIMIT 1""",
-                (*pred_params, provider_base),
+                (*pred_params, provider.lower()),
```
OlivierTrudeau, line 192: Shouldn't you get rid of the whole provider_base thing then?

server/utils/db/db_utils.py
```diff
@@ -1089,6 +1089,25 @@ def initialize_tables():
                     CREATE INDEX IF NOT EXISTS idx_kubectl_connections_token ON active_kubectl_connections(token);
                     CREATE UNIQUE INDEX IF NOT EXISTS idx_kubectl_connections_cluster_id ON active_kubectl_connections(cluster_id);
                 """,
+                "kubeconfig_clusters": """
+                    CREATE TABLE IF NOT EXISTS kubeconfig_clusters (
+                        id SERIAL PRIMARY KEY,
+                        user_id TEXT NOT NULL,
+                        org_id VARCHAR(255),
```
OlivierTrudeau, line 1096: Can't you just modify the previous table with a migration script? We're starting to have a lott of tables I feel like

client/src/app/kubectl/auth/page.tsx
```diff
@@ -63,380 +98,353 @@ export default function KubectlAuthPage() {
...
     }
   };
 
-  const CommandRow = ({ label, command, copyKey }: { label: string; command: string; copyKey: string }) => {
-    const highlightCommand = (cmd: string) => {
-      return cmd.split('\n').map((line, idx) => {
-        const parts: JSX.Element[] = [];
-        let remaining = line;
-        let key = 0;
+  const handleFileSelect = async (e: React.ChangeEvent<HTMLInputElement>) => {
```
OlivierTrudeau, line 132: is there a way to not drop like selected files at a time? Because what if a user has nested files? Like one directory with multiple subdirs that each contain their own kubeconfig files

client/src/app/kubectl/manage/page.tsx
```diff
@@ -20,46 +21,100 @@ interface Connection {
...
+  cluster_id: string;
+  cluster_name: string;
+  context_name: string;
+  server_url?: string;
+  created_at: string;
+  updated_at: string;
+}
+
+type MergedCluster = {
+  cluster_id: string;
```
OlivierTrudeau, line 34: Not super intuitive what a mergedcluster is. Maybe I'm being to nitpicky tho

server/chat/backend/agent/tools/list_clusters_tool.py
OlivierTrudeau, line 1: While you're on this could you rename this to smtg like list clusters tool or smtg a little more obvious if it's not a big lift

server/routes/kubeconfig/kubeconfig_routes.py
```diff
@@ -0,0 +1,230 @@
...
+_ERR_NO_ORG = "Could not resolve organization"
+
+
+def _make_cluster_id(org_id: str, context_name: str) -> str:
+    digest = hashlib.sha256(f"{org_id}:{context_name}".encode()).hexdigest()[:16]
+    return f"kubeconfig-{digest}"
+
+
+def _make_vault_provider(org_id: str, filename: str) -> str:
+    digest = hashlib.sha256(f"{org_id}:{filename}".encode()).hexdigest()[:12]
```
OlivierTrudeau, line 27: make vault key ref or something instead? Provider sounds like gcp secrets vs aws secrets

server/routes/kubeconfig/kubeconfig_routes.py
```diff
@@ -0,0 +1,230 @@
...
+
+    for entry in kubeconfigs:
+        _process_kubeconfig_entry(entry, user_id, org_id, registered, errors)
+
+    return jsonify({"registered": registered, "errors": errors}), 200 if registered else 400
+
+
+@kubeconfig_bp.route("/api/kubeconfig/clusters", methods=["GET"])
+@require_permission("connectors", "read")
+def list_kubeconfig_clusters(user_id):
```
OlivierTrudeau, line 166: Why not consolidate with the current lister? If we put them in the same table it'll make even more sense

server/utils/internal/api_handler.py
```diff
@@ -66,32 +62,123 @@ async def _handle_kubectl_execute(headers, body, writer):
...
+        os.fchmod(fd, 0o600)
+        os.write(fd, kubeconfig_yaml.encode())
+        os.close(fd)
+    except Exception:
+        os.close(fd)
+        os.unlink(kubeconfig_path)
+        raise
+
+    atexit.register(lambda p=kubeconfig_path: os.unlink(p) if os.path.exists(p) else None)
+
```
OlivierTrudeau: Apperently this will cause a memory leak. Better off just using the finally block to kill the process

server/utils/internal/api_handler.py
```diff
@@ -66,32 +62,123 @@ async def _handle_kubectl_execute(headers, body, writer):
...
+    try:
+        exec_env = {
+            "PATH": os.environ.get("PATH", ""),
+            "HOME": private_dir,
+            "KUBECONFIG": kubeconfig_path,
+        }
+        cmd_parts = ["kubectl", f"--context={context_name}"] + command.split()
+
```
OlivierTrudeau, line 160: If a command has arguments with spaces (e.g. get pods -l "app=my service"), it'll break

server/utils/internal/api_handler.py
```diff
@@ -66,32 +62,123 @@ async def _handle_kubectl_execute(headers, body, writer):
...
+    import atexit
+    import tempfile
+
+    private_dir = os.path.join(tempfile.gettempdir(), 'aurora_kubeconfig')
+    os.makedirs(private_dir, mode=0o700, exist_ok=True)
+
+    fd, kubeconfig_path = tempfile.mkstemp(prefix='aurora_kc_', suffix='.yaml', dir=private_dir)
+    try:
+        os.fchmod(fd, 0o600)
+        os.write(fd, kubeconfig_yaml.encode())
```
OlivierTrudeau, line 146: The permissions are false safety because the owner is the pod/container so another user running in the same pod could access the content of someone else. Difficult to do in practice but is a possible attack vector. Not sure what the alternative is tho

OlivierTrudeau (changes requested):

client/src/app/kubectl/auth/page.tsx
damianloch: I don't think its in scope to bend over backwards for this, we already support multi select and coding agetns can easily copy hundreds of nested files into a common dir instead of us hardcoding it somehow

server/utils/db/db_utils.py
damianloch: These tables have different lifecycles: `active_kubectl_connections` is transient session state (rows created/deleted by WebSocket connect/disconnect, with heartbeats and token refs), while `kubeconfig_clusters` is persistent config (uploaded creds stored in Vault, never auto-disconnect). Merging would require nullable columns for both schemas and conflicting lifecycle semantics. The UNION in the agent tool already gives a unified view to consumers.

server/routes/kubeconfig/kubeconfig_routes.py
damianloch: The AI agent tool (`get_connected_clusters`) already does a UNION ALL across both tables — that's the consolidated view for the chatbot. This REST endpoint (`/api/kubeconfig/clusters`) serves the management UI specifically, which needs different fields (vault ref, namespace, updated_at) and only shows kubeconfig-sourced clusters for CRUD operations. Different consumers, different data shapes.

client/src/app/api/kubeconfig/[id]/route.ts
Comment (damianloch) [reply client/src/app/api/kubeconfig/[id]/route.ts:1]: Done — consolidated into a single `[...path]/route.ts` catch-all, matching the pattern used by bitbucket/gcp/etc.

server/utils/secrets/secret_ref_utils.py
damianloch: The `provider_base` split serves a purpose — it validates the provider family is recognized before hitting the DB (e.g. rejects garbage like `banana_xyz`). The actual DB queries already use the full `provider.lower()` string. Removing it entirely would be a cross-cutting change to shared auth infra used by all connectors — better as a separate refactor if we decide to.

client/src/app/kubectl/manage/page.tsx
damianloch: Fair — renamed to `ConnectedCluster`.

server/chat/backend/agent/tools/list_clusters_tool.py
damianloch: Done — renamed to `list_clusters_tool.py`.

server/routes/kubeconfig/kubeconfig_routes.py
damianloch: Renamed the helper to `_make_vault_key`. The variable stays as `vault_provider` since it maps directly to the DB column of the same name in `user_tokens` — renaming the column would be a larger cross-cutting change.

server/utils/internal/api_handler.py
damianloch: Good catch — removed. The `finally` block already handles cleanup so atexit was just leaking lambdas.

server/utils/internal/api_handler.py
damianloch: Fixed — now using `shlex.split(command)` which handles quoted arguments correctly.

client/src/app/api/kubeconfig/[...path]/route.ts
```diff
@@ -0,0 +1,13 @@
+import { NextRequest } from 'next/server';
+import { forwardRequest } from '@/lib/backend-proxy';
+
+async function handler(
+  request: NextRequest,
+  { params }: { params: Promise<{ path: string[] }> },
+) {
+  const { path } = await params;
+  const backendPath = '/api/kubeconfig/' + path.join('/');
+  return forwardRequest(request, request.method, backendPath, 'kubeconfig');
```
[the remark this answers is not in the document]
Comment (damianloch) [reply client/src/app/api/kubeconfig/[...path]/route.ts:10]: Not applicable — cluster IDs are always `kubeconfig-<hex>` (SHA-256 hexdigest), so they never contain reserved URL characters.

client/src/app/kubectl/auth/page.tsx
```diff
@@ -28,6 +31,35 @@ const getHelmInstallCommand = (token: string) => hasRequiredEndpoints
...
+        <pre className="overflow-auto rounded-lg bg-zinc-900 border border-zinc-800 p-2 pr-12 text-xs leading-relaxed font-mono text-zinc-100">
+          {command}
+        </pre>
+        <Button
+          variant="ghost"
+          size="sm"
+          onClick={() => onCopy(command, copyKey)}
+          className={`absolute right-1 text-zinc-400 hover:text-zinc-100 hover:bg-transparent ${command.includes('\n') ? 'top-1' : 'top-1/2 -translate-y-1/2'}`}
+        >
+          {copied === copyKey ? <Check className="h-4 w-4" /> : <Copy className="h-4 w-4" />}
```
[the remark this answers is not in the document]
damianloch: Internal DevOps tool — deferring a11y improvements to a future pass.

client/src/app/kubectl/auth/page.tsx
```diff
@@ -63,380 +98,353 @@ export default function KubectlAuthPage() {
...
+      const hasApiVersion = lines.some(l => l.startsWith('apiVersion:'));
+      const hasClusters = lines.some(l => l.startsWith('clusters:'));
+      const hasContexts = lines.some(l => l.startsWith('contexts:'));
+      if (!hasApiVersion || !hasClusters || !hasContexts) {
+        setUploadError(`${file.name} does not appear to be a valid kubeconfig file`);
+        continue;
+      }
+      setPendingFiles(prev => [...prev.filter(f => f.filename !== file.name), { filename: file.name, content }]);
+    }
+  };
```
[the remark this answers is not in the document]
damianloch: Good catch — fixed. Added e.target.value = "" at end of handler.

client/src/app/kubectl/manage/page.tsx
```diff
@@ -20,46 +21,100 @@ interface Connection {
...
     jsonFetcher,
     { staleTime: 10_000, retryCount: 2, revalidateOnFocus: true },
   );
 
-  const connections = data?.connections ?? [];
+  const { data: kubeconfigData, isLoading: kcLoading, mutate: reloadKubeconfigs } = useQuery<{ clusters: KubeconfigCluster[] }>(
+    '/api/kubeconfig/clusters',
+    jsonFetcher,
+    { staleTime: 10_000, retryCount: 2, revalidateOnFocus: true },
+  );
```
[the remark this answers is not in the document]
damianloch: The manage page already calls fetchClusters() after a successful delete in the handler. Event propagation to other pages is low priority since this is the only consumer.

client/src/app/kubectl/manage/page.tsx
```diff
@@ -20,46 +21,100 @@ interface Connection {
...
-      if (activeConns.length === 0) {
+
+      if (clusterToDelete.source === 'agent') {
+        const res = await fetch(`/api/kubectl/connections/${clusterToDelete.cluster_id}`, { method: 'DELETE' });
+        if (!res.ok) throw new Error('Failed to disconnect');
+        const data = await res.json();
+        if (data.delete_command) { setDeleteCommand(data.delete_command); setShowCommandDialog(true); }
+        toast({ title: "Token revoked", description: `${clusterToDelete.cluster_name} disconnected` });
+      } else {
+        const res = await fetch(`/api/kubeconfig/${clusterToDelete.cluster_id}`, { method: 'DELETE' });
```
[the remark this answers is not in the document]
damianloch: Not applicable — cluster IDs are kubeconfig-<sha256 hex>, no reserved URL characters possible.

server/chat/backend/agent/skills/integrations/kubectl_onprem/SKILL.md
```diff
@@ -21,7 +21,7 @@ metadata:
 # On-Prem Kubernetes (kubectl) Integration
 
 ## Overview
-On-prem Kubernetes cluster integration for running kubectl commands on connected clusters. Commands are relayed through the Aurora agent installed on the cluster.
+On-prem Kubernetes cluster integration for running kubectl commands on connected clusters. Commands are either relayed through the Aurora agent installed on the cluster, or executed directly using uploaded kubeconfig credentials.
```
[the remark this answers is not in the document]
damianloch: Fixed — updated description and index in front-matter to mention kubeconfig support.

server/main_compute.py
```diff
@@ -361,6 +361,10 @@ def enforce_user_org_binding():
 from routes.kubectl_token_routes import kubectl_token_bp
 app.register_blueprint(kubectl_token_bp)
 
+# --- Kubeconfig Upload Routes ---
+from routes.kubeconfig.kubeconfig_routes import kubeconfig_bp
+app.register_blueprint(kubeconfig_bp)
```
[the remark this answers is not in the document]
damianloch: This matches the existing pattern used by several blueprints in this codebase that bake the full path into route decorators. Not changing for consistency.

server/routes/kubeconfig/kubeconfig_routes.py
```diff
@@ -0,0 +1,230 @@
...
+
+
+def _make_cluster_id(org_id: str, context_name: str) -> str:
+    digest = hashlib.sha256(f"{org_id}:{context_name}".encode()).hexdigest()[:16]
+    return f"kubeconfig-{digest}"
+
+
+def _make_vault_key(org_id: str, filename: str) -> str:
+    digest = hashlib.sha256(f"{org_id}:{filename}".encode()).hexdigest()[:12]
+    return f"kubeconfig_{digest}"
```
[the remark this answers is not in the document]
damianloch: The ON CONFLICT clause is intentional — uploading the same context from the same org should overwrite (e.g. rotating credentials). SHA-256 with 16 hex chars has negligible collision probability for realistic cluster counts per org.

server/routes/kubeconfig/kubeconfig_routes.py
```diff
@@ -0,0 +1,230 @@
...
+    return contexts
+
+
+def _process_kubeconfig_entry(entry, user_id, org_id, registered, errors):
+    """Process a single kubeconfig entry. Appends to registered/errors in place."""
+    filename = entry.get("filename", "unknown.yaml")
+    content = entry.get("content", "")
+
+    if len(content.encode()) > MAX_FILE_SIZE:
+        errors.append({"filename": filename, "error": f"File exceeds {MAX_FILE_SIZE // 1024}KB limit"})
```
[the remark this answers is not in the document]
damianloch: Fixed — added type validation for entry (must be dict) and content (must be string) before processing.

server/routes/kubeconfig/kubeconfig_routes.py
```diff
@@ -0,0 +1,230 @@
...
+                """, (user_id, org_id, cluster_id, ctx["context_name"],
+                      ctx["cluster_name"], ctx["server_url"], ctx["namespace"],
+                      vault_provider))
+                registered.append({
+                    "cluster_id": cluster_id,
+                    "context_name": ctx["context_name"],
+                    "cluster_name": ctx["cluster_name"],
+                    "server_url": ctx["server_url"],
+                })
+            conn.commit()
```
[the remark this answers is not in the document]
damianloch: Low risk in practice — orphan Vault secrets are harmless (no DB row references them) and get overwritten on re-upload of the same filename. Adding rollback complexity is not worth it for this edge case.

server/services/discovery/tasks.py
```diff
@@ -34,7 +35,16 @@ def _query_kubectl_orgs(cur):
...
+
+        UNION ALL
+
+        SELECT kc.org_id, MIN(u.id) AS user_id, kc.cluster_id, kc.cluster_name
+        FROM kubeconfig_clusters kc
+        JOIN users u ON u.org_id = kc.org_id
+        WHERE kc.is_active = TRUE AND kc.org_id IS NOT NULL
+        GROUP BY kc.org_id, kc.cluster_id, kc.cluster_name
+
+        ORDER BY 1
```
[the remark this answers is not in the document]
damianloch: Agent-connected and kubeconfig clusters use different cluster_id formats (token-based vs SHA hash) so duplicates are not possible in practice. The UNION ALL is intentional to avoid the overhead of dedup sorting.

server/utils/internal/api_handler.py
```diff
@@ -66,32 +63,120 @@ async def _handle_kubectl_execute(headers, body, writer):
...
+    # Fallback: try kubeconfig-based direct execution
+    kubeconfig_info = _get_kubeconfig_for_cluster(user_id, cluster_id)
+    if kubeconfig_info:
+        result = await _execute_kubectl_with_kubeconfig(
+            command, kubeconfig_info['yaml'], kubeconfig_info['context_name'], timeout
+        )
+        await _send_json_response(writer, result)
+        return
     
+    await _send_json_response(writer, {'success': False, 'error': f"No active agent or kubeconfig for cluster '{cluster_id}'"})
```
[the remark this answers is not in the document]
damianloch: Fixed — wrapped the kubeconfig fallback in try/except to return a structured JSON error instead of terminating the socket.

server/utils/internal/api_handler.py
```diff
@@ -66,32 +63,120 @@ async def _handle_kubectl_execute(headers, body, writer):
...
     
-    command_id = str(uuid.uuid4())
-    future = asyncio.Future()
-    register_command_response_handler(command_id, future)
+    # Fallback: try kubeconfig-based direct execution
+    kubeconfig_info = _get_kubeconfig_for_cluster(user_id, cluster_id)
+    if kubeconfig_info:
+        result = await _execute_kubectl_with_kubeconfig(
+            command, kubeconfig_info['yaml'], kubeconfig_info['context_name'], timeout
+        )
```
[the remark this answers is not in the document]
damianloch: Exec-based auth was intentionally enabled — GKE/EKS kubeconfigs require exec plugins (gke-gcloud-auth-plugin, aws). The chatbot container has these CLIs installed. Users cannot upload arbitrary binaries; exec commands are limited to what exists in the container PATH. This is expected behavior, not a vulnerability.

server/utils/internal/api_handler.py
```diff
@@ -66,32 +63,120 @@ async def _handle_kubectl_execute(headers, body, writer):
...
+        os.close(fd)
+        os.unlink(kubeconfig_path)
+        raise
+
+    try:
+        exec_env = {
+            "PATH": os.environ.get("PATH", ""),
+            "HOME": private_dir,
+            "KUBECONFIG": kubeconfig_path,
+        }
```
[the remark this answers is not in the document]
damianloch: Valid — fixed. HOME is now a unique temp dir per invocation (mkdtemp) and cleaned up in the finally block via shutil.rmtree. Prevents auth plugin credential caching from leaking between requests.

server/utils/internal/api_handler.py
```diff
@@ -66,32 +63,125 @@ async def _handle_kubectl_execute(headers, body, writer):
...
-        
-        await websocket.send(json.dumps({'type': 'kubectl_command', 'command_id': command_id, 'command': command, 'timeout': timeout}))
-        result = await asyncio.wait_for(future, timeout=timeout)
-        await _send_json_response(writer, result)
-    except asyncio.TimeoutError:
-        await _send_json_response(writer, {'success': False, 'error': 'No response from agent'})
     except Exception as e:
-        logger.error(f"Error communicating with kubectl agent: {e}", exc_info=True)
-        await _send_json_response(writer, {'success': False, 'error': f'Agent communication error: {str(e)}'})
+        logger.error("Kubeconfig fallback failed for cluster %s: %s", cluster_id, e, exc_info=True)
```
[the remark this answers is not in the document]
damianloch: done

server/routes/kubeconfig/kubeconfig_routes.py
OlivierTrudeau: How it's an uncreated table no? you just rename it

OlivierTrudeau (dismissed):

OlivierTrudeau (approved):

PR #483, opened 2026-06-05
fix: enforce seat limits in admin create_user route

beng360:
## Summary
- The `before_add_member` hook from #474 was only wired into `org_routes.py` (add_member + create_invitation)
- The admin panel uses `admin_routes.py:create_user()` which bypassed the seat check entirely
- Adds the same hook check before the INSERT in create_user

## Test plan
- Try adding a member via admin panel when at seat limit — should get 403
- Adding members under the limit should still work

## Summary by CodeRabbit

* **New Features**
  * Seat limit enforcement when admins create new users or invitations.
  * Seat limit enforcement when users move to a different organization.
  * Limit checks include current members and pending invitations; displays "Seat limit reached" (or a provided message) when denied.
  * Best-effort enforcement for org joins; concurrent requests may temporarily exceed the limit.

### DIFF
--- a/server/routes/admin_routes.py
+++ b/server/routes/admin_routes.py
@@ -123,6 +123,15 @@ def create_user(user_id):
                     (org_id, target_email),
                 )
 
+                from utils.hooks import get_hook
+                cur.execute("SELECT COUNT(*) FROM users WHERE org_id = %s", (org_id,))
+                _cnt = cur.fetchone()[0]
+                cur.execute("SELECT COUNT(*) FROM org_invitations WHERE org_id = %s AND status = 'pending'", (org_id,))
+                _cnt += cur.fetchone()[0]
+                allowed, msg = get_hook("before_add_member")(org_id, _cnt)
+                if not allowed:
+                    return jsonify({"error": msg or "Seat limit reached"}), 403
+
                 invitation_id = str(_uuid.uuid4())
                 expires_at = datetime.now(timezone.utc) + timedelta(days=7)
 
--- a/server/routes/admin_routes.py
+++ b/server/routes/admin_routes.py
@@ -162,7 +162,6 @@ def create_user(user_id):
             _cnt += cur.fetchone()[0]
             allowed, msg = get_hook("before_add_member")(org_id, _cnt)
             if not allowed:
-                conn.close()
                 return jsonify({"error": msg or "Seat limit reached"}), 403
 
             password_hash = bcrypt.hashpw(password.encode("utf-8"), bcrypt.gensalt()).decode("utf-8")
--- a/server/routes/admin_routes.py
+++ b/server/routes/admin_routes.py
@@ -155,6 +155,16 @@ def create_user(user_id):
             if len(password) < 8:
                 return jsonify({"error": "Password must be at least 8 characters"}), 400
 
+            from utils.hooks import get_hook
+            cur.execute("SELECT COUNT(*) FROM users WHERE org_id = %s", (org_id,))
+            _cnt = cur.fetchone()[0]
+            cur.execute("SELECT COUNT(*) FROM org_invitations WHERE org_id = %s AND status = 'pending'", (org_id,))
+            _cnt += cur.fetchone()[0]
+            allowed, msg = get_hook("before_add_member")(org_id, _cnt)
+            if not allowed:
+                conn.close()
+                return jsonify({"error": msg or "Seat limit reached"}), 403
+
             password_hash = bcrypt.hashpw(password.encode("utf-8"), bcrypt.gensalt()).decode("utf-8")
 
             try:
--- a/server/routes/org_routes.py
+++ b/server/routes/org_routes.py
@@ -849,6 +849,15 @@ def join_org(user_id):
                         conn.commit()
                     return jsonify({"error": "You are already a member of this organization"}), 409
 
+                from utils.hooks import get_hook
+                cursor.execute("SELECT COUNT(*) FROM users WHERE org_id = %s", (new_org_id,))
+                _cnt = cursor.fetchone()[0]
+                cursor.execute("SELECT COUNT(*) FROM org_invitations WHERE org_id = %s AND status = 'pending'", (new_org_id,))
+                _cnt += cursor.fetchone()[0]
+                allowed, msg = get_hook("before_add_member")(new_org_id, _cnt)
+                if not allowed:
+                    return jsonify({"error": msg or "Seat limit reached"}), 403
+
                 set_rls_context(cursor, conn, user_id, log_prefix="[OrgJoin]")
                 _transfer_user_to_org(cursor, user_id, old_org_id, new_org_id, new_role)
 
--- a/server/routes/org_routes.py
+++ b/server/routes/org_routes.py
@@ -853,7 +853,11 @@ def join_org(user_id):
                 from utils.hooks import get_hook
                 cursor.execute("SELECT COUNT(*) FROM users WHERE org_id = %s", (new_org_id,))
                 _cnt = cursor.fetchone()[0]
-                cursor.execute("SELECT COUNT(*) FROM org_invitations WHERE org_id = %s AND status = 'pending'", (new_org_id,))
+                cursor.execute(
+                    """SELECT COUNT(*) FROM org_invitations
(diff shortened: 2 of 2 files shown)


damianloch (commented):
## Review findings

### ⚠️ Double `conn.close()` in `admin_routes.py`

**File:** `server/routes/admin_routes.py` (seat-limit early-return path)

The early return calls `conn.close()` explicitly, but the function is wrapped in `try/finally` which also calls `conn.close()`. psycopg2 won't crash on double-close, but it's dead code that misleads readers into thinking the `finally` block doesn't exist. The other early returns (password/email checks) correctly rely on the `finally` — this one should too.

**Fix:** Remove the `conn.close()` from the early-return path.

---

### ⚠️ Inconsistent seat counting — backend excludes pending invitations, UI includes them

**Files:** `server/routes/admin_routes.py` and `server/routes/org_routes.py` (`join_org`) vs `get_seat_info`

In both enforcement points (`create_user` and `join_org`), the hook is called with only the **user count**. But `get_seat_info` defines `atLimit` as `(used + pending) >= max_seats`, including pending invitations. This means:

- If you're at 9/10 seats with 1 pending invite, the UI disables the "Add" button (correctly per its own logic), but the backend would still allow the request if someone hits the API directly (hook only sees 9 users).
- Conversely, once a pending invite is cancelled/expired, the UI unblocks but the enforcement was never actually blocking anyway.

**Fix:** Pass `member_count + pending_invitation_count` to the hook in both `admin_routes.py` and `org_routes.py:join_org` to match what `get_seat_info` reports — or document that enforcement is intentionally users-only and adjust the UI's `atLimit` formula to match.

---

### ⚠️ TOCTOU race on seat check

**Files:** `server/routes/admin_routes.py`, `server/routes/org_routes.py`

The `SELECT COUNT(*) → hook check → INSERT` sequence is not atomic. Two concurrent requests can both read count=9 (limit 10), both pass the hook, and both INSERT — resulting in 11 members. Default `READ COMMITTED` isolation doesn't prevent this since neither request holds a lock.

This is a known trade-off with hook-based enforcement and likely acceptable for a soft limit, but worth acknowledging. If strict enforcement is needed later, a `SELECT ... FOR UPDATE` on the `organizations` row would serialise the check+insert.

server/routes/admin_routes.py
```diff
@@ -155,6 +155,16 @@ def create_user(user_id):
             if len(password) < 8:
                 return jsonify({"error": "Password must be at least 8 characters"}), 400
 
+            from utils.hooks import get_hook
+            cur.execute("SELECT COUNT(*) FROM users WHERE org_id = %s", (org_id,))
+            _cnt = cur.fetchone()[0]
+            cur.execute("SELECT COUNT(*) FROM org_invitations WHERE org_id = %s AND status = 'pending'", (org_id,))
+            _cnt += cur.fetchone()[0]
+            allowed, msg = get_hook("before_add_member")(org_id, _cnt)
+            if not allowed:
```
damianloch, line 173: This `conn.close()` is redundant — the function is wrapped in `try/finally` (line ~196) which already closes the connection on any return path. The other early returns in this function (password check, email check, etc.) correctly rely on the `finally` block. This should be removed to stay consistent and avoid a double-close.

server/routes/admin_routes.py
```diff
@@ -155,6 +155,16 @@ def create_user(user_id):
             if len(password) < 8:
                 return jsonify({"error": "Password must be at least 8 characters"}), 400
 
+            from utils.hooks import get_hook
+            cur.execute("SELECT COUNT(*) FROM users WHERE org_id = %s", (org_id,))
+            _cnt = cur.fetchone()[0]
+            cur.execute("SELECT COUNT(*) FROM org_invitations WHERE org_id = %s AND status = 'pending'", (org_id,))
```
damianloch, line 170: This only counts actual users, but `get_seat_info` (which drives the UI's "at limit" state) computes `atLimit` as `(used + pending) >= max_seats` — including pending invitations. The mismatch means the UI can show "seat limit reached" while the backend would still allow the request, or vice versa.

Should this include pending invitations to match the UI logic?
```python
cur.execute("SELECT COUNT(*) FROM users WHERE org_id = %s", (org_id,))
member_count = cur.fetchone()[0]
cur.execute("SELECT COUNT(*) FROM org_invitations WHERE org_id = %s AND status = 'pending'", (org_id,))
member_count += cur.fetchone()[0]
```
(Same applies to the `join_org` path in `org_routes.py`.)

server/routes/org_routes.py
```diff
@@ -849,6 +849,15 @@ def join_org(user_id):
                         conn.commit()
                     return jsonify({"error": "You are already a member of this organization"}), 409
 
+                from utils.hooks import get_hook
+                cursor.execute("SELECT COUNT(*) FROM users WHERE org_id = %s", (new_org_id,))
+                _cnt = cursor.fetchone()[0]
```
damianloch, line 855: Nit / known gap: the `SELECT COUNT → hook check → INSERT/transfer` sequence isn't atomic. Two concurrent requests can both read count=9 (limit 10), both pass the hook, and both succeed — ending up at 11 members. Acceptable as a soft limit, but worth a comment noting this is best-effort enforcement. A `SELECT ... FOR UPDATE` on the `organizations` row would serialise if strict enforcement is ever needed.

server/routes/admin_routes.py
beng360: fixed in f67c526e — removed the redundant conn.close(), the finally block handles it.

server/routes/admin_routes.py
beng360: already addressed in cd76eee — both paths now count pending invitations.

server/routes/org_routes.py
beng360: added a comment noting it's best-effort. acceptable as a soft limit for now — strict enforcement would need SELECT FOR UPDATE on the org row.

server/routes/org_routes.py
```diff
@@ -849,6 +849,14 @@ def join_org(user_id):
                         conn.commit()
                     return jsonify({"error": "You are already a member of this organization"}), 409
 
+                # Hook: check if target org can accept more members (seat limit)
+                from utils.hooks import get_hook
+                cursor.execute("SELECT COUNT(*) FROM users WHERE org_id = %s", (new_org_id,))
+                member_count = cursor.fetchone()[0]
+                hook_allowed, hook_message = get_hook("before_add_member")(new_org_id, member_count)
+                if not hook_allowed:
+                    return jsonify({"error": hook_message or "This organization has reached its seat limit."}), 403
```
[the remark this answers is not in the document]
beng360: already done in cd76eee.

server/routes/admin_routes.py
```diff
@@ -155,6 +155,15 @@ def create_user(user_id):
...
+            _cnt += cur.fetchone()[0]
+            allowed, msg = get_hook("before_add_member")(org_id, _cnt)
+            if not allowed:
+                return jsonify({"error": msg or "Seat limit reached"}), 403
```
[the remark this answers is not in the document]
beng360: fixed in 709f1ef5 — added the same hook check to Branch A before inserting the invitation.

server/routes/org_routes.py
```diff
@@ -849,6 +849,16 @@ def join_org(user_id):
...
 
+                # Best-effort seat enforcement (not serialised — concurrent requests can overshoot by 1)
+                from utils.hooks import get_hook
+                cursor.execute("SELECT COUNT(*) FROM users WHERE org_id = %s", (new_org_id,))
+                _cnt = cursor.fetchone()[0]
+                cursor.execute("SELECT COUNT(*) FROM org_invitations WHERE org_id = %s AND status = 'pending'", (new_org_id,))
+                _cnt += cursor.fetchone()[0]
+                allowed, msg = get_hook("before_add_member")(new_org_id, _cnt)
+                if not allowed:
+                    return jsonify({"error": msg or "Seat limit reached"}), 403
```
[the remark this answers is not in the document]
beng360: fixed in 709f1ef5 — now excludes the invitation being accepted via IS DISTINCT FROM.

damianloch (approved):

PR #484, opened 2026-06-05
fix: always load Casbin policy from DB to prevent stale role cache across gunicorn workers

Harrio-6:
Load policy from DB on every permission check to eliminate stale RBAC cache across gunicorn workers, so that a gunicorn worker never get a stale cache
Check for db state every time a role is changed and refresh if necessary
Remove time-based reload interval and reload_policies() helper (no longer needed)
Pass SQLAlchemy engine with connection pooling to Casbin adapter instead of raw URL
Remove reload_policies() calls from admin routes (assign/revoke/delete)

## Summary by CodeRabbit

* **Refactor**
  * Authorization enforcer initialization centralized; policy loading now coordinates a shared version check so all workers observe consistent permission state.

* **Bug Fixes**
  * Role and user permission changes now update a shared policy version to propagate updates reliably across workers and avoid stale enforcement.
  * Permission checks now verify freshness before evaluating access, improving immediate consistency.

* **Tests**
  * Authorization tests and fixtures updated to align with the new versioned policy and initialization behavior.

### DIFF
--- a/server/routes/admin_routes.py
+++ b/server/routes/admin_routes.py
@@ -14,7 +14,7 @@ import uuid as _uuid
 from datetime import datetime, timedelta, timezone
 
 from utils.auth.rbac_decorators import require_permission
-from utils.auth.enforcer import get_enforcer
+from utils.auth.enforcer import get_enforcer, increment_policy_version
 from utils.auth.stateless_auth import get_org_id_from_request, set_rls_context
 from utils.db.db_utils import connect_to_db_as_user
 from utils.log_sanitizer import sanitize
@@ -279,6 +279,7 @@ def assign_role(user_id, target_user_id):
         enforcer.add_grouping_policy(target_user_id, role)
 
     enforcer.save_policy()
+    increment_policy_version()
 
     # Keep the convenience column in sync
     conn = connect_to_db_as_user()
@@ -328,6 +329,7 @@ def revoke_role(user_id, target_user_id, role):
             enforcer.add_grouping_policy(target_user_id, "viewer")
 
     enforcer.save_policy()
+    increment_policy_version()
 
     if org_id:
         fallback_role = (enforcer.get_roles_for_user_in_domain(target_user_id, org_id) or ["viewer"])[0]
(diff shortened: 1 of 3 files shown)


server/utils/auth/enforcer.py
damianloch, line 1: This fixes the correctness issue but introduces a full `load_policy()` DB round-trip on every single permission check (which fires on nearly every API request). Under load that'll add up.

Since we already have Redis in the stack, a cheap improvement: use a generation counter.

```python
# enforcer.py
POLICY_VERSION_KEY = "casbin:policy_version"
_cached_version: str | None = None

def enforce_with_reload(...) -> bool:
    global _cached_version
    enforcer = get_enforcer()
    version = get_redis_client().get(POLICY_VERSION_KEY) or "0"
    if version != _cached_version:
        with _lock:
            if version != _cached_version:
                enforcer.load_policy()
                _cached_version = version
    return enforcer.enforce(user_id, org_id, resource, action)
```

Then `INCR casbin:policy_version` after `save_policy()` in the admin routes. This gives you instant cross-worker propagation at the cost of a single Redis GET (~0.1ms) instead of a full policy table load on every request.

damianloch (changes requested):

server/routes/admin_routes.py
```diff
@@ -279,7 +279,7 @@ def assign_role(user_id, target_user_id):
         enforcer.add_grouping_policy(target_user_id, role)
```
damianloch: get_enforcer() returns the singleton without checking the Redis version counter. On a multi-worker deploy, this worker's in-memory policy can be stale. The admin reads current_roles from stale state, then decides which policies to remove/add based on that stale read.

server/routes/admin_routes.py
damianloch: get_enforcer() returns the singleton without checking the Redis version counter. On a multi-worker deploy, this worker's in-memory policy can be stale. The admin reads current_roles from stale state, then decides which policies to remove/add based on that stale read.

damianloch (changes requested):

damianloch (approved):

PR #485, opened 2026-06-05
Remove agent-based MCP tools (chat_with_aurora, ask_incident, regenerate_rca)

damianloch:
## Summary

Removes three agent-workflow MCP tools that degrade the MCP client experience:

- **`chat_with_aurora`** — full autonomous agent (blocks 45s polling)
- **`ask_incident`** — incident-scoped investigation (blocks 40s polling)
- **`regenerate_rca`** — re-runs full RCA pipeline (blocks 60s)

These tools dominate context windows with large docstrings, produce unpredictable latency, and pull LLMs toward slow agent paths when fast direct-read tools would answer the same questions.

## What changes

- **Deleted** `server/aurora_mcp/chat_bridge.py` (267 lines) — sole consumer was `chat_with_aurora`
- **Deleted** `server/tests/mcp/test_chat_bridge.py` — tests for the removed module
- **Removed** tool registrations + backing helpers from `tools_always_on.py` (~100 lines)
- **Updated** MCP server `instructions` string to guide toward `trigger_rca` + `search_tools`
- **Cleaned** all textual references across 7 files (dispatch.py, registry.py, resources.py, prompts.py, tools_gated.py, mcp_server.py, chat_routes.py)
- **Updated** `scripts/measure_mcp_surface.py` tier1 set
- **Updated** public docs (`website/docs/integrations/mcp.md`)

**12 files changed, +31 / -550**

## What is NOT changing

- `trigger_rca` remains (fire-and-forget, returns `incident_id` immediately)
- All Flask REST endpoints remain operational (web UI, API consumers unaffected)
- The LangGraph agent, Celery background tasks, and chatbot WebSocket are untouched
- Tier-2 and Tier-3 tool surfaces remain intact
- `search_tools` / `call_tool` dispatch layer unaffected

## Rationale

The MCP protocol is designed for fast, deterministic tool calls. Agent-based tools that poll for 40-60s violate this contract and create several problems:

1. **Context window pollution** — the verbose docstrings (especially `chat_with_aurora` at ~1200 tokens) crowd out useful tools
2. **Unpredictable latency** — MCP clients expect sub-second responses; 45s polls cause timeouts and partial-result states that clients handle poorly
3. **LLM attraction** — models are drawn to the most capable-sounding tool, so they route through the slow agent even when a direct read suffices
4. **Session state complexity** — `session_id` threading adds failure modes that don't exist in stateless reads

Users who need the full agent workflow can use the Aurora web UI, which is purpose-built for that interaction pattern.

## Test plan

- [ ] `python -m py_compile server/aurora_mcp/tools_always_on.py` passes
- [ ] `python -m py_compile server/mcp_server.py` passes
- [ ] `pytest server/tests/mcp/ -x` passes (dispatch tests unaffected)
- [ ] MCP client `tools/list` no longer returns the three removed tools
- [ ] `trigger_rca` still works via MCP
- [ ] Web UI chat and RCA features work normally (endpoints untouched)

## Summary by CodeRabbit

* **New Features**
  * Introduced trigger_rca to start incident investigations from a problem description.

* **Removed Features**
  * Deprecated/removed interactive chat-style incident tools (chat_with_aurora, ask_incident, regenerate_rca) from the primary tool surface.

* **Behavior Changes**
  * Tiering and tool-selection guidance updated to favor direct read tools and trigger_rca for RCAs; search hint text simplified.

* **Documentation**
  * MCP integration docs and tool docs updated to reflect the new workflow.

* **Tests**
  * Removed tests covering the deprecated chat bridge behavior.


scripts/measure_mcp_surface.py
```diff
@@ -45,7 +45,7 @@ def measure() -> Dict[str, Any]:
 
     all_tools = list(mcp._tool_manager._tools.values())  # noqa: SLF001
     tier1_names = {
-        "chat_with_aurora", "list_incidents", "get_incident", "ask_incident",
+        "list_incidents", "get_incident",
         "trigger_rca", "knowledge_base_search", "search_runbooks",
```
beng360, line 52: fix the coderabbit comment of the same nature

server/mcp_server.py
```diff
@@ -287,14 +287,11 @@ async def _api(
...
-        "user's connected systems — multi-source investigation/RCA OR taking action "
-        "(provisioning or changing infra via Terraform/kubectl/cloud CLIs, applying "
-        "code fixes, remediating). It runs the full agent workflow and is slower, so "
-        "it's the escalation path, not the default. It is NOT for questions about "
-        "the Aurora product itself (how the app works, its features, UI, settings) — "
-        "answer those from your own knowledge. Invoke discovered tools via call_tool."
+        "find the direct tool before assuming it's missing. Use `trigger_rca` to "
+        "start a new investigation from a problem description. It is NOT for "
+        "questions about the Aurora product itself (how the app works, its "
+        "features, UI, settings) — answer those from your own knowledge. "
```
beng360, line 293: no migration path right now for ask_incident or regenerate_rca, maybe we should indicate to users to reconnect their MCP idk

damianloch: Addressed — added deprecation stubs in 506075f5 (signature fix in 596a3143). Clients with stale caches will get an instant response telling them the tool is removed, what to use instead, and to disconnect/reconnect.

Cursor ignores `notifications/tools/list_changed` (known conformance gap), so there's no way to force a refresh server-side. The stubs are the cleanest middle ground — zero latency, clear guidance, safe to rip out after a release cycle.

scripts/measure_mcp_surface.py
damianloch: Fixed in 555e8987 — added all missing Tier-1 tool names. Was a pre-existing bug, good cleanup.

OlivierTrudeau (approved):

PR #486, opened 2026-06-06
fix(mcp): remove broken JSON truncation from tool responses

damianloch:
## Summary

- Removes the naive `encoded[:60000]` character-slice truncation from `truncate_payload()` in `server/aurora_mcp/response.py` that was producing broken/unparseable JSON for LLMs
- `truncate_payload()` now just ensures the response is a dict (FastMCP contract) and returns it unmodified
- Modern LLMs have 100K-200K token context windows; a full `list_incidents` response (~106K chars / ~26K tokens) is well within capacity

## Test plan

- [x] All 9 MCP response tests pass
- [ ] Verify `list_incidents` via MCP returns full valid JSON with 74 incidents (localhost)
- [ ] Confirm LLM clients (Cursor, Claude) can parse the full response without issues

## Summary by CodeRabbit

* **New Features**
  * Added offset-based pagination for incident listing (API and listing tool accept offset).

* **Refactor**
  * Removed hard response-size cap and simplified responses to always use a consistent dictionary shape.

* **Tests**
  * Updated tests to expect large payloads to pass through unchanged and to enforce a consistent wrapped listing envelope.

### DIFF
--- a/server/aurora_mcp/response.py
+++ b/server/aurora_mcp/response.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 from typing import Any, Dict, Iterable, List, Optional, Tuple
 
 
-def truncate_payload(payload: Any, *, tool_name: str = "mcp") -> Dict[str, Any]:
+def truncate_payload(payload: Any, **_kwargs: Any) -> Dict[str, Any]:
     """Guarantee a dict return shape for MCP tool responses.
 
     MCP tools are declared with a `Dict[str, Any]` return type. FastMCP
@@ -39,11 +39,10 @@ def paginate(
 
 
 def wrap_listing(
-    items: List[Any], *, cursor: Optional[str] = None, limit: int = 20, tool_name: str = "mcp"
+    items: List[Any], *, cursor: Optional[str] = None, limit: int = 20,
 ) -> Dict[str, Any]:
     """Standard envelope for list-returning tools: {items, next_cursor, total}."""
     page, next_cursor = paginate(items, cursor=cursor, limit=limit)
     return truncate_payload(
         {"items": page, "next_cursor": next_cursor, "total": len(items)},
-        tool_name=tool_name,
     )
--- a/server/aurora_mcp/response.py
+++ b/server/aurora_mcp/response.py
@@ -1,22 +1,12 @@
-"""Shared response helpers for MCP tools.
-
-Wraps `utils.tool_output_cap` so list-returning tools enforce the 15k-token
-cap consistently, plus a small cursor-paginate helper.
-"""
+"""Shared response helpers for MCP tools."""
 
 from __future__ import annotations
 
-import json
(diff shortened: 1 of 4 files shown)


beng360:
tested locally with 82 incidents (161k chars response). before: truncated at 60k boundary producing broken unparseable json. after: full valid json returned, all 82 incidents intact. fix works.

server/aurora_mcp/response.py
```diff
@@ -1,22 +1,12 @@
...
-import json
 from typing import Any, Dict, Iterable, List, Optional, Tuple
 
-# Roughly 15k tokens at 4 chars/token; aligned with PASS_THROUGH_CHARS in
-# chat.backend.agent.utils.tool_output_cap (40_000 chars ≈ 10–12k tokens; we
-# go slightly higher because MCP doesn't run an LLM summarizer in-process).
-RESPONSE_HARD_CHAR_CAP = 60_000
-
 
 def truncate_payload(payload: Any, *, tool_name: str = "mcp") -> Dict[str, Any]:
```
beng360: i'm pretty sure tool_name here is completeyly uselsess. drop the param or clean up call sites.

server/routes/incidents_routes.py
```diff
@@ -313,6 +313,12 @@ def get_incidents(user_id):
                 query += " LIMIT %s"
                 params.append(limit)
 
+                offset = request.args.get("offset", 0, type=int)
+                offset = max(0, offset)
+                if offset > 0:
```
beng360, line 327: no way to know when to stop paging the caller. return total count alongside incidentns array

server/aurora_mcp/tools_always_on.py
```diff
@@ -46,10 +46,12 @@ def _slim_incident(incident: Any) -> Any:
     return incident
 
 
-async def _do_list_incidents(api_call: ApiCall, status: Optional[str], limit: int) -> Dict[str, Any]:
+async def _do_list_incidents(api_call: ApiCall, status: Optional[str], limit: int, offset: int) -> Dict[str, Any]:
```
beng360, line 49: because this touches tool_always_on.py in the area which 485 removes somethings, it will create merge conflicts. make sure to rebase off main after 485 merge

server/aurora_mcp/response.py
damianloch: Done — changed to `**_kwargs` so existing callers don't need updates but the param is effectively dead.

server/routes/incidents_routes.py
damianloch: Done — added a COUNT query and the response is now `{"incidents": [...], "total": N}`. Also added it to the tool docstring so LLMs know when to stop paging.

server/aurora_mcp/tools_always_on.py
damianloch: Noted — will rebase after 485 merges.

beng360 (approved):

PR #491, opened 2026-06-08
fix: seed tool permissions on org creation

Harrio-6:
New orgs had 0 rows in org_tool_permissions, causing all gated tools to be blocked in background chats
Seed default tool permissions at org creation (both /register and /setup-org), matching existing seed_default_command_policy pattern

## Summary by CodeRabbit

* **New Features**
  * Tool permissions are now automatically initialized during organization setup and user registration; the system reports how many permissions were seeded.

* **Bug Fixes**
  * Seeding is idempotent and now uses robust logic that won’t fail the registration or setup if individual permission entries can’t be created; failures are handled gracefully and logged.

### DIFF
--- a/server/routes/auth_routes.py
+++ b/server/routes/auth_routes.py
@@ -146,6 +146,12 @@ def register():
                 except Exception as policy_err:
                     logging.warning("Failed to seed command policy for org %s", org_id, exc_info=policy_err)
 
+                try:
+                    from utils.auth.tool_registry import seed_org_tool_permissions
+                    seed_org_tool_permissions(org_id, user_id)
+                except Exception as tool_perm_err:
+                    logging.warning("Failed to seed tool permissions for org %s: %s", org_id, tool_perm_err)
+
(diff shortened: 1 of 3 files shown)


isiddharthsingh:
`seed_org_tool_permissions(org_id, user_id)` takes the target `org_id`, but then calls `set_rls_context(cur, conn, user_id)`, which re-resolves the org from `user_id` instead of using the passed `org_id`.

Why that matters: during `/setup-org`, a user may be moved from the Default Organization to a newly created org. If the user-to-org cache still contains the old org, RLS gets set to the old org while the insert tries to write rows for the new org. PostgreSQL RLS rejects those inserts, `auth_routes.py` logs the error and continues, and the new org ends up without seeded tool-permission rows.

Fix direction: in `server/utils/auth/tool_registry.py`, set `myapp.current_org_id` directly from the provided `org_id` when seeding, or otherwise clear/bypass the user org cache before seeding.

damianloch (approved):

PR #495, opened 2026-06-10
feat: github app env-var fallback for AWS Secrets Manager compatibility

Harrio-6:
Add can_handle_ref() check so the active backend gracefully declines incompatible secret references
Add GITHUB_APP_PRIVATE_KEY env var as fallback when Vault is not the secrets backend
Align get_app_private_key() with the try-backend-then-env pattern already used by get_app_webhook_secret()

## Summary by CodeRabbit

* **New Features**
  * GitHub App credentials now use a configurable secrets backend; AWS Secrets Manager supported alongside Vault
  * Private key must be read from the secrets backend (no env fallback); webhook secret prefers backend with existing env fallback preserved
  * Private key resolution is cached for performance

* **Documentation**
  * Updated GitHub App setup and troubleshooting to cover both Vault and AWS Secrets Manager backends

* **Tests**
  * Added regression tests covering backend reference mapping, resolution, fallback, and caching behavior


isiddharthsingh:
The approach needs to change. Right now the PR makes GitHub App work on AWS SM by falling back to a GITHUB_APP_PRIVATE_KEY env var when the backend can't handle the hardcoded vault: ref. That means the private key never actually lives in AWS Secrets Manager — it's just an env var, and AWS SM is bypassed. We want the key stored in and read from the active secrets backend (Vault or AWS SM), the same way every other connector is backend-agnostic. So: drop the env-var fallback and make the system-secret reference be built by the active backend.

⚠️ Note: just deleting GITHUB_APP_PRIVATE_KEY won't work — that brings back the original failure (the hardcoded vault: ref still can't be parsed by AWS SM, and now with no fallback). You have to add the backend-agnostic read below.

1. Add build_system_ref() to the secrets backend

server/utils/secrets/base.py — add to SecretsBackend:
def build_system_ref(self, logical_name: str) -> str:
    """Build a reference to a system-scoped (non per-user) secret.

    System secrets (e.g. the GitHub App private key) are operator-provisioned
    and read-only at runtime. Each backend maps the logical name to its own
    reference format so callers never hardcode a backend-specific prefix.
    """
    raise NotImplementedError

server/utils/secrets/vault_backend.py — override (this reproduces today's exact Vault ref, so existing Vault deployments are unaffected):
def build_system_ref(self, logical_name: str) -> str:
    # vault:kv/data/{mount}/system/{logical_name}
    return f"{VAULT_REF_PREFIX}{self.mount_point}/system/{logical_name}"
With the default mount aurora and logical_name="github-app/private-key", this yields vault:kv/data/aurora/system/github-app/private-key — identical to the current hardcoded constant.

server/utils/secrets/aws_sm_backend.py — add a system prefix constant near AWSSM_REF_PREFIX and override:
AWSSM_SYSTEM_PREFIX = "aurora/system"  # system secrets, separate from AWS_SM_PREFIX (aurora/users)

def build_system_ref(self, logical_name: str) -> str:
    # awssm:{region}:aurora/system/{logical_name}
    return f"{AWSSM_REF_PREFIX}{self.region}:{AWSSM_SYSTEM_PREFIX}/{logical_name}"
Don't reuse self.prefix (aurora/users) — system secrets get their own aurora/system namespace. get_secret() already reads the SecretId verbatim from the ref and validates the region, so no other changes are needed there.

2. Rewrite vault_keys.py to use it (and remove the env-var fallback)

server/connectors/github_connector/vault_keys.py:
- Replace the two hardcoded _..._SECRET_REF constants with logical names:
_PRIVATE_KEY_LOGICAL_NAME = "github-app/private-key"
_WEBHOOK_SECRET_LOGICAL_NAME = "github-app/webhook-secret"
- Rename _read_from_backend → _read_system_secret(logical_name, *, secret_label); build the ref from the active backend instead of taking a hardcoded ref, and drop the can_handle_ref gate (now redundant — the ref always comes from the active backend):
def _read_system_secret(logical_name: str, *, secret_label: str) -> str:
    backend = get_secrets_backend()
    if not backend.is_available():
        raise GitHubAppConfigError(
            f"Secrets backend is unavailable while reading GitHub App {secret_label}."
        )
    secret_ref = backend.build_system_ref(logical_name)
    try:
        secret_value = backend.get_secret(secret_ref)
    except GitHubAppConfigError:
        raise
    except Exception as exc:
        raise GitHubAppConfigError(
            f"Failed to read GitHub App {secret_label} from secrets backend."
        ) from exc
    if not secret_value:
        raise GitHubAppConfigError(
            f"GitHub App {secret_label} is empty in secrets backend."
        )
    return secret_value
- Delete _PRIVATE_KEY_ENV_VARS, _read_private_key_from_env(), and the env-fallback branch in get_app_private_key(). It becomes a straight backend read:
def get_app_private_key() -> str:
    global _cached_private_key
    if _cached_private_key is not None:
        return _cached_private_key
    _cached_private_key = _read_system_secret(
        _PRIVATE_KEY_LOGICAL_NAME, secret_label="private key",
    )
    return _cached_private_key
- Keep get_app_webhook_secret()'s existing env fallback (GITHUB_APP_WEBHOOK_SECRET etc.) — that predates this PR and isn't what we're removing. Just point it at _read_system_secret(_WEBHOOK_SECRET_LOGICAL_NAME, ...).

3. Revert the env/compose additions

- .env.example — remove the 3 added GITHUB_APP_PRIVATE_KEY lines.
- docker-compose.yaml and docker-compose.prod-local.yml — remove GITHUB_APP_PRIVATE_KEY: ${GITHUB_APP_PRIVATE_KEY} from the x-common-env block. (No new env vars are introduced by this approach.)

4. Operator provisioning (goes in the docs, not code)

On AWS SM deployments, the operator creates the two secrets — note the aurora/system/... prefix (NOT aurora/users/AWS_SM_PREFIX) and the region must match AWS_SM_REGION:
aws secretsmanager create-secret --name aurora/system/github-app/private-key \
  --secret-string file://github-app-private-key.pem --region "$AWS_SM_REGION"
aws secretsmanager create-secret --name aurora/system/github-app/webhook-secret \
  --secret-string 'your-webhook-secret' --region "$AWS_SM_REGION"
(Storing the PEM in SecretString keeps real newlines, so the \n-unescaping he added isn't needed anymore.)

5. Add the missing tests + docs (the PR has neither)

- Tests in server/tests/secrets/: build_system_ref returns the right ref for both backends; with the AWS SM backend mocked, get_app_private_key()/get_app_webhook_secret() resolve from it instead of raising "bad prefix"; Vault default still returns the identical ref (regression guard).
- Docs: website/docs/integrations/connectors.md (around L396, L442–443, L526) — show the aws secretsmanager create-secret commands next to the existing vault kv put ones; website/docs/configuration/environment.md (L396, L410) — reword "Vault path" → "secrets backend path (aurora/system/github-app/*)".

Verification (all three GitHub auth methods)

1. SECRETS_BACKEND=vault (default): get_app_private_key() returns the same value/ref as before — no regression.
2. SECRETS_BACKEND=aws_secrets_manager with the key stored at aurora/system/github-app/private-key: mint_app_jwt() succeeds and an installation token can be fetched (no more "bad prefix").
3. OAuth connect still works (unchanged); the agent/MCP GitHub tool resolves a token (passes once #2 works, since it falls back to App installation tokens).

OlivierTrudeau (approved):

PR #497, opened 2026-06-10
fix: RCA button navigating to action chat + show actions on incident page

OlivierTrudeau:
## What this fixes

1. **RCA button was broken** — clicking "Root Cause Analysis" on an incident took you to the action chat instead of the actual RCA. Actions were overwriting `aurora_chat_session_id` on the incident.
   - `server/chat/background/task.py` — guard the session/status writes with `if not is_action_source`

2. **Actions visible on incident page** — added an "Actions" tab (like Waterfall) so you can see all actions that ran for an incident and jump to their chat.
   - `server/routes/incidents_routes.py` — new `GET /api/incidents/<id>/action-runs` endpoint
   - `client/src/app/incidents/components/IncidentActionRuns.tsx` — new component
   - `client/src/app/incidents/components/IncidentCard.tsx` — added the tab button + collapsible panel
   - `client/src/app/api/incidents/[id]/action-runs/route.ts` — Next.js proxy route

3. **Postmortem generation spinner was broken** — resolving an incident triggered postmortem gen but the UI would stop spinning and show "Generate Postmortem" button while it was still generating in the background.
   - Root cause: backend returns HTTP 202 with `{status: "generating"}`, but `apiGet` treats 202 as success (not an error). The catch block checking `apiErr.status === 202` was dead code.
   - `client/src/lib/services/incidents.ts` — check `data.status === 'generating'` in the success path

4. **Actions dispatching was broken** — `dispatch_on_incident_actions` had a UUID type mismatch (`uuid = text`) in the dedup query that crashed the entire function silently, preventing all on-incident actions (including postmortem gen on resolve) from firing.
   - `server/services/actions/executor.py` — cast to `::uuid[]` / `::uuid` in the dedup query

## Summary by CodeRabbit

* **New Features**
  * Collapsible "Actions" panel on incident cards with animated chevron.
  * Action runs list showing status (live indicator when running), action name, duration, start time, truncated error tooltip, and "View" link to related chat sessions.
  * Loading and empty states ("Loading actions..." / "No actions have run for this incident.").

* **New API**
  * Endpoint to fetch action runs for an incident.

* **Bug Fixes**
  * Prevent duplicate action deliveries and avoid triggering incident-side workflows for action-originated runs.
  * Avoid dispatching duplicate action runs; postmortem generation state correctly reported.

### DIFF
--- a/client/src/app/incidents/components/IncidentActionRuns.tsx
+++ b/client/src/app/incidents/components/IncidentActionRuns.tsx
@@ -1,0 +1,130 @@
+'use client';
+
+import { useEffect, useState } from 'react';
+import Link from 'next/link';
+import { ExternalLink, CheckCircle2, AlertCircle, Clock } from 'lucide-react';
+
+interface ActionRun {
+  id: string;
+  action_id: string;
+  action_name: string;
+  status: string;
+  chat_session_id: string | null;
+  started_at: string | null;
+  completed_at: string | null;
+  duration_ms?: number;
+  error: string | null;
+}
+
+interface IncidentActionRunsProps {
+  readonly incidentId: string;
+}
+
+function StatusIcon({ status }: { readonly status: string }) {
+  switch (status) {
+    case 'success':
+      return <CheckCircle2 className="w-3.5 h-3.5 text-green-400" />;
+    case 'error':
+      return <AlertCircle className="w-3.5 h-3.5 text-red-400" />;
+    case 'running':
+      return (
+        <span className="relative flex h-3.5 w-3.5">
+          <span className="animate-ping absolute inline-flex h-full w-full rounded-full bg-orange-400 opacity-75"></span>
+          <span className="relative inline-flex rounded-full h-3.5 w-3.5 bg-orange-500"></span>
+        </span>
+      );
+    default:
+      return <Clock className="w-3.5 h-3.5 text-zinc-500" />;
+  }
+}
+
+function formatDuration(ms: number): string {
+  if (ms < 1000) return `${ms}ms`;
+  const seconds = Math.round(ms / 1000);
+  if (seconds < 60) return `${seconds}s`;
+  const minutes = Math.floor(seconds / 60);
+  const remainingSeconds = seconds % 60;
+  return `${minutes}m ${remainingSeconds}s`;
+}
+
+function formatTime(isoString: string): string {
+  const date = new Date(isoString);
+  return date.toLocaleTimeString([], { hour: '2-digit', minute: '2-digit' });
+}
+
+export default function IncidentActionRuns({ incidentId }: IncidentActionRunsProps) {
+  const [runs, setRuns] = useState<ActionRun[]>([]);
+  const [loading, setLoading] = useState(true);
+
+  useEffect(() => {
+    let cancelled = false;
+    async function fetchRuns() {
+      try {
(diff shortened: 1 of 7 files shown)


client/src/app/incidents/components/IncidentActionRuns.tsx
```diff
@@ -0,0 +1,130 @@
...
+        if (!cancelled) setRuns(data.runs || []);
+      } catch {
+        // Silently fail
+      } finally {
+        if (!cancelled) setLoading(false);
+      }
+    }
+    fetchRuns();
+    return () => { cancelled = true; };
+  }, [incidentId]);
```
damianloch, line 95: so this doesnt update unless navigate away and back?

client/src/app/incidents/components/IncidentActionRuns.tsx
```diff
@@ -0,0 +1,130 @@
...
+  useEffect(() => {
+    let cancelled = false;
+    async function fetchRuns() {
+      try {
+        const res = await fetch(`/api/incidents/${incidentId}/action-runs`);
+        if (!res.ok) return;
+        const data = await res.json();
+        if (!cancelled) setRuns(data.runs || []);
+      } catch {
+        // Silently fail
```
damianloch: On fetch failure, falls through to showing "No actions have run", which is misleading if they actually did

server/services/actions/executor.py
```diff
@@ -117,13 +117,28 @@ def dispatch_on_incident_actions(user_id: str, incident_id: str, timing: str = "
...
+            already_running = set()
+            if action_ids:
+                cur.execute(
+                    """SELECT DISTINCT action_id FROM action_runs
+                       WHERE action_id = ANY(%s::uuid[]) AND incident_id = %s::uuid
+                         AND status IN ('pending', 'running')""",
+                    (action_ids, incident_id),
+                )
+                already_running = {str(r[0]) for r in cur.fetchall()}
+
```
damianloch: cursor is telling me this: The dedup query checks for status IN ('pending', 'running') and then dispatches outside the with block (connection is released). Between the SELECT and the dispatch, another concurrent call to dispatch_on_incident_actions (e.g. from a retry or duplicate webhook) could pass the same check and dispatch the same action — classic TOCTOU. The PR description mentions this was the original bug (the UUID cast fix), but the new dedup is still advisory, not atomic.

damianloch (changes requested):
Need to test as well

server/services/actions/executor.py
OlivierTrudeau: Was the same comment as the coderabbit comment. Addressed it by locking with action id and incident id

client/src/app/incidents/components/IncidentActionRuns.tsx
OlivierTrudeau: No says failed to fetch

client/src/app/incidents/components/IncidentActionRuns.tsx
OlivierTrudeau: Polls when there's a running action

damianloch (approved):

client/src/app/incidents/components/IncidentActionRuns.tsx
damianloch: YOURE ABSOLUTELY RIGHT

PR #508, opened 2026-06-13
fix: Slack OAuth token loss for large workspaces

Harrio-6:
Rewrite create_incidents_channel to always create a new channel with fallback naming (incidents → aurora_incidents → aurora_incidents_<hex>) instead of searching for existing ones
Add 429 retry logic with Retry-After support to _make_request
Remove list_channels() method (conversations.list) that paginated all workspace channels
Remove unused /slack/channels GET endpoint
Update error hint to reflect create-only behavior

## Summary

* **Bug Fixes**
  * Improved Slack reliability with automatic retries for rate limits and transient request errors, plus better handling of Slack API “ok” responses.

* **Improvements**
  * Channel listings now show only channels the Slack bot is a member of.
  * Incidents-channel setup during Slack connection is more resilient: it reuses a previously stored channel when possible or creates/join an appropriate one.
  * OAuth redirects now provide clearer, encoded guidance when incidents-channel creation fails.

* **Chores**
  * Removed the client-side Slack channel listing endpoint and related channel-listing method.


server/connectors/slack_connector/client.py
damianloch, line 1: Lets spam logs here so that we can at least further debug if other users have issues going forward

damianloch (changes requested):

OlivierTrudeau (approved):

damianloch (approved):

PR #513, opened 2026-06-15
fix: close MCP subprocess pipe handles to eliminate ResourceWarning fd leak (INC-538)

beng360:
## Problem

The `aurora-chatbot` container emits `ResourceWarning: unclosed file <_io.BufferedWriter/BufferedReader>` approximately every 10 minutes as the MCP preloader re-initializes the GitHub MCP server and Context7 subprocesses. GCP Cloud Logging captures these warnings at **ERROR severity** (because they arrive on stderr), inflating the error count and contributing to the **"Aurora Staging - Error Log Spike"** alert threshold breach.

This alert has now fired **11 times** (INC-538, 2026-06-15 19:50 UTC) without this underlying code defect being fixed.

## Root Cause

`RealMCPServerManager.start_mcp_server()` creates subprocesses with:
```python
subprocess.Popen(cmd, stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE, ...)
```

In **four code paths**, the `Popen` object is discarded (via `terminate()` / `kill()` + `del`) **without explicitly closing the three pipe file handles**. When Python's garbage collector later collects the `Popen` object, it emits one `ResourceWarning` per unclosed handle (up to 3 per discarded process). These land on stderr → GCP assigns ERROR severity → alert fires.

## Fix

Add explicit pipe closure in all four affected paths in `server/chat/backend/agent/tools/mcp_tools.py`:

### 1. `cleanup()` — after terminate/kill, before `server_processes.clear()`
```python
# Always close pipe handles to avoid ResourceWarning on GC
if process:
    for pipe in (process.stdin, process.stdout, process.stderr):
        if pipe is not None:
            try:
                pipe.close()
            except Exception:
                pass
```

### 2. `start_mcp_server()` — force-restart path (github/aws credential refresh, ~line 175)
Same pipe-close block added after `process.wait()` / `process.kill()`, before `del self.server_processes[server_type]`.

### 3. `start_mcp_server()` — AWS dead-process cleanup block (~line 208)
Same pipe-close block added after the terminate/kill sequence, before `del self.server_processes[server_type]`.

### 4. `start_mcp_server()` — failed-start path after `communicate()` (~line 346)
```python
# communicate() drains the pipes but does not close the underlying
# file handles — close them explicitly to suppress ResourceWarning.
for pipe in (process.stdin, process.stdout, process.stderr):
    if pipe is not None:
        try:
            pipe.close()
        except Exception:
            pass
```

[...]

PR #523, opened 2026-06-17
Add file-size guard on artifact creation via MCP GitHub tools

### LINKED ISSUE #521: Add file-size guard to mcp_create_or_update_file
## Problem

On June 15, staging Aurora's agent used `mcp_create_or_update_file` to rewrite `mcp_tools.py` (~81KB). The LLM hit its output token budget and only generated ~2KB of the file header/imports. This truncated stub was pushed to GitHub, destroying the file and producing broken PR #513.

The GitHub Contents API requires the **full file content** in the request body. When the LLM can't fit the entire file in its output, it silently truncates — there's no error, just a partial file committed.

## Root Cause

`mcp_create_or_update_file` has no validation on the content being written relative to the file it's replacing. A 2KB replacement for an 81KB file should be rejected.

## Proposed Fix

1. Before writing, fetch the current file size from GitHub (or use the SHA lookup already being done)
2. If the new content is less than ~50% of the original file size AND the original is above a threshold (e.g., 10KB), reject the write with an error message instructing the agent to use incremental edits instead
3. Add an absolute cap (e.g., 50KB) on content passed through `mcp_create_or_update_file` — files larger than this should always use `push_files` with git or line-level edits

## Evidence

GCP logs , celery-worker, 2026-06-15 20:09-20:12 UTC) show:
- `mcp_create_or_update_file` called at 20:09:02
- LLM streamed only the file header as `chat_output`
- Agent retried at 20:11:03 (duplicate detected)
- Agent then tried git clone fallback, noting "The mcp_push_files tool should handle large files"
- Truncated content was already committed by then

totto:
## Summary

When the LLM hits its output token limit, `create_or_update_file` silently pushes a truncated stub to GitHub, destroying the original file (the pattern that caused PR #513 on staging). This adds a pre-flight guard inside the MCP tool wrapper that:

- Rejects content exceeding a 50 KB absolute cap for both `create_or_update_file` and `push_files` (per-entry)
- For file updates (`sha` present), fetches the existing file size via `get_file_contents` and rejects writes smaller than 50% of the original when the original exceeds 10 KB
- Fail-open: guard failures are logged and the write proceeds

The guard runs after the HITL confirmation gate but before the MCP server call, so it catches both foreground and background writes.

## Test plan
- [ ] Verify `create_or_update_file` with content > 50 KB is rejected
- [ ] Verify `create_or_update_file` updating a 20 KB file with 5 KB content (`sha` present) is rejected with ratio error
- [ ] Verify `create_or_update_file` creating a new file (no `sha`) is not subject to ratio check
- [ ] Verify `push_files` with a per-entry > 50 KB is rejected
- [ ] Verify guard failure (e.g., `get_file_contents` timeout) allows the write through

Closes #521

## Summary by CodeRabbit

* **New Features**
  * Added a file size safety check for GitHub repository write operations to help prevent truncated or suspiciously sized content from overwriting existing files.
  * When a write is flagged, the operation is blocked and a notification is returned instead of executing the update.

### DIFF
--- a/server/chat/backend/agent/tools/mcp_tools.py
+++ b/server/chat/backend/agent/tools/mcp_tools.py
@@ -66,6 +66,113 @@ def is_destructive_mcp_tool(tool_name: str) -> bool:
     return False
 
 
+# ---------------------------------------------------------------------------
+# File-size guard for GitHub write MCP tools (issue #521)
+# ---------------------------------------------------------------------------
+# When the LLM's output is silently truncated, create_or_update_file pushes
+# a partial file to GitHub, destroying the original.  The guard rejects
+# writes where the new content is suspiciously small relative to the
+# existing file or exceeds a hard size cap.
+_FILE_SIZE_GUARD_ABS_CAP = 50_000       # reject content > 50 KB outright
+_FILE_SIZE_GUARD_MIN_ORIGINAL = 10_000   # only compare ratio when original > 10 KB
+_FILE_SIZE_GUARD_MIN_RATIO = 0.50        # new must be >= 50% of original
+
+
+def _check_file_size_guard(
+    original_tool_name: str,
+    kwargs: dict,
+    manager: "RealMCPServerManager",
+    run_async,
+) -> str | None:
+    """Return an error string if the write should be blocked, else ``None``.
+
+    For ``create_or_update_file``: if the file already exists on GitHub
+    (``sha`` is present), fetches the current size via ``get_file_contents``.
+    Rejects when the new content is < 50% of the original (and the original
+    exceeds 10 KB) -- the signature of a truncated LLM output.
+
+    For ``push_files``: applies only the absolute 50 KB cap per file entry.
+    """
+    if original_tool_name == "create_or_update_file":
+        content = kwargs.get("content", "")
+        new_size = len(content.encode("utf-8", errors="replace")) if isinstance(content, str) else 0
+
+        # Hard cap
+        if new_size > _FILE_SIZE_GUARD_ABS_CAP:
+            return (
+                f"Blocked: content is {new_size:,} bytes which exceeds the "
+                f"{_FILE_SIZE_GUARD_ABS_CAP:,}-byte safety cap for create_or_update_file. "
+                "Use push_files with git or incremental edits for large files."
+            )
+
+        # Ratio check -- only when updating an existing file (sha present)
+        owner = kwargs.get("owner")
+        repo = kwargs.get("repo")
+        path = kwargs.get("path")
+        branch = kwargs.get("branch")
+        sha = kwargs.get("sha")
+
+        if sha and owner and repo and path:
+            try:
+                get_kwargs: dict = {"owner": owner, "repo": repo, "path": path}
+                if branch:
+                    get_kwargs["branch"] = branch
+                result = run_async(
+                    manager.call_mcp_tool("github", "get_file_contents", get_kwargs)
+                )
+                existing_size = _extract_github_file_size(result)
+                if (
+                    existing_size is not None
+                    and existing_size > _FILE_SIZE_GUARD_MIN_ORIGINAL
+                    and new_size < existing_size * _FILE_SIZE_GUARD_MIN_RATIO
+                ):
+                    return (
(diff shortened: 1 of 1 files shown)


beng360:
hey @totto, can you answer the coderabbit comments please? Thanks (same thing for your other PRs)

server/chat/backend/agent/tools/mcp_tools.py
```diff
@@ -66,6 +66,125 @@ def is_destructive_mcp_tool(tool_name: str) -> bool:
...
+
+def _check_file_size_guard(
+    original_tool_name: str,
+    kwargs: dict,
+    manager: "RealMCPServerManager",
+    run_async,
+) -> str | None:
+    """Return an error string if the write should be blocked, else ``None``."""
+    if original_tool_name == "create_or_update_file":
+        return _guard_create_or_update_file(kwargs, manager, run_async)
```
beng360, line 89: The guard reads `kwargs.get("content")` but at invocation time the LLM can pass args nested inside `kwargs["kwargs"]` (the wrapper unwraps this at line 1326 below). Move the guard after the `actual_kwargs` extraction, or unwrap before calling `_check_file_size_guard`. Otherwise the guard silently passes everything on the nested-args shape.

server/chat/backend/agent/tools/mcp_tools.py
```diff
@@ -66,6 +66,125 @@ def is_destructive_mcp_tool(tool_name: str) -> bool:
     return False
 
 
+# ---------------------------------------------------------------------------
+# File-size guard for GitHub write MCP tools (issue #521)
+# ---------------------------------------------------------------------------
+# When the LLM's output is silently truncated, create_or_update_file pushes
```
beng360, line 72: 50 KB cap is too low. Our codebase has legitimate files in the 50-100 KB range (large config YAMLs, generated schemas). A user doing `create_or_update_file` on those will get blocked every time. Consider 100 KB or 150 KB — the real danger (truncated LLM output) is more like 5-15 KB stubs overwriting 80 KB files, which the ratio check already catches.

server/chat/backend/agent/tools/mcp_tools.py
```diff
@@ -66,6 +66,125 @@ def is_destructive_mcp_tool(tool_name: str) -> bool:
...
+
+    if new_size > _FILE_SIZE_GUARD_ABS_CAP:
+        return (
+            f"Blocked: content is {new_size:,} bytes which exceeds the "
+            f"{_FILE_SIZE_GUARD_ABS_CAP:,}-byte safety cap for create_or_update_file. "
+            "Use push_files with git or incremental edits for large files."
+        )
+
+    owner = kwargs.get("owner")
+    repo = kwargs.get("repo")
```
beng360, line 117: `get_file_contents` MCP call here has no timeout. If the GitHub API is slow this blocks the entire tool execution indefinitely. Pass a timeout to `run_async` or wrap in `asyncio.wait_for`. Also — this fires on every update to a file >10 KB, adding latency to the hot path. Consider caching or making the ratio check opt-in via an env var.

server/chat/backend/agent/tools/mcp_tools.py
```diff
@@ -66,6 +66,125 @@ def is_destructive_mcp_tool(tool_name: str) -> bool:
...
+        logging.warning(
+            "[FileGuard] Could not fetch existing file for %s/%s/%s -- "
+            "skipping ratio check: %s", owner, repo, path, exc,
+        )
+
+    return None
+
+
+def _guard_push_files(kwargs: dict) -> str | None:
+    """Size guard for ``push_files``: absolute cap per file entry."""
```
beng360, line 155: The `import json as _json` inside the loop body runs on every iteration. Move it to module-level or at least outside the loop. Also the outer try/except wraps the entire loop — if the first `content` item is a status text line (not JSON), the `json.loads` raises and the loop exits without checking remaining items that might have the metadata. Narrow the try/except to the parse line.

beng360 (changes requested):
Good idea, this guard would have prevented the #513 incident. A few issues to fix before merging.

beng360:

PR #524, opened 2026-06-17
Strip GitHub write MCP tools from non-RCA chat sessions

### LINKED ISSUE #522: Non-RCA sessions bypass ModeAccessController write restrictions
## Problem

On June 15, a non-RCA session on staging had `push_files` and `create_or_update_file` MCP tools loaded without read-only restrictions. This allowed the agent to push a truncated file directly to GitHub (PR #513), destroying `mcp_tools.py`.

Logs show `github_fix=excluded (not RCA)` for this session — meaning it correctly excluded the safe `github_fix` tool (which uses line-level edits). But it still had raw MCP GitHub write tools available, which are far more dangerous.

## Evidence from GCP logs

- **20:38 UTC** — A different session properly blocked `mcp_push_files` via `ModeAccessController` (read-only mode enforced)
- **20:09-20:11 UTC** — The session that created PR #513 had `mcp_create_or_update_file` available without any read-only restriction

This inconsistency suggests the `ModeAccessController` enforcement depends on session configuration that isn't always applied to non-RCA sessions.

## Proposed Fix

1. Non-RCA sessions should NEVER have `push_files`, `create_or_update_file`, or `create_branch` tools loaded — these should be exclusive to RCA sessions that go through the `github_fix` → `github_apply_fix` flow
2. If raw GitHub write tools must be available for specific use cases, they should always go through `ModeAccessController` with read-only mode as the default
3. Add an integration test that verifies non-RCA sessions cannot access GitHub write tools

## Impact

PR #513 replaced the entire 81KB `mcp_tools.py` with a ~2KB truncated stub. The PR was caught and closed manually, but if merged would have broken all MCP functionality.

totto:
## Summary

`ModeAccessController.filter_tools()` only gates tools in `"ask"` mode. Non-RCA sessions running in `"agent"` or default mode were getting raw GitHub write MCP tools (`create_or_update_file`, `push_files`, `create_branch`, etc.) without the structured `github_fix` guardrails that RCA sessions use. This was a contributing factor in the staging incident (PR #513).

The fix adds a `_GITHUB_MCP_WRITE_TOOLS` frozenset (defined alongside the existing `_NON_RCA_SOURCES` constant) and strips those tools in `get_cloud_tools()` when `is_rca_context is False`. RCA sessions continue to receive all tools.

## Test plan
- [ ] Verify non-RCA interactive chat session does not have `mcp_create_or_update_file`, `mcp_push_files`, or `mcp_create_branch` in its tool list
- [ ] Verify RCA background session retains all MCP tools including write tools
- [ ] Verify log message `"Stripped N GitHub write MCP tools from non-RCA session"` appears for non-RCA sessions with GitHub connected
- [ ] Verify read-only MCP tools (`mcp_get_file_contents`, `mcp_list_commits`, etc.) are still available in non-RCA sessions

Closes #522

## Summary by CodeRabbit

* **Bug Fixes**
  * Restricted certain GitHub write operations to Root Cause Analysis (RCA) sessions only. Non-RCA users will no longer have access to these tools, with enhanced logging showing how many tools are filtered per session.

### DIFF
--- a/server/chat/backend/agent/tools/cloud_tools.py
+++ b/server/chat/backend/agent/tools/cloud_tools.py
@@ -981,6 +981,23 @@ from .mcp_tools import (
 
 _NON_RCA_SOURCES = frozenset(('action', 'prediscovery'))
 
+# GitHub write MCP tools that MUST be restricted to RCA sessions (issue #522).
+# Non-RCA sessions (interactive chat, ask, actions) should never have these
+# loaded -- they bypass ModeAccessController (which only gates "ask" mode) and
+# can push destructive changes without the structured github_fix guardrails.
+# Tool names match the StructuredTool naming: "mcp_{original_tool_name}".
+_GITHUB_MCP_WRITE_TOOLS = frozenset({
+    "mcp_create_or_update_file",
+    "mcp_push_files",
+    "mcp_create_branch",
+    "mcp_create_pull_request",
+    "mcp_merge_pull_request",
+    "mcp_delete_file",
+    "mcp_create_repository",
+    "mcp_fork_repository",
+    "mcp_update_pull_request_branch",
+})
+
 
 def _is_background_rca(state_context, is_background: bool) -> bool:
     """True when the session is a background RCA investigating a real incident.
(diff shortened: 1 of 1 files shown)


server/chat/backend/agent/tools/cloud_tools.py
```diff
@@ -981,6 +981,23 @@ def get_current_tool_call_id(tool_name: str = None, tool_kwargs: dict = None):
...
+# GitHub write MCP tools that MUST be restricted to RCA sessions (issue #522).
+# Non-RCA sessions (interactive chat, ask, actions) should never have these
+# loaded -- they bypass ModeAccessController (which only gates "ask" mode) and
+# can push destructive changes without the structured github_fix guardrails.
+# Tool names match the StructuredTool naming: "mcp_{original_tool_name}".
+_GITHUB_MCP_WRITE_TOOLS = frozenset({
+    "mcp_create_or_update_file",
+    "mcp_push_files",
+    "mcp_create_branch",
+    "mcp_create_pull_request",
```
beng360, line 993: The frozenset should also include `mcp_create_or_update_file_with_pr` — this tool variant exists in the GitHub MCP server and also does writes. Check `get_github_tool_schemas()` in `mcp_schema_extractor.py` for the full list of write tools exposed. Missing one here means the guard has a bypass.

server/chat/backend/agent/tools/cloud_tools.py
```diff
@@ -2560,13 +2577,31 @@ def _pinned_trigger(action_id: str = "", _pid=pinned_id, _fn=final_func, **kw):
...
                 run_async_in_thread=run_async_in_thread
             )
+
+            # Issue #522: strip GitHub write MCP tools from non-RCA sessions.
+            # RCA sessions use the structured github_fix -> github_apply_fix
+            # flow with line-level edits.  Non-RCA sessions must not have raw
+            # write tools that bypass ModeAccessController and github_fix.
+            if not is_rca_context:
+                before_count = len(mcp_tools)
+                mcp_tools = [
```
beng360, line 2594: This filter runs before `ModeAccessController.filter_tools()` at line 2602, which is fine — but the log message should include the `mode` so we can distinguish ask-mode filtering (which also strips MCP tools) from this RCA-context filter in debugging. Something like `"Stripped %d GitHub write MCP tools from non-RCA session (user=%s, mode=%s)"`.

beng360 (changes requested):
Correct approach. Two issues to address.

beng360:

PR #527, opened 2026-06-17
Expand Vertex AI support for Gemini models

beng360:
## Summary
- Add Vertex-routed Gemini models (3.1 Pro, 3 Flash, 2.5 Pro, 2.5 Flash) to model registry, UI selector, pricing, and context managers
- Fix guardrails crash when `LLM_PROVIDER_MODE=vertex` — NeMo calls `.bind(max_tokens=3)` but Gemini only accepts `max_output_tokens`, now properly wrapped
- Pass `GOOGLE_APPLICATION_CREDENTIALS` through docker-compose for file-based Vertex auth

## Test plan
- [x] Model mapper: reverse mapping collision fixed (vertex entries no longer overwrite google entries)
- [x] Provider routing: vertex/gemini-* routes through VertexAIProvider, anthropic/* falls back to direct
- [x] Guardrails: wrapper correctly applied to google/vertex models, NOT applied to anthropic/openai under vertex mode
- [x] End-to-end: invoked `vertex/gemini-2.5-flash` via Vertex AI on `sublime-flux-414616` — response received
- [x] Guardrails e2e: `_GuardrailsLLMCompat` with `rename_max_tokens=True` + `.bind(max_tokens=50)` returns plain string (thinking blocks stripped)
- [x] Existing test suite: 191/191 pass (1 pre-existing RBAC failure unrelated)

Closes DEV-1281

## Release Notes

* **New Features**
  * Added Vertex AI provider support with Gemini 3.5 Flash and 3.1 Pro Preview model options
  * Enhanced model selection and provider switching capabilities
  * Improved model compatibility across multiple providers

* **Documentation**
  * Updated LLM provider documentation with latest model examples and configurations

### DIFF
--- a/client/src/components/ModelSelector.tsx
+++ b/client/src/components/ModelSelector.tsx
@@ -41,10 +41,8 @@ const modelPricing: Record<string, string> = {
   'openai/gpt-5.5': 'Premium Cost ($5/$30 per 1M)',
   'anthropic/claude-sonnet-4.6': 'Medium Cost ($3/$15 per 1M)',
   'anthropic/claude-opus-4.7': 'High Cost ($5/$25 per 1M)',
+  'vertex/gemini-3.5-flash': 'Low Cost ($0.50/$3 per 1M)',
   'vertex/gemini-3.1-pro-preview': 'Medium Cost ($2/$12 per 1M)',
-  'vertex/gemini-3-flash': 'Low Cost ($0.50/$3 per 1M)',
-  'vertex/gemini-2.5-pro': 'Medium Cost ($1.25/$10 per 1M)',
-  'vertex/gemini-2.5-flash': 'Low Cost ($0.30/$2.50 per 1M)',
 };
 
 const modelOptions: ModelOption[] = [
@@ -77,38 +75,20 @@ const modelOptions: ModelOption[] = [
     isSlow: true
   },
   {
+    id: 'vertex/gemini-3.5-flash',
+    name: 'gemini-3.5-flash',
+    displayName: 'Gemini 3.5 Flash',
+    provider: 'Vertex AI',
+    tier: 'free',
+    contextLength: '1M',
+    hasReasoning: true
+  },
+  {
     id: 'vertex/gemini-3.1-pro-preview',
     name: 'gemini-3.1-pro-preview',
     displayName: 'Gemini 3.1 Pro',
     provider: 'Vertex AI',
     tier: 'pro',
-    contextLength: '1M',
-    hasReasoning: true
-  },
-  {
-    id: 'vertex/gemini-3-flash',
-    name: 'gemini-3-flash',
-    displayName: 'Gemini 3 Flash',
-    provider: 'Vertex AI',
-    tier: 'free',
-    contextLength: '1M',
-    hasReasoning: true
-  },
-  {
-    id: 'vertex/gemini-2.5-pro',
-    name: 'gemini-2.5-pro',
-    displayName: 'Gemini 2.5 Pro',
-    provider: 'Vertex AI',
-    tier: 'pro',
-    contextLength: '1M',
-    hasReasoning: true
-  },
-  {
-    id: 'vertex/gemini-2.5-flash',
-    name: 'gemini-2.5-flash',
-    displayName: 'Gemini 2.5 Flash',
-    provider: 'Vertex AI',
-    tier: 'free',
     contextLength: '1M',
     hasReasoning: true
   },
--- a/client/src/components/ModelSelector.tsx
+++ b/client/src/components/ModelSelector.tsx
@@ -41,8 +41,10 @@ const modelPricing: Record<string, string> = {
   'openai/gpt-5.5': 'Premium Cost ($5/$30 per 1M)',
   'anthropic/claude-sonnet-4.6': 'Medium Cost ($3/$15 per 1M)',
   'anthropic/claude-opus-4.7': 'High Cost ($5/$25 per 1M)',
-  'vertex/gemini-3.5-flash': 'Low Cost ($0.50/$3 per 1M)',
-  'vertex/gemini-3.1-pro-preview': 'Medium Cost ($2/$12 per 1M)',
+  'google/gemini-3.5-flash': 'Low Cost ($0.50/$3 per 1M)',
+  'google/gemini-3.1-pro-preview': 'Medium Cost ($2/$12 per 1M)',
+  'google/gemini-2.5-pro': 'Medium Cost ($1.25/$10 per 1M)',
+  'google/gemini-2.5-flash': 'Low Cost ($0.30/$2.50 per 1M)',
 };
 
 const modelOptions: ModelOption[] = [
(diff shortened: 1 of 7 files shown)


docker-compose.yaml
```diff
@@ -94,6 +94,7 @@ x-common-env: &common-env
   VERTEX_AI_PROJECT: ${VERTEX_AI_PROJECT}
   VERTEX_AI_LOCATION: ${VERTEX_AI_LOCATION}
   VERTEX_AI_SERVICE_ACCOUNT_JSON: ${VERTEX_AI_SERVICE_ACCOUNT_JSON}
+  GOOGLE_APPLICATION_CREDENTIALS: ${GOOGLE_APPLICATION_CREDENTIALS}
```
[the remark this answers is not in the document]
beng360: Good catch. Added GOOGLE_APPLICATION_CREDENTIALS to docker-compose.prod-local.yml and .env.example.

server/chat/backend/agent/utils/llm_usage_tracker.py
```diff
@@ -331,7 +331,19 @@ def store_usage(cls, usage: LLMUsage) -> bool:
...
+                # Implementations must be non-blocking (buffer + flush pattern).
+                try:
+                    from utils.hooks import get_hook
+                    get_hook("report_usage")(usage.org_id, usage.estimated_cost, {
+                        "model": usage.model_name,
+                        "input_tokens": usage.input_tokens,
+                        "output_tokens": usage.output_tokens,
+                    })
+                except Exception as hook_err:
+                    logger.debug("report_usage hook error (non-fatal): %s", hook_err)
```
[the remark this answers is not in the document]
beng360: This code is no longer in the PR after rebase — the metering hooks commits were removed. Only the Vertex support commit remains.

server/guardrails/input_rail.py
```diff
@@ -151,9 +151,9 @@ def _build_llm() -> BaseChatModel:
     llm = create_chat_model(model_name, temperature=0.0, streaming=False)
 
     provider, _ = ModelMapper.split_provider_model(model_name)
-    provider_mode = os.getenv("LLM_PROVIDER_MODE", "direct").lower()
-    if provider == "google" and provider_mode == "direct":
+    if provider in ("google", "vertex"):
         return _GuardrailsLLMCompat(inner=llm, rename_max_tokens=True)
+    provider_mode = os.getenv("LLM_PROVIDER_MODE", "direct").lower()
```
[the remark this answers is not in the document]
beng360: The prefix-based check is intentional here. The guardrails model is selected by GUARDRAILS_LLM_MODEL (or MAIN_MODEL fallback) and the wrapper decision is based on whether that model is a Gemini model — which is determined by its prefix. The actual provider routing happens inside create_chat_model() which reads LLM_PROVIDER_MODE from env. If the model has a google/ or vertex/ prefix, it will always be served by a Gemini-compatible provider (either Google direct or Vertex), so the rename is always correct. We explicitly avoided using provider_mode == "vertex" here because that would incorrectly wrap non-Gemini models (e.g. anthropic/ model used as a guardrails override under vertex mode).

deploy/helm/aurora/values.yaml
```diff
@@ -107,6 +107,17 @@ autoscaling:
     maxReplicas: 6
     targetCPU: 70
 
+# ------------------------------------------------------------------------------
```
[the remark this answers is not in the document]
beng360: This code is no longer in the PR after rebase — the extraEnv/Helm changes were in the metering hooks commits which have been removed. Only the Vertex support commit remains.

docker-compose.yaml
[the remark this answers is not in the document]
beng360: Removed GOOGLE_APPLICATION_CREDENTIALS from all compose files and .env.example — it was only needed for local testing with file-based credentials. Production uses VERTEX_AI_SERVICE_ACCOUNT_JSON which is already wired up.

server/guardrails/input_rail.py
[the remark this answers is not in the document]
beng360: Fixed in a3c84403. Now uses resolve_provider_name() to check the actual serving provider instead of the model-name prefix. Under LLM_PROVIDER_MODE=openrouter, a google/ model routes through OpenRouter (ChatOpenAI) which expects max_tokens, so the rename is correctly skipped.

server/chat/backend/agent/model_mapper.py
```diff
@@ -47,8 +47,47 @@
...
+        "vertex": "gemini-3.1-pro-preview",
+        "provider": "vertex",
+    },
+    "google/gemini-3-flash": {
+        "openrouter": "google/gemini-3-flash",
+        "google": "gemini-3-flash",
+        "vertex": "gemini-3-flash",
+        "provider": "google",
+    },
+    "vertex/gemini-3-flash": {
```
damianloch: im getting 404 for this on vertex, gemini-3-flash-preview is a valid one apparently

client/src/components/ModelSelector.tsx
damianloch, line 1: there is a 3.5 flash I think that should be the flash model to use

server/chat/backend/agent/providers/vertex_provider.py
damianloch, line 1: vertex is super slow any idea why or if we can do smtg about this

server/guardrails/input_rail.py
damianloch, line 1: now google models via openrouter dont work with guaurdrails

client/src/components/ModelSelector.tsx
damianloch, line 1: Update documentation with supported models for vertex + anything else missing for vertex setup for OSS users

damianloch (changes requested):

server/chat/backend/agent/providers/vertex_provider.py
beng360: Fixed — thinking mode was the culprit. Gemini 3.5 flash defaults to medium thinking which adds ~3s per LLM call. Set it to minimal (thinking_level=minimal), brings it down to ~1s per call. RCA completes in ~7min now vs timing out at 30min before. Also added GEMINI_THINKING_LEVEL env var to override. This led to great RCAs on vertex.

client/src/components/ModelSelector.tsx
beng360: yeah added it

beng360: Updated — refreshed the model list in llm-providers.md (removed old 2.5/3-flash entries, added gemini-3.5-flash for both google/ and vertex/ prefixes), and documented the GEMINI_DISABLE_THINKING env var.

server/chat/backend/agent/model_mapper.py
beng360: fixed this

server/guardrails/input_rail.py
beng360: fixed; issue was we weren't detecting open router models properly. now we use resolve_provider_name() to check actual serving provider, and only rename when its actually Google/Vertex

client/src/components/ModelSelector.tsx
damianloch: nuh uh

damianloch (dismissed):

server/chat/backend/agent/utils/model_cutoff_manager.py
```diff
@@ -64,11 +64,32 @@ def cutoff_date(year: int, month: int, day: int = 1) -> datetime:
...
+            ),
+            "vertex/gemini-3-flash": ModelInfo(
+                "gemini-3-flash", "vertex", cutoff_date(2025, 4, 1), True, True
+            ),
+            "vertex/gemini-2.5-pro": ModelInfo(
+                "gemini-2.5-pro", "vertex", cutoff_date(2025, 3, 1), True, True
+            ),
+            "vertex/gemini-2.5-flash": ModelInfo(
+                "gemini-2.5-flash", "vertex", cutoff_date(2025, 3, 1), True, True
+            ),
```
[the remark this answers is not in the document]
beng360: Added gemini-3.5-flash entries for both google/ and vertex/ prefixes, and removed the stale gemini-3-flash entries.

client/src/components/ModelSelector.tsx
```diff
@@ -102,6 +112,7 @@ export default function ModelSelector({
       onModelChange(savedModel);
     } else if (!isValidModel) {
       localStorage.removeItem('selectedModel');
+      onModelChange(modelOptions[0].id);
     }
```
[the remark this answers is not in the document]
beng360: Fixed. Now only resets to modelOptions[0] when the current selectedModel prop is also invalid.

damianloch (approved):

PR #538, opened 2026-06-19
fix: remove noisy cloud provider warning on chat send

### LINKED ISSUE #518: Fix: 'No cloud provider connected' toast fires incorrectly on every message send
## Problem

The "No cloud provider selected. Connect a provider for reliable execution." toast fires on every single message send, even when the user has a cloud provider (e.g. AWS) connected via `user_connections`.

## Root Cause

1. `/api/connected-accounts` returns a dict of accounts keyed by provider name
2. `useConnectedAccounts` hook fetches that and extracts `providerIds` (the keys lowercased)
3. `useChatSendHandlers.ts:81-86` filters for infra providers: `['gcp', 'aws', 'azure', 'ovh', 'scaleway', 'tailscale', 'kubectl', 'onprem']`
4. Line 310: If none of those infra providers are in the response, shows the toast

The `/api/connected-accounts` response likely doesn't include `aws` in its keys even though AWS is connected via `user_connections`. The key naming or casing doesn't match the hardcoded list.

## Expected Behavior

- Toast should not fire if user has a cloud provider connected
- Toast probably shouldn't fire on every message send regardless — it's disruptive for general chat

## Fix

- Verify what keys `/api/connected-accounts` actually returns and align with the infra provider list in `useChatSendHandlers.ts`
- Consider only showing the toast once per session, or only when the user is explicitly trying to run an infra command

Linear: https://linear.app/arvoai/issue/DEV-1285

cc @math-alpha — good first issue, frontend fix in `useChatSendHandlers.ts`

roian6:
## Summary
- Remove the repeated “No cloud provider selected” toast from the general chat send flow.
- Drop the connected-accounts lookup that was only used to decide whether to show that warning.
- Keep normal message sending, provider preferences, and existing error toasts unchanged.

## Context
Issue #518 reports that the provider warning fires on every message send. The issue discussion points toward deleting this notification rather than changing connected-account key matching, so this keeps the fix limited to the chat send warning path.

## Testing
- ✅ `python3` regression check: confirmed the warning string, `anyProviderConnected`, and `useConnectedAccounts` are absent from `client/src/app/chat/components/useChatSendHandlers.ts`
- ✅ `git diff --check`
- ✅ `npm run build`

Attempted but currently blocked by existing baseline/tooling issues:
- ⚠️ `npm run lint` fails with `The requested module 'minimatch' does not provide an export named 'default'`
- ⚠️ `npx tsc --noEmit` fails on existing Next/type errors, including generated `.next/types` route context errors and pre-existing nullability errors in `useChatSendHandlers.ts`

Fixes #518

## Summary by CodeRabbit

* **Refactor**
  * Simplified provider handling in chat messaging. Provider warnings have been consolidated, and logic has been streamlined for improved efficiency.

### DIFF
--- a/client/src/app/chat/components/useChatSendHandlers.ts
+++ b/client/src/app/chat/components/useChatSendHandlers.ts
@@ -1,9 +1,8 @@
 "use client";
 
-import { MutableRefObject, useCallback, useEffect, useMemo, useRef, useState, startTransition } from "react";
+import { MutableRefObject, useCallback, useEffect, useRef, useState, startTransition } from "react";
 import { useToast } from "@/hooks/use-toast";
 import { providerPreferencesService } from "@/lib/services/providerPreferences";
-import { useConnectedAccounts } from "@/hooks/useConnectedAccounts";
 import { Message } from "../types";
 import { AppRouterInstance } from "next/dist/shared/lib/app-router-context.shared-runtime";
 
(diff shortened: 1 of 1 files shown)


beng360:

roian6:
Hi @beng360, happy to sign the CLA.

I checked the repository docs, contribution guide, and Aurora docs, but I couldn't find a public CLA signing link. I also don't think I've received the separate document yet.

Could you resend it or point me to the correct signing link/email flow? Thanks!

roian6:
Hi @beng360 — following up on the CLA step. I'm still happy to sign, but I haven't received the separate document and couldn't find a public signing link. Could you resend the document or share the signing URL/process when convenient? The PR remains ready on my side. Thanks!

PR #540, opened 2026-06-19
fix: Improve Slack action notifications with result summaries and reduced noise

Harrio-6:
Add LLM-generated one-line result summary to action completion notifications
Delete "Action Started" message when action completes to halve channel noise
Add delete_message method to Slack client

## Summary by CodeRabbit

* **New Features**
  * Action notifications can now generate and include an optional one-line result summary in Slack “action completed” updates.
  * “Action started” Slack notifications are now tracked (timestamp/channel) and automatically deleted when the action finishes.
* **Improvements**
  * The Slack notification flow now uses returned send metadata to reliably associate subsequent completion updates with the correct prior message.
  * Completion messages can append the result summary in addition to existing error details when available.

### DIFF
--- a/server/utils/notifications/dispatcher.py
+++ b/server/utils/notifications/dispatcher.py
@@ -201,7 +201,7 @@ def _get_action_data(user_id: str, trigger_metadata: Dict[str, Any], session_id:
     if run_id:
         action_name, started_at = _fetch_action_run_details(run_id, user_id)
 
-    if session_id and status != 'running':
+    if session_id and status == 'success' and not error_message:
         last_bot_text = _fetch_last_bot_message(session_id, user_id)
         if last_bot_text:
             result_summary = _summarize_action_result(action_name, last_bot_text, user_id)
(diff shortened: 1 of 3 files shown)


server/utils/notifications/dispatcher.py
```diff
@@ -179,6 +218,102 @@ def _get_action_data(user_id: str, trigger_metadata: Dict[str, Any], session_id:
...
+        )
+
+        response = tracked_invoke(
+            llm,
+            [HumanMessage(content=prompt)],
+            user_id=user_id,
+            session_id=None,
+            model_name=ModelConfig.INCIDENT_REPORT_SUMMARIZATION_MODEL,
+            request_type="action_result_summary",
+        )
```
damianloch, line 254: If the LLM provider hangs, this blocks indefinitely (until the Celery task times out). A timeout parameter on the model or a wrapper with a deadline would be safer. The existing except Exception will catch eventual failures, but a 30+ second hang is undesirable.

damianloch (changes requested):

damianloch (approved):

PR #541, opened 2026-06-19
Feat: Add new-org onboarding flow with connector selection

Harrio-6:
Add multi-step onboarding wizard for new organizations with connector selection, meeting booking, and guided setup
Persist selections server-side and walk users through connector configuration via a queue-based bottom bar

## Summary by CodeRabbit

* **New Features**
  * Added a full onboarding wizard for connector setup with animated progress, step navigation, review, and post-completion redirect.
  * Introduced onboarding visuals (animated Aurora background, selectable connector tiles, and a setup progress bar).
  * Added onboarding completion support and persistent onboarding tracking.
  * Reworked the sign-in/sign-up experience with improved validation and onboarding routing.
* **Bug Fixes**
  * Improved connector highlighting by honoring the URL `highlight` parameter.
* **Other**
  * Updated routing to redirect `/sign-up` to `/sign-in` (with signup mode) and expanded public access for selected endpoints.


client/src/app/api/auth/register/route.ts
damianloch, line 1: this uses forwardRequest which always calls getAuthenticatedUser() in client/src/lib/auth-helper.ts:22. Registration is unauthenticated so this fails for new user with no sessoin cookie. Failed for me first time then worked second time

server/routes/onboarding_routes.py
```diff
@@ -0,0 +1,65 @@
...
+from utils.auth.rbac_decorators import require_permission
+from utils.db.connection_pool import db_pool
+
+logger = logging.getLogger(__name__)
+
+onboarding_bp = Blueprint('onboarding', __name__)
+
+
+@onboarding_bp.route('/complete', methods=['POST'])
+@require_permission("connectors", "write")
```
damianloch: this is a Casbin race. server/routes/onboarding_routes.py:16 uses @require_permission("connectors", "write"). The Casbin role gets assigned during registration (server/routes/auth_routes.py:137) but the enforcer may not have the new policy loaded yet when the onboarding request hits seconds later.

server/utils/db/db_utils.py
```diff
@@ -2986,6 +2999,20 @@ def initialize_tables():
...
+            # Migration: Add onboarding_completed column to organizations table
+            try:
+                cursor.execute(
+                    "ALTER TABLE organizations ADD COLUMN IF NOT EXISTS onboarding_completed BOOLEAN DEFAULT FALSE;"
+                )
+                conn.commit()
+                logging.info("Ensured onboarding_completed column exists on organizations table.")
+            except Exception as e:
+                logging.exception(
+                    "FATAL: Failed to ensure organizations.onboarding_completed"
```
damianloch: This migration re-raises on failure (raise) and logs as "FATAL", while most other migrations in this file just logging.warning and conn.rollback() without re-raising. This means a failure here (e.g. permissions issue) will prevent the entire app from starting, unlike other column migrations which degrade gracefully.

client/src/app/onboarding/page.tsx
```diff
@@ -0,0 +1,337 @@
...
+        body: JSON.stringify({ selected_connectors: selectedIds }),
+        signal: controller.signal,
+      })
+      if (!res.ok) {
+        throw new Error(`Onboarding completion failed: ${res.status}`)
+      }
+    } catch (e) {
+      console.error("Finish onboarding error:", e)
+      setIsFinishing(false)
+      return
```
damianloch, line 79: When /api/onboarding/complete fails, the catch block logs to console and resets isFinishing but shows no error message to the user. They're just left on the final step with no indication of what went wrong

client/src/app/sign-in/page.tsx
```diff
@@ -33,84 +67,347 @@ function SignInForm() {
...
+
+    if (!/^[\w\s\-.,'&()]+$/u.test(orgName.trim())) {
+      setError("Organization name can only contain letters, numbers, spaces, hyphens, periods, commas, apostrophes, ampersands, and parentheses")
+      return
+    }
+
+    setIsLoading(true)
+
+    try {
+      const response = await fetch('/api/auth/register', {
```
damianloch, line 113: etch('/api/auth/register', ...) hits client/src/app/api/auth/register/route.ts which uses forwardRequest → getAuthenticatedUser() → 401 for unauthenticated users. Every new user attempting to register will get a 401.

damianloch (changes requested):

client/src/components/OnboardingConnectorBar.tsx
damianloch, line 1: there is abiltiy to move forward, but not backwards in onboarding process

client/src/app/onboarding/page.tsx
damianloch, line 1: rn the book a meeting page look like the only option forward, make it a bit more clear to user they can continue forward without this

client/src/app/onboarding/page.tsx
```diff
@@ -0,0 +1,337 @@
...
+              </div>
+            </div>
+          </div>
+
+          {/* Step 3: Alerting and Communication */}
+          <div className="w-full flex-shrink-0 flex items-start justify-center px-6 pt-6 pb-24 overflow-y-auto hide-scrollbar">
+            <div className="w-full max-w-[640px] space-y-6">
+              <div>
+                <h2 className="text-xl font-semibold text-white">
+                  Where do your alerts and team notifications go?
```
damianloch: This page is sort of odd its a mix of a bunch of things. First category is alerting platforms (platforms that deliver webhooks to Aurora). There should be at least 1 alerting platform delivering webhooks to Aurora (probably its own page) otherwise our app literally has no use. I think this should be its own step and it should be pretty clear of the consequences of not setting this part up (app just won't receive alerts to do RCA on). Then there are communication channels. These are just for communication purposes during an alert. These can be on the same page but maybe a subcategory and seperate from alerting platforms because they don't trigger webhooks

damianloch (changes requested):

client/src/app/sign-in/page.tsx
Harrio-6: Fixed, replaced with direct unauthenticated proxy; still sends X-Internal-Secret for inter-service trust

client/src/app/onboarding/page.tsx
Harrio-6: Fixed, Added errorMessage state with visible banner; distinguishes timeout vs generic failure

server/utils/db/db_utils.py
Harrio-6: Fixedm, changed to logging.warning + conn.rollback() matching all other migrations

server/routes/onboarding_routes.py
Harrio-6: fixed, switched to @require_auth_only; route still validates org membership via DB query

client/src/app/api/auth/register/route.ts
Harrio-6: fixed, replaced forwardRequest with a direct proxy that skips session validation entirely. It still sends X-Internal-Secret for inter-service trust, handles timeouts, and safely parses error responses — but no longer requires a session cookie.

client/src/components/OnboardingConnectorBar.tsx
Harrio-6: addressed

client/src/app/onboarding/page.tsx
Harrio-6: addressed

client/src/app/onboarding/page.tsx
Harrio-6: addressed

damianloch (approved):

OlivierTrudeau (approved):

PR #544, opened 2026-06-22
Move Usage & Cost to Settings

Harrio-6:
Moved the Usage & Cost tab from the Monitor page into Settings, visible only to admins
Hidden token counts and token-related UI (charts, columns, streaming stats) when AURORA_ENV is not dev
Exposed AURORA_ENV to the client runtime via docker-entrypoint.sh

## Summary by CodeRabbit

* **New Features**
  * Added an admin-only “Usage & Cost” tab to Settings (with a lazy-loaded Usage panel).
* **Changes**
  * Removed “Usage & Cost” from the Monitor page; usage analytics are now consolidated in Settings.
  * Token usage and enhanced usage/session details are shown only in development.
* **Security**
  * Enforced admin-only access for LLM usage endpoints (models, summary, cost over time).
* **Configuration**
  * Added runtime support for `AURORA_ENV` in generated environment configuration.


isiddharthsingh (dismissed):

isiddharthsingh:
One difference I can see is that the user's cost that you did only shows 30 days but the previous one was having 7 days, 30 days, and 90 days. please bring that here as well

PR #556, opened 2026-06-25
feat: Replace Weaviate knowledge base with unified memory system

OlivierTrudeau:
## Description

Replaces the Weaviate-based Knowledge Base with a unified **Memory** system stored in PostgreSQL. Org knowledge (context, runbooks, infrastructure, learned patterns, postmortems) lives in the `artifacts` table with categories, versioning, and a table-of-contents index injected into the agent system prompt. The agent reads memory on demand via tools instead of vector search. Existing KB data is migrated automatically on upgrade via a one-time Celery task. Weaviate is removed from the Docker Compose dev stack.

## What changed

- Memory stored in `artifacts` table with new `category` and `description` columns; unique index on `(org_id, category, title)`
- Memory categories: `context`, `runbook`, `infrastructure`, `learned`, `postmortem` (legacy rows backfilled as `artifact`)
- New REST API at `/api/memory` — list, create, get, delete entries; file upload for `.md`, `.txt`, `.pdf`
- RBAC: `memory:read` for viewers, `memory:write` for editors
- New Settings → Memory UI replacing Knowledge Base settings
- New agent tools: `list_memories`, `read_memory`, `write_memory`, `append_to_memory`, `edit_memory`, `grep_memories`
- Memory index (`build_memory_index`) injected into system prompt so the agent knows what exists and calls `read_memory()` on demand
- New `memory.md` skill — shared behavioral contract for foreground chat and background RCA
- Documents stored whole (no splitting) — agent paginates large entries via `offset`/`limit` params on `read_memory`
- PDF uploads converted to text via `pypdf` before storage
- Versioning on every write via `services/artifacts/store.py`
- Postmortems migrated from `postmortems` table into `artifacts` with `category='postmortem'`
- One-time Celery migration (`migrate_kb_to_memory`) — auto-triggered on startup if old KB data exists and nothing migrated yet
  - `knowledge_base_memory` → `context` ("Org Context")
  - `infrastructure_context` → `infrastructure` ("Infrastructure Context")
  - `knowledge_base_documents` (S3) → `runbook` entries (PDF/text extracted)
  - `postmortems` → `postmortem` entries (titled from incident alert_title)
- MCP: added `list_memories` and `read_memory`; `search_runbooks` now queries memory runbooks
- Prediscovery prompt updated to write infrastructure via `write_memory` instead of old KB tools
- Sub-agent orchestrator metadata updated for memory tool capability tags
- Removed Weaviate container, env vars, `weaviate_client` dependency, and health check from Docker Compose
- Removed Knowledge Base UI, API routes, and blueprint registration
- Removed old agent tools: `knowledge_base_search`, `rag_indexer`, `infra_context_tool`, `save_discovery_finding`
- Removed incident feedback UI and Weaviate-backed feedback routes
- Removed Aurora Learn user preference
- Removed `knowledge_base.cleanup_stale_documents` Celery periodic task
- MCP deprecated stubs for removed tools (`chat_with_aurora`, `ask_incident`, `regenerate_rca`)

## What to test

- [x] Fresh install: server starts without Weaviate; `artifacts` table has `category`/`description` columns and unique index
- [x] Upgrade migration: existing KB data migrates into Memory on startup; check celery logs for `[Migration] Complete`; restart does not duplicate entries
- [x] Mulitple user knowledge base migrated without collisions or issues
- [x] Settings → Memory tab loads and replaces old Knowledge Base tab
- [x] Create, list, filter by category, and delete memory entries in the UI
- [x] Viewer can read but not write; editor has full write access
- [x] Upload small `.md`, `.txt`, and `.pdf` files — content stored correctly, PDF text extracted with page markers
- [x] Upload large files — stored whole as single entry; no splitting
- [x] Manual create rejects content over 500 KB
- [x] Chat: agent reads memory when asked ("what do you remember about…")
- [x] Chat: agent writes memory when asked to remember something — entry appears in Settings
- [x] Chat: `grep_memories`, `append_to_memory`, and `edit_memory` work as expected
- [x] Chat: agent can delete/forget memory entries on request
- [x] Chat works normally with empty memory (no errors)
- [x] Agent `edit_memory` rejects ambiguous matches (multiple occurrences of old_text)
- [x] Agent `read_memory` paginates large entries correctly via offset/limit
- [x] Trigger RCA — agent uses memory tools during investigation
- [x] Prediscovery (if enabled) creates/updates `infrastructure/Infrastructure Context` entry
- [x] MCP: `list_memories`, `read_memory`, and `search_runbooks` return expected results
- [x] Postmortems visible in Memory UI under "Postmortem" category
- [x] Incidents still work; incident feedback UI is gone (expected)
- [x] `/health` passes without Weaviate
- [x] No Weaviate/knowledge-base errors in server or celery logs during chat/RCA
- [x] Memory entries are org-scoped — Org A data not visible to Org B
- [x] Same title allowed in different categories; same `(category, title)` upserts instead of duplicating


server/routes/memory/routes.py
```diff
@@ -0,0 +1,298 @@
...
+                    source="manual", set_current=True,
+                )
+                created_entries.append({"id": artifact_id, "title": title})
+
+            conn.commit()
+
+        logger.info(
+            f"[Memory] Uploaded {file.filename} as {category}/{base_title} "
+            f"({total_parts} part{'s' if total_parts > 1 else ''}) for org {org_id}"
+        )
```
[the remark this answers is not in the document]
OlivierTrudeau: We're using sanitize to make sure this is fine

server/chat/backend/agent/prompt/composer.py
```diff
@@ -126,10 +124,15 @@ def build_prompt_segments(
...
-    # Build knowledge base memory context for authenticated users
-    knowledge_base_memory = ""
-    if state and hasattr(state, 'user_id'):
-        knowledge_base_memory = build_knowledge_base_memory_segment(state.user_id)
+    # Build memory index for authenticated users
+    memory_context = ""
+    user_id = getattr(state, "user_id", None) if state else None
+    if user_id:
+        try:
+            memory_context = build_memory_index(user_id) or ""
```
damianloch, line 133: this only runs in chats, not RCAs for example. Was this intentional or just missed accidentally. Would need to be supported by some arguments otherwise

client/src/components/MemorySettings.tsx
damianloch, line 1: this doesn't allow for multiple file select which is annoying

damianloch, line 1: Why is it that large documents show on UI as 1/n, 1/n, 3/n? This is ok if it happens in the backend but why should a user be aware of this?

Image

damianloch, line 1: artifact is a category on memory settings page, but my artifacts don't show up in that knowledge base yet so this is a inconsistency

damianloch: same with postmortem

damianloch, line 1: not sure I understand the reason a user can upload a "learned" memory in the memory system. Does a user need to be exposed to possiblity to upload all these types. It looks like frontned is too heavily inspired upon our backend, rather than inspired upon what a user wants most

damianloch, line 1: Image
There is a top right category for doc type, and also middle left. I still don't know which controls what

docker-compose.airtight.yml
damianloch: orphaned

docker-compose.prod-local.yml
damianloch: orphaned

docker-compose.yaml
damianloch: orphaned

damianloch, line 1: helm chart still has tv2 as well. the entire yaml is there and also in values.yaml

server/chat/backend/agent/tools/memory_tool.py
```diff
@@ -0,0 +1,564 @@
...
+            if not org_id:
+                raise ValueError("no_org")
+            yield cursor, conn, org_id
+
+
+def _validate_category(category: str, allow_empty: bool = False) -> str | None:
+    """Validate category, return error JSON string if invalid, None if OK."""
+    if allow_empty and not category:
+        return None
+    if not category or category not in ALL_CATEGORIES:
```
damianloch, line 62: write_memory/read_memory can technically touch artifact rows, though the index/list/grep never surface them). Are we going to eventually move away from artifact specific tools or what is the plan here?

server/services/memory/index_builder.py
```diff
@@ -0,0 +1,108 @@
...
+            safe_desc = _prompt_label(description)
+            desc_suffix = f"  # {safe_desc}" if safe_desc else ""
+            lines.append(f"- {category}/{safe_title}{desc_suffix}")
+
+        result = "\n".join(lines)
+
+        # Enforce line budget — prevent the index from consuming too much prompt space
+        truncated = False
+        result_lines = result.split("\n")
+        if len(result_lines) > MAX_INDEX_LINES:
```
damianloch: does claude code also cap like this? I don't know if this is the right direction to go in if we expect large orgs to be truncating as the norm. In that case we can trim down what is built by index builder purposefuly

server/utils/db/db_utils.py
```diff
@@ -3021,6 +3003,81 @@ def initialize_tables():
...
+                    has_source_data = cursor.fetchone()[0]
+
+                    cursor.execute("""
+                        SELECT EXISTS (
+                            SELECT 1 FROM artifacts
+                            WHERE category IN ('context', 'infrastructure', 'runbook')
+                            LIMIT 1
+                        )
+                    """)
+                    already_migrated = cursor.fetchone()[0]
```
damianloch, line 3150: "Already migrated" guard is coarse. It checks "any row in context/infrastructure/runbook exists." If an org manually created one memory entry before upgrade-triggered migration ran, the guard would treat it as "already migrated" and skip importing their real KB data.

server/utils/db/db_utils.py
```diff
@@ -3021,6 +3003,81 @@ def initialize_tables():
...
+
+                    if has_source_data and not already_migrated:
+                        from services.memory.migration_task import migrate_kb_to_memory
+                        migrate_kb_to_memory.delay()
```
damianloch: .delay() on startup; success/failure only in logs. If migration partially fails (per-row savepoints swallow errors into a stats counter), the user sees a half-populated Memory tab with no indication anything went wrong or is retryable. There's max_retries=0, so a transient S3 blip during doc migration just drops that document permanently.

server/main_chatbot.py
damianloch: not a big deal but stale and there is probs some other stale things

docker-compose.yaml
damianloch: ok i see its actually used by alert correlation, but would be nice to not need to download a 10GB model to do alert correlations. The entire reason we download model locally was because indexing large docs rate limited us but alerts payloads are much less frequent.

server/chat/backend/agent/tools/memory_tool.py
damianloch, line 1: Image

All tools calls to memory tool are just called memory instead of update memory, read memory and all that. Small user experience thing

damianloch (changes requested):

client/src/components/MemorySettings.tsx
OlivierTrudeau: I mean it's more transparent. We could consolidate and reconstruct but adds complexity for not much tbh...

docker-compose.yaml
OlivierTrudeau: will be addressed in separate PR

OlivierTrudeau: Separate PR

server/chat/backend/agent/prompt/composer.py
OlivierTrudeau: I'm not sure why you are saying this. Unless you are referring to the skills loading above? But that's outside of this scopes's PR. I can fix it tho

client/src/components/MemorySettings.tsx
OlivierTrudeau: Now handles multiple uplaods

server/services/memory/index_builder.py
OlivierTrudeau: Yes this is how they do it. I do agree I have as a future upgrade to change the way it handles too large of an index

server/main_chatbot.py
OlivierTrudeau: removed

server/chat/backend/agent/tools/memory_tool.py
OlivierTrudeau: a bit more info added. Will potentially supplement even more in the future

docker-compose.prod-local.yml
OlivierTrudeau: not really. Still needed but removed in a new PR

docker-compose.airtight.yml
OlivierTrudeau: Removed in new PR

server/utils/db/db_utils.py
OlivierTrudeau: No really an issue. It's a one time migration. Someone would have to create a memory between the time this is merged and the migration finishes

server/utils/db/db_utils.py
OlivierTrudeau: added task retry and individual row retry

client/src/components/MemorySettings.tsx
OlivierTrudeau: removed one. Now, you upload the doc and it classifies as context by default but user can change

OlivierTrudeau: Removed artifacts. For postmortem and learn, I'll do a future Pr to migrate postmortems to the artifacts table and for learned, I want the user to be able to modify anything they want and have visibility. We can revisit if the learned category should be removed in general or if we should provide better guidance regarding the categories of memory

OlivierTrudeau: yeah We can revisit how we want to do this. For the moment I chose visibility, but maybe it would be better to never let the user know how it's classified

server/chat/backend/agent/tools/memory_tool.py
OlivierTrudeau: not sure tbh. What dou you think we shoulld do. Let me know and I'll put it on my list to align with that

client/src/components/MemorySettings.tsx
damianloch: Make it 1 file

damianloch: postmortem in scope of this PR to hit

server/services/memory/splitter.py
damianloch, line 1: delete this file it's unused now

server/utils/db/db_utils.py
```diff
@@ -3021,7 +3003,87 @@ def initialize_tables():
...
+                cursor.execute("""
+                    DROP INDEX IF EXISTS idx_artifacts_org_title;
+                    CREATE UNIQUE INDEX IF NOT EXISTS idx_artifacts_org_cat_title
+                        ON artifacts(org_id, category, title);
+                    CREATE INDEX IF NOT EXISTS idx_artifacts_org_category
+                        ON artifacts(org_id, category);
+                """)
+            except Exception as e:
+                logging.error(f"CRITICAL: Failed to migrate artifacts table — memory writes will fail: {e}")
+                conn.rollback()
```
damianloch, line 3038: On migration failure it rolls back (reverting the column adds) and continues to commit()/success, leaving old schema under new code so every memory write 500s

server/utils/db/db_utils.py
```diff
@@ -3021,7 +3003,87 @@ def initialize_tables():
...
+                    ALTER TABLE artifacts ALTER COLUMN category SET NOT NULL;
+                """)
+                # Replace legacy (org_id, title) uniqueness with (org_id, category, title)
+                cursor.execute("""
+                    DROP INDEX IF EXISTS idx_artifacts_org_title;
+                    CREATE UNIQUE INDEX IF NOT EXISTS idx_artifacts_org_cat_title
+                        ON artifacts(org_id, category, title);
+                    CREATE INDEX IF NOT EXISTS idx_artifacts_org_category
+                        ON artifacts(org_id, category);
+                """)
```
damianloch, line 3034: Backfilling all legacy rows to category='artifact' can create (org_id,'artifact',title) duplicates that make CREATE UNIQUE INDEX throw (then hit the CRITICAL rollback). Maybe dedupe beofore creating the index like knowlegde base memory does

client/src/components/MemorySettings.tsx
```diff
@@ -0,0 +1,559 @@
...
+        await fetchEntries();
+      } else {
+        const data = await res.json();
+        throw new Error(data.error || "Failed to create entry");
+      }
+    } catch (error) {
+      toast({
+        title: "Failed to create",
+        description: error instanceof Error ? error.message : "An error occurred",
+        variant: "destructive",
```
damianloch, line 160: Error branches do await res.json() unconditionally; if the backend returns a non-JSON error body (e.g. 502/HTML from the proxy), .json() throws and the catch reports "An error occurred" instead of the actual failure.

This happens a few times in file

server/aurora_mcp/tools_always_on.py
```diff
@@ -75,34 +75,35 @@ async def _do_trigger_rca(
...
+
+
+async def _do_read_memory(api_call: ApiCall, entry_id: str) -> Dict[str, Any]:
+    return truncate_payload(
+        await api_call("GET", f"/api/memory/entries/{entry_id}"),
+        tool_name="read_memory",
     )
 
 
 async def _do_search_runbooks(api_call: ApiCall, query: str, limit: int) -> Dict[str, Any]:
```
damianloch, line 93: Looks like returns all runbooks regardless of what the caller searched for. Doesn't take query or limit into account like previously

server/aurora_mcp/tools_always_on.py
```diff
@@ -296,13 +297,18 @@ async def trigger_rca(
         )
 
     @mcp.tool()
-    async def knowledge_base_search(query: str, limit: int = 5) -> Dict[str, Any]:
-        """Semantic search across Aurora's knowledge base (uploaded docs, indexed runbooks)."""
-        return await _do_knowledge_base_search(api_call, query, limit)
+    async def list_memories(category: str = "") -> Dict[str, Any]:
+        """List memory entries. Optionally filter by category: context, runbook, infrastructure, learned, postmortem, artifact."""
```
damianloch: not artifact, cuz the endpoint validates against MEMORY_CATEGORIES, which excludes artifact

server/chat/backend/agent/skills/loader.py
```diff
@@ -17,6 +17,14 @@
 # Regex to split YAML frontmatter (between --- delimiters) from body
 _FRONTMATTER_RE = re.compile(r"^---\s*\n(.*?)\n---\s*\n(.*)", re.DOTALL)
 
+# Standard skill directories (relative to this file)
+_SKILLS_ROOT = os.path.normpath(os.path.join(os.path.dirname(__file__)))
+
+CORE_SKILLS_DIR = os.path.join(_SKILLS_ROOT, "core")
+BACKGROUND_RCA_DIR = os.path.join(_SKILLS_ROOT, "rca", "background")
+RCA_SEGMENTS_DIR = os.path.join(_SKILLS_ROOT, "rca", "segments")
```
damianloch: segments live in rca/background, this path doesn't exist

server/chat/backend/agent/skills/loader.py
```diff
@@ -17,6 +17,14 @@
...
+INTEGRATIONS_DIR = os.path.join(_SKILLS_ROOT, "integrations")
```
damianloch: INTEGRATIONS_DIR is now defined twice. The new one in loader.py (exported via __init__.py) is never actually used by a consumer, while registry.py:33 defines its own local INTEGRATIONS_DIR and uses that one. So the new loader constant is redundant with an existing definition that everyone actually uses.

server/chat/backend/agent/tools/cloud_tools.py
```diff
@@ -1737,18 +1736,6 @@ def _pinned_trigger(action_id: str = "", _pid=pinned_id, _fn=final_func, **kw):
         description="Analyze a ZIP attachment: list, extract a file, or detect project structure",
     ))
 
-    # Add RAG indexer for ZIPs
-    tools.append(StructuredTool.from_function(
-        func=rag_index_zip,
-        name="rag_index_zip",
```
damianloch, line 1743: this tools was deleted but there are stale references to it in codebase mihght as well completely clean it up

server/chat/background/prediscovery_task.py
```diff
@@ -88,7 +74,7 @@ def build_prediscovery_prompt(user_id: str, providers: List[str], integrations:
 
 ## YOUR DELIVERABLE
 
-At the END of your investigation, you MUST call save_infrastructure_context() with a single
+At the END of your investigation, you MUST call write_memory(category='infrastructure', title='Infrastructure Context', ...) with a single
```
damianloch: so now reusing a title overwrites the previous doc the concern is that this might happen by accident

server/chat/backend/agent/skills/core/memory.md
```diff
@@ -0,0 +1,91 @@
...
+WHEN TO ACCESS MEMORY:
+- When memories seem relevant to the current conversation or investigation.
+- You MUST access memory when the user explicitly asks you to check, recall, or remember.
+- If the user says to ignore or not use memory: proceed as if no memories exist.
+  Do not apply remembered facts, cite, compare against, or mention memory content.
+
+HOW TO VERIFY BEFORE ACTING ON MEMORY:
+- If memory names a service or resource: verify it still exists via your tools
+- If memory names a procedure: confirm the steps are still valid (commands, endpoints)
+- If memory is old: treat as possibly stale and verify key facts
```
damianloch, line 22: i would add a disclaimer here saying to not always trust memories. It could pass all these checks and still be not a valid memory.

damianloch: especially valid is the learned rca root cause memories. This can pollute current RCAs as it gives the LLM an easy root cause out

server/chat/backend/agent/skills/core/memory.md
```diff
@@ -0,0 +1,91 @@
+MEMORY SYSTEM — BEHAVIORAL CONTRACT:
+You have access to a persistent memory system shared across all conversations for this organization.
+Build up this memory over time so that future conversations have complete context about the org's infrastructure, procedures, patterns, and preferences.
+
+WHEN TO WRITE MEMORY:
+- infrastructure: When you discover service topology, deployment chains, monitoring configs, or dependencies during investigation.
+- runbook: When the user shares or you identify step-by-step procedures for known issues.
+- context: When you learn team preferences, escalation paths, on-call structures, org policies, or behavioral instructions (how to respond, what to avoid, communication style, workflows the user wants you to follow, conventions, or any other org/user-specific guidance). 
+- learned: After resolving an incident where the root cause or resolution was non-obvious. Save the pattern so future investigations can match against it.
+- postmortem: When a postmortem is generated or the user shares one.
```
damianloch: shouldn't write_postmortem automatically do this instead of LLM needing to call write memory again after?

server/routes/memory/routes.py
```diff
@@ -0,0 +1,315 @@
...
+        return jsonify({"error": f"category must be one of: {', '.join(MEMORY_CATEGORIES)}"}), 400
+
+    try:
+        raw_bytes = file.read()
+
+        # Extract text based on file type
+        if ext == "pdf":
+            content = _extract_pdf_text(raw_bytes)
+        else:
+            content = raw_bytes.decode("utf-8", errors="replace")
```
damianloch, line 280: _extract_pdf_text and .decode("utf-8", errors="replace") can yield \x00, and Postgres text rejects NUL (ValueError: A string literal cannot contain NUL (0x00)), so the INSERT throws and returns generic 500. I got this when testing on the UI

damianloch: this applies to all other places in file as well

server/routes/memory/routes.py
```diff
@@ -0,0 +1,315 @@
...
+            artifact_id = str(row[0])
+
+            create_version(
+                cursor, artifact_id, org_id, user_id, content,
+                source="manual", set_current=True,
+            )
+            conn.commit()
+
+        logger.info("[Memory] Uploaded file (%d chars)", len(content))
+        return jsonify({"entries": [{"id": artifact_id, "title": base_title}], "parts": 1}), 201
```
damianloch, line 324: services/memory/splitter.py chunks large content in migration_task.py, but the upload/create routes don't import or use it, so big manual uploads bypass the chunking that the rest of the system relies on for the index. i think u said this was supposed to be not used anymore but i think migration still uses splitter.py

server/routes/memory/routes.py
```diff
@@ -0,0 +1,315 @@
...
+    ext = file.filename.rsplit(".", 1)[-1].lower() if "." in file.filename else ""
+    if ext not in ALLOWED_EXTENSIONS:
+        return jsonify({"error": f"File type not allowed. Supported: {', '.join(ALLOWED_EXTENSIONS)}"}), 400
+
+    category = request.form.get("category", "runbook").strip()
+    if category not in MEMORY_CATEGORIES:
+        return jsonify({"error": f"category must be one of: {', '.join(MEMORY_CATEGORIES)}"}), 400
+
+    try:
+        raw_bytes = file.read()
```
damianloch: file.read() loads the entire body into memory; the endpoint's 50 MB limit is checked only on extracted text afterward, so an authenticated user can force a ~1 GB in-memory read and potentially OOM a worker.

server/services/memory/migration_task.py
```diff
@@ -0,0 +1,257 @@
...
+
+                        if ext == "pdf":
+                            reader = PdfReader(io.BytesIO(raw_bytes))
+                            content = "\n\n".join(
+                                f"[Page {i+1}]\n{p.extract_text()}"
+                                for i, p in enumerate(reader.pages)
+                                if p.extract_text()
+                            )
+                        else:
+                            content = raw_bytes.decode("utf-8", errors="replace")
```
damianloch, line 152: same NUL bytes issue as in routes.py https://github.com/Arvo-AI/aurora/pull/556/changes#r3520557951

server/routes/memory/routes.py
```diff
@@ -0,0 +1,315 @@
...
+        with db_pool.get_user_connection() as conn:
+            cursor = conn.cursor()
+            set_rls_context(cursor, conn, user_id, log_prefix="[Memory]")
+
+            cursor.execute(
+                """UPDATE artifacts
+                   SET category = %s, last_edited_by = 'user', updated_at = CURRENT_TIMESTAMP
+                   WHERE id = %s AND org_id = %s AND category = ANY(%s)""",
+                (category, entry_id, org_id, list(MEMORY_CATEGORIES)),
+            )
```
damianloch, line 233: two incidents sharing an alert title collide and the 2nd+ hit DO NOTHING so a real postmortem is dropped with no error. Its pretty possible for them to have the same title cuz its just alert title which is not unique

server/routes/memory/routes.py
```diff
@@ -0,0 +1,315 @@
...
+    """Upload a .md, .txt, or .pdf file as a memory entry."""
+    org_id = get_org_id_from_request()
+
+    if "file" not in request.files:
+        return jsonify({"error": "No file provided"}), 400
+
+    file = request.files["file"]
+    if not file.filename:
+        return jsonify({"error": "No file selected"}), 400
+
```
damianloch, line 257: if stats["errors"] > 0: raise self.retry(...) re-runs the entire migration for any error, but permanently-bad rows (NUL, corrupt PDF, missing S3 object decode) will fail identically every retry, so the task always exhausts retries and ends in a failed state even though all migratable data succeeded on attempt 1

server/chat/backend/agent/tools/memory_tool.py
```diff
@@ -0,0 +1,586 @@
...
+            else:
+                cursor.execute(
+                    """INSERT INTO artifacts
+                           (org_id, user_id, title, content, category,
+                            last_edited_by, updated_at)
+                       VALUES (%s, %s, %s, %s, %s, 'agent', CURRENT_TIMESTAMP)
+                       RETURNING id""",
+                    (org_id, user_id, title.strip(), new_content, category),
+                )
+                artifact_id = str(cursor.fetchone()[0])
```
damianloch, line 388: description TEXT has no DB default, and the create-branch INSERT omits it (append has no description arg at all), so a new entry created by append gets NULL description and shows blank in the index the agent relies on to discover memory.

server/chat/backend/agent/tools/memory_tool.py
```diff
@@ -0,0 +1,586 @@
...
+                   ON CONFLICT (org_id, category, title)
+                   DO UPDATE SET content = EXCLUDED.content,
+                                 description = EXCLUDED.description,
+                                 user_id = EXCLUDED.user_id,
+                                 last_edited_by = 'agent',
+                                 updated_at = CURRENT_TIMESTAMP
+                   RETURNING id""",
+                (org_id, user_id, title.strip(), content,
+                 category, description.strip() if description else None),
+            )
```
damianloch, line 282: Agent-supplied content goes straight into Postgres text; if it ever contains \x00 (pasted tool output/binary), the write throws and returns a generic "Failed to write memory." This happens multiple times in file

damianloch (changes requested):
UI/UX comments for memory: 
1. Be more explicit about the source of memories. I couldn’t tell which ones were auto-discovered or migrated for example. 
2. For auto discovered memories, there isn’t much point in showing them on UI without showing what is in them. If transparency is the goal we should show contents (titles shows nothing). Otherwise maybe hide them
3. I uploaded a file that happened to contain a NUL bye in extracted text. Postgres rejected entirely and gave ValueError. The use toast also showed failure but UI and DB showed it succeeded on the retry (inconsistent state)
4. Memory entries can be of category learned, postmortem, infrastructure… This doesn’t really make sense for a user inputed markdown block to be classified as learned
5. Better name “context” and “learned” memory categories

server/utils/db/db_utils.py
OlivierTrudeau: Non issue because current tables before migration have a uniqueness condition

server/utils/db/db_utils.py
OlivierTrudeau: Added loud failure if fails

client/src/components/MemorySettings.tsx
OlivierTrudeau: ok now catches different scenarios

server/aurora_mcp/tools_always_on.py
OlivierTrudeau: added mixed word matching. Issue is we don't have weaviate anymore to do vector search

server/aurora_mcp/tools_always_on.py
OlivierTrudeau: removed from docstring

server/chat/backend/agent/skills/loader.py
OlivierTrudeau: removed mb

server/chat/backend/agent/skills/loader.py
OlivierTrudeau: yeah removed it

server/chat/backend/agent/tools/cloud_tools.py
OlivierTrudeau: yeah good catch cleaned up

server/chat/background/prediscovery_task.py
OlivierTrudeau: now write has an overwrite field and check

server/services/memory/splitter.py
OlivierTrudeau: done

server/chat/backend/agent/skills/core/memory.md
OlivierTrudeau: Created a new PR to get rid of the 2 unsynced tables for postmortem

server/chat/backend/agent/skills/core/memory.md
OlivierTrudeau: yeah added segment

server/routes/memory/routes.py
OlivierTrudeau: now strips it

OlivierTrudeau: eventually will probably replace this library with one that can process images or something

server/routes/memory/routes.py
OlivierTrudeau: yeah removed the chunking here

server/routes/memory/routes.py
OlivierTrudeau: now adds the incident_id to avoid collision

server/routes/memory/routes.py
OlivierTrudeau: now have an early check in header to make sure we don't accept if too large

server/services/memory/migration_task.py
OlivierTrudeau: yeah fixed with the new helper function

server/routes/memory/routes.py
OlivierTrudeau: yeah now doesn't retry on row error

server/chat/backend/agent/tools/memory_tool.py
OlivierTrudeau: Added a default. Will either make this llm generated in the future or have the dream consolidation take care of this

server/chat/backend/agent/tools/memory_tool.py
OlivierTrudeau: yeah fixed like others

OlivierTrudeau:
## Triaged all 27 unresolved review threads before merge

Went through every unresolved comment against the **current** branch code. Summary: almost all were already addressed by later commits; **one genuine live bug remained**, now fixed in af8cf6a8.

### Fixed now
- **Backfill category mismatch** (`db_utils.py`): the artifacts migration backfilled legacy rows to `category='context'`, but `services/artifacts/store.py:upsert_artifact_by_title` inserts `category='artifact'` and dedupes via `ON CONFLICT (org_id, category, title)`. Legacy artifacts would therefore never match the conflict key → duplicate row on every subsequent write. Backfill + column DEFAULT now use `'artifact'`. (Confirmed this migration has never shipped to `main`/any release tag, so no already-deployed rows need corrective re-migration.)

### Verified already fixed (safe to resolve)
- **NUL bytes (`\x00`)** in memory writes → `strip_nul` now applied in `routes.py` (create/update/upload), `memory_tool.py` (`write_memory` + `append_to_memory`), and `migration_task.py`.
- **Unbounded `file.read()` OOM** → upload now does a bounded `file.read(MAX_RAW_UPLOAD_BYTES+1)` with a Content-Length pre-check (100MB cap).
- **Migration retries on permanently-bad rows** → per-row SAVEPOINTs + `stats["errors"]`; row-level failures no longer trigger `self.retry`.
- **`append_to_memory` NULL description** → falls back to a content snippet instead of NULL.
- **`search_runbooks` ignoring query/limit** → now token-filters memory results and applies `[:limit]`.
- **Postmortem alert-title collision** → migrated title now includes `[incident_id[:8]]`, so distinct incidents no longer collide under `ON CONFLICT DO NOTHING`.
- **RLS blocking the backfill SELECT** → handled via `NO FORCE ROW LEVEL SECURITY` around the backfill.
- **Migration rollback-then-commit** → artifacts block now `raise`s on failure instead of continuing.

### Non-blocking (left as-is; cosmetic / product decisions)
- UI showing large docs as `1/n` chunk labels — product/UX call.
- "Don't always trust memories" disclaimer in `memory.md`.
- CodeQL implicit string-concatenation warnings in `recurrence_agent.py` (style only).
- `MemorySettings.tsx` artifact-category visibility consistency — product call.

CI is green. Ready to merge from my side.

damianloch (approved):

PR #567, opened 2026-06-30
feat(cost): gate infrastructure visualization behind VISUALIZATION_ENABLED (default off)

damianloch:
## Summary
The infrastructure visualization feature fires an LLM extraction roughly every ~30s during every incident RCA, plus one full extraction at RCA completion. This wastes an estimated **$285+/mo** in Opus/Sonnet calls for a feature that is not core to the RCA workflow.

This PR gates both LLM trigger sites behind a new `VISUALIZATION_ENABLED` env flag (default `false`) and hides the UI entry points. **No code is deleted** — the Celery task, SSE route, generator/trigger modules, and React components all remain, so the feature is fully reversible by flipping the flag/constant back on.

## Changes
- `server/chat/backend/agent/tools/cloud_tools.py` — skip the periodic (~30s) `update_visualization.delay(...)` trigger unless `VISUALIZATION_ENABLED=true`. Added `import os` (was not previously imported in this file). Early skip avoids building the `tool_calls_json` payload when disabled.
- `server/chat/background/task.py` — skip the final `update_visualization.apply_async(force_full=True)` call unless `VISUALIZATION_ENABLED=true`, while preserving the surrounding try/except.
- `client/src/app/incidents/components/IncidentCard.tsx` — added a `const VISUALIZATION_ENABLED = false;` constant and guarded both the "Visualization" toggle button and the `<InfrastructureVisualization>` panel render with it. The `showVisualization` state and the `InfrastructureVisualization` import are intentionally kept (still referenced) so re-enabling is a one-line revert and to avoid unused-var issues.
- `docker-compose.yaml`, `docker-compose.prod-local.yml` — added `VISUALIZATION_ENABLED: ${VISUALIZATION_ENABLED:-false}` to the shared `x-common-env` anchor (used by aurora-server and celery_worker).
- `docker-compose.airtight.yml` — added `VISUALIZATION_ENABLED: ${VISUALIZATION_ENABLED:-false}` to the shared `x-common-env` anchor, mirroring the other compose variants.
- `deploy/helm/aurora/values.yaml` — added the chart default `VISUALIZATION_ENABLED: "false"` under `config:`. The helm chart is sourced in this OSS repo, so this is its source of truth.
- `.env.example` — documented `VISUALIZATION_ENABLED=false` near the other feature flags.

> Env-sync is now complete across all compose variants (`docker-compose.yaml`, `docker-compose.prod-local.yml`, `docker-compose.airtight.yml`) and the helm chart (`deploy/helm/aurora/values.yaml`).

## How to re-enable
- **Backend**: set `VISUALIZATION_ENABLED=true` in the environment (`.env` / compose / helm / terraform). Both trigger sites read `os.getenv("VISUALIZATION_ENABLED", "false")`.
- **Frontend**: flip `const VISUALIZATION_ENABLED = false;` to `true` in `IncidentCard.tsx` (one-line change). No other edits needed; the button and panel come back as-is.

## Human follow-up required
- [ ] Confirm prod deploy config (helm/terraform/k8s manifests) leaves `VISUALIZATION_ENABLED` unset or `false`. Since the var is brand new, it defaults to off everywhere, but this was **not** audited in this PR — verify there is no override that sets it true.
- [ ] Decide whether to fully remove the visualization feature later (this PR only disables it; the dead-but-reachable code remains).
- [ ] QA: confirm incident cards render correctly without the Visualization toggle (no layout gaps) across `running`/`summarizing`/`complete` states.
- [ ] The frontend uses a local `const` rather than a `NEXT_PUBLIC_` env var. A `NEXT_PUBLIC_VISUALIZATION_ENABLED` could be wired later if runtime toggling without a rebuild is desired; not done here to keep the change minimal.
- [ ] Frontend lint was **not** run (`client/node_modules` is not installed in this worktree); changes were eyeballed and the IDE linter reported no errors. Run `cd client && npm run lint` in CI/locally to confirm.

## Test plan
1. Backend (disabled, default): trigger an incident RCA and tail celery worker logs (`docker logs -f aurora-celery_worker-1`). Confirm **no** `[Visualization] Triggered 30s update` logs and no `visualization_extraction` LLM calls fire; you should see `[BackgroundChat] Visualization disabled; skipping final visualization` at completion.
2. Backend (re-enabled): set `VISUALIZATION_ENABLED=true`, restart, run an RCA, and confirm the periodic + final visualization triggers fire again.
3. Frontend: load the incidents page and confirm the "Visualization" toggle button and the Infrastructure Analysis panel no longer appear on incident cards.

## Summary by CodeRabbit

* **New Features**
  * Added environment-based feature toggles for incident infrastructure visualization (off by default), including a browser-exposed flag to control the UI.
  * Visualization UI controls and the infrastructure visualization panel are now conditionally available only when the flag is enabled.
* **Bug Fixes**
  * Prevented periodic and final infrastructure visualization extraction from running during RCA unless the backend feature flag is explicitly enabled.


client/src/app/incidents/components/IncidentCard.tsx
```diff
@@ -107,6 +107,9 @@ function isSafeUrl(url: string | undefined): boolean {
   }
 }
 
+// Visualization disabled for cost (re-enable: set true / VISUALIZATION_ENABLED env). See cost-optimization design doc.
+const VISUALIZATION_ENABLED = false;
```
OlivierTrudeau: Why hardcoded here?

OlivierTrudeau (changes requested):

client/src/app/incidents/components/IncidentCard.tsx
damianloch: my bad I changed it to check feature flag like we conventionally do

OlivierTrudeau (approved):

PR #573, opened 2026-07-01
Remove t2v-transformers container

OlivierTrudeau:


deploy/helm/aurora/values.yaml
```diff
@@ -240,6 +239,15 @@ config:
...
 
+  # --- Embedding model for alert correlation (optional) ---
+  # Improves alert-to-incident grouping with semantic similarity.
+  # Requires a DIRECT API key — OpenRouter does NOT support embeddings.
+  # If empty, uses Jaccard token similarity (works fine for most cases).
+  # Format: provider/model-name
+  # Examples: openai/text-embedding-3-small, google/text-embedding-004,
+  #           bedrock/amazon.titan-embed-text-v2:0
+  EMBEDDING_MODEL: ""
+
```
beng360, line 250: have you tested with the Jaccard token similarity. how close is it?

deploy/helm/aurora/templates/t2v-transformers-deployment.yaml
beng360, line 1: should we add a comment somewhere for local dev that says to remove orphans (for dev the transformer will stay orphaned)

beng360 (changes requested):

deploy/helm/aurora/values.yaml
beng360: tested and it should be fine. maybe add in docs that for better correlation for high volumes that have similar titles to add an embedding model

PR #575, opened 2026-07-02
feat: feature paywall with free/enterprise tier enforcement

Harrio-6:
Add plan_tier column to organizations (defaults to free)
Free tier limited to 1 seat (org creator only); enterprise unlimited
Invite button shows upgrade modal with cal.com booking link on free tier
Upgrade modal accessible from both Members tab and Overview settings
PATCH /api/orgs/plan route handles downgrade cleanup (revokes non-creator access, resets Casbin roles, expires invitations)
Paywall bypassed entirely when AURORA_ENV=dev

## Summary by CodeRabbit

* **New Features**
  * Added organization plan awareness across settings, overview, and member management, including free-tier upgrade prompts (hidden in dev).
  * Added an “Upgrade to Enterprise” modal with an enterprise features list and a direct meeting-booking link.
  * Introduced an API to update an org’s plan tier with enforced free↔enterprise rules.

* **Bug Fixes**
  * Expanded invitation cleanup to prevent duplicate conflicts by removing accepted invitations during user re-invite.
  * Enforced free-plan seat limits and improved the hook wrapper’s behavior to avoid unintended error cascades.


server/utils/db/db_utils.py
```diff
@@ -3115,6 +3115,17 @@ def initialize_tables():
                 logging.warning(f"Error de-duplicating organization names: {e}")
                 conn.rollback()
 
+            # Add plan_tier to organizations for org tier paywall
+            try:
+                cursor.execute("SAVEPOINT sp_plan_tier")
+                cursor.execute(
+                    "ALTER TABLE organizations ADD COLUMN IF NOT EXISTS plan_tier VARCHAR(20) DEFAULT 'free';"
+                )
```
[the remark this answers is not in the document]
Harrio-6: The column already has DEFAULT 'free', the update_plan route validates input (if new_tier not in ("free", "enterprise")), and old_tier is coalesced with row[0] or "free". Adding a CHECK constraint + NOT NULL + backfill UPDATE is a schema migration that should be its own PR with proper rollout consideration (particularly for existing rows and deployment ordering). The application-level enforcement is sufficient for now.

server/routes/org_routes.py
```diff
@@ -1156,3 +1160,89 @@ def update_org_preferences(user_id):
...
+                        """UPDATE org_invitations SET status = 'expired'
+                           WHERE org_id = %s AND email IN (
+                               SELECT email FROM users WHERE org_id = %s AND id != %s
+                           )""",
+                        (org_id, org_id, creator_id),
+                    )
+                    cursor.execute(
+                        "DELETE FROM user_tokens WHERE org_id = %s AND user_id != %s",
+                        (org_id, creator_id),
+                    )
```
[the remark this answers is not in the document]
Harrio-6: delete_user_secret works per-user-per-provider (one at a time), and here we're doing a bulk org-wide cleanup of all non-creator tokens across all providers in a single transaction. Calling the helper in a loop would be N queries + N Vault calls inside a transaction that's already doing coordinated cleanup. The user_tokens table stores secret_ref pointers, but on downgrade we're removing the users entirely from the org — their Vault secrets (scoped to provider connections) become orphaned regardless.

client/src/app/org/components/OrgMembers.tsx
```diff
@@ -359,7 +364,11 @@ export default function OrgMembers({ org, currentUserId, isAdmin, onMembersChang
         <p className="text-sm text-muted-foreground">
           {org.members.length} member{org.members.length !== 1 ? "s" : ""}
         </p>
-        {isAdmin && <AddUserDialog onCreated={handleAddUserCreated} />}
+        {isAdmin && (
+          isFree
+            ? <><Button size="sm" variant="outline" className="gap-1.5 text-xs" onClick={() => setUpgradeOpen(true)}><Plus className="h-3.5 w-3.5" />Add</Button><UpgradeModal open={upgradeOpen} onOpenChange={setUpgradeOpen} /></>
+            : <AddUserDialog onCreated={handleAddUserCreated} />
+        )}
```
damianloch, line 373: The isFree swap replaces AddUserDialog with the upgrade-modal button using only planTier, with no AURORA_ENV=dev check, so in dev a free org's Members tab has no way to add a member even though the backend hook and the Overview tab both bypass the paywall in dev.

client/src/app/org/components/UpgradeModal.tsx
```diff
@@ -0,0 +1,49 @@
...
+            <li key={f} className="flex items-center gap-3 text-sm">
+              <Check className="h-4 w-4 flex-shrink-0" />
+              {f}
+            </li>
+          ))}
+        </ul>
+        <div className="border-t pt-4">
+          <Button
+            className="w-full"
+            onClick={() => window.open("https://cal.com/arvo-ai?ref=999998", "_blank", "noopener,noreferrer")}
```
damianloch: this should not be hardcoded in the repo

server/routes/org_routes.py
```diff
@@ -1156,3 +1171,98 @@ def update_org_preferences(user_id):
...
+        (org_id, creator_id),
+    )
+    cursor.execute(
+        "DELETE FROM user_connections WHERE org_id = %s AND user_id != %s",
+        (org_id, creator_id),
+    )
+    cursor.execute(
+        "DELETE FROM user_manual_vms WHERE org_id = %s AND user_id != %s",
+        (org_id, creator_id),
+    )
```
damianloch, line 1201: The downgrade deletes connector/token/VM rows for all non-creator users, but the audit event at L1262 records only old_tier/new_tier, not which users or connections were purged, so an accidental downgrade is unrecoverable and untraceable. We should include removed_user_ids (and counts) in the record_audit_event details payload.

server/utils/hooks.py
```diff
@@ -38,6 +38,15 @@ def _default_after_llm_call(org_id: Optional[str], user_id: str, metadata: dict)
...
     """Called before adding a member to an org. Return (False, message) to block."""
+    if os.getenv("AURORA_ENV", "production") == "dev":
+        return True, None
+    from utils.db.db_utils import db_pool
+    with db_pool.get_admin_connection() as conn:
+        with conn.cursor() as cur:
+            cur.execute("SELECT plan_tier FROM organizations WHERE id = %s", (org_id,))
+            row = cur.fetchone()
+    if row and row[0] == "free" and current_member_count >= 1:
+        return False, "Seat limit reached for your plan. Upgrade to add more members."
```
damianloch, line 51: i think this should fail close

server/utils/hooks.py
```diff
@@ -87,35 +96,48 @@ def _load_hooks():
...
+        logger.exception("Hook '%s' raised — failing closed", name)
+        return False, "Entitlement verification unavailable"
+    logger.exception("Hook '%s' raised — failing open", name)
+    if fn is _HOOK_REGISTRY[name]:
+        return True, None
+    try:
+        return _HOOK_REGISTRY[name](*args, **kwargs)
+    except Exception:
+        logger.exception("Hook '%s' default also raised — failing open", name)
+        return True, None
```
damianloch, line 123: this should also fail close

damianloch (changes requested):

client/src/app/org/components/OrgMembers.tsx
Harrio-6: fixed

client/src/app/org/components/UpgradeModal.tsx
Harrio-6: fixed, added env variable

server/routes/org_routes.py
Harrio-6: added

server/utils/hooks.py
Harrio-6: fixed

server/utils/hooks.py
Harrio-6: fixed

PR #578, opened 2026-07-04
Eliminate postmortem tables — use artifacts as single source of truth

OlivierTrudeau:
## What this does

Right now postmortems live in their own `postmortems` + `postmortem_versions` tables, completely separate from the memory/artifacts system. This means we have two parallel storage systems for what is essentially the same thing — versioned documents tied to incidents.

This PR removes all code paths that read/write the old postmortem tables and makes `artifacts` (with `category='postmortem'`) the single source of truth. The old tables and their data are **not deleted** — they're just orphaned and can be dropped in a follow-up once we're confident nothing breaks.

## What changed

- **DB migration** — adds `incident_id` and `generation_session_id` columns to `artifacts`, backfills existing postmortem data on startup (idempotent), drops the FK on `postmortem_exports` so it can reference artifact UUIDs
- **Agent tool** (`postmortem_tool.py`) — completely rewritten to upsert artifacts with versioning via `create_version`
- **API routes** (`postmortem_routes.py`) — all 10 endpoints rewired to query `artifacts` + `artifact_versions`. Response shapes are identical so the frontend is untouched
- **Generation action** (`postmortem_action.py`) — "reserve" logic now inserts a NULL-content artifact row to signal generation-in-progress
- **Notion export** — reads content from artifacts, removed the legacy `UPDATE postmortems` metadata write
- **Introspection tools** — `list_postmortems` queries artifacts with LEFT JOINs to `postmortem_exports` for export URLs
- **Memory migration task** — removed the postmortem section (migration now happens at startup in db_utils)

## What stays the same

- All API URLs and response JSON shapes (frontend doesn't need changes)
- `postmortem_exports` table still exists and works — just uses artifact IDs now
- The old `postmortems` / `postmortem_versions` tables still exist with all their data
- RLS policies on the old tables remain (harmless, can clean up later)

### Pre-flight
- [ ] Stack starts without errors
- [ ] Server logs show `"Migrated X postmortems into artifacts table"` on startup (or 0 if fresh DB)

### Core postmortem flows
- [ ] View existing postmortem for an incident — content, timestamps, export badges load correctly
- [ ] Edit postmortem content in the UI → save → persists on reload
- [ ] Version history shows correct entries with timestamps
- [ ] Restore a previous version → content reverts
- [ ] Regenerate an existing postmortem → enters generating state → completes with new content

### New postmortem generation
- [ ] Resolve an incident with no postmortem → trigger generation → postmortem appears

### Exports
- [ ] Export to Confluence → succeeds, badge/URL appears
- [ ] Export to Jira → succeeds, badge/URL appears
- [ ] Export to Notion → succeeds, badge/URL appears

### List view
- [ ] Postmortems list page shows all org postmortems with correct incident titles and export status

### Agent tools
- [ ] Ask agent "read the postmortem for incident X" → returns content
- [ ] Postmortem system action (auto-generate on resolve) → creates postmortem without errors

### Edge cases
- [ ] Memory UI does NOT show postmortem entries mixed with regular memory
- [ ] Celery worker logs show no errors referencing `postmortems` table
- [ ] `SELECT count(*) FROM postmortems` — old data still exists (not deleted)

## Risk / rollback

Low risk since the old data is preserved. If something breaks, reverting this branch restores all the old code paths and the data is still there untouched.

### DIFF
--- a/server/routes/postmortem_routes.py
+++ b/server/routes/postmortem_routes.py
@@ -1,4 +1,8 @@
-"""API routes for postmortem CRUD operations and Confluence export."""
+"""API routes for postmortem CRUD operations and exports.
+
+Backed by the `artifacts` / `artifact_versions` tables (category='postmortem').
+Export metadata lives in `postmortem_exports`.
+"""
 
 import logging
 import time
@@ -13,6 +17,7 @@ from connectors.confluence_connector.client import (
     ConfluenceClient,
     markdown_to_confluence_storage,
 )
+from services.artifacts.store import create_version
 from utils.auth.token_management import get_token_data, store_tokens_in_db
 from utils.auth.rbac_decorators import require_permission
 from utils.auth.stateless_auth import get_org_id_from_request, set_rls_context
@@ -20,47 +25,20 @@ from connectors.confluence_connector.auth import refresh_access_token
 from utils.db.connection_pool import db_pool
 from utils.log_sanitizer import sanitize
 from utils.query_helpers import iso_utc
-from utils.validation import is_valid_uuid
+from utils.validation import is_valid_uuid, strip_nul
 
 logger = logging.getLogger(__name__)
 
 postmortem_bp = Blueprint("postmortem", __name__)
 _LOG_PREFIX = "[Postmortem]"
 
-
-def _create_version(
-    cursor, postmortem_id: str, org_id: str, user_id: str, content: str,
-    source: str = "manual", *, set_current: bool = True,
-) -> int:
-    """Insert a new version row for a postmortem atomically and return the version number.
-
-    When set_current=True (default), also advances the current_version_id pointer.
-    Snapshot versions (e.g. pre_edit) should pass set_current=False.
-    """
-    cursor.execute(
-        """INSERT INTO postmortem_versions
-           (postmortem_id, org_id, user_id, content, version_number, source)
-           VALUES (%s, %s, %s, %s,
-                   (SELECT COALESCE(MAX(version_number), 0) + 1
-                    FROM postmortem_versions WHERE postmortem_id = %s),
-                   %s)
-           RETURNING id, version_number""",
-        (postmortem_id, org_id, user_id, content, postmortem_id, source),
-    )
-    row = cursor.fetchone()
-    version_id, version_number = row[0], row[1]
-    if set_current:
-        cursor.execute(
-            "UPDATE postmortems SET current_version_id = %s WHERE id = %s",
-            (str(version_id), postmortem_id),
-        )
-    return version_number
+_CATEGORY = "postmortem"
 
 
 def with_incident_postmortem(require_postmortem=False):
     """Decorator that validates incident_id, resolves org_id, opens DB, sets RLS.
 
-    Injects keyword args: org_id, conn, cursor, postmortem_id (if require_postmortem).
+    Injects keyword args: org_id, conn, cursor, artifact_id (if require_postmortem).
     """
     def decorator(fn):
         @wraps(fn)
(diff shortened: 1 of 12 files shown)


server/utils/db/db_utils.py
```diff
@@ -3031,6 +3031,48 @@ def initialize_tables():
                 conn.rollback()
                 raise
 
+            # artifacts: add incident_id + generation_session_id for postmortem unification
+            try:
```
OlivierTrudeau, line 3035: Is it worth tracking and showing the session that generated the postmortem?

server/routes/postmortem_routes.py
```diff
@@ -137,50 +116,125 @@ def _refresh_confluence_credentials(user_id: str, creds: Dict[str, Any]) -> Opti
...
+        "jiraIssueId": jira.get("external_id"),
+        "jiraIssueKey": jira.get("external_key"),
+        "jiraIssueUrl": jira.get("external_url"),
+        "jiraExportedAt": iso_utc(jira.get("exported_at")),
+        "notionPageId": notion.get("external_id"),
+        "notionPageUrl": notion.get("external_url"),
+        "notionExportedAt": iso_utc(notion.get("exported_at")),
+        "notionDatabaseId": notion.get("external_database_id"),
+        "generationSessionId": artifact_row["generation_session_id"],
+    }
```
beng360, line 175: is this the best way to get these fields?

server/services/actions/postmortem_action.py
```diff
@@ -265,15 +265,29 @@ def _reserve_postmortem_row(user_id: str, incident_id: str, session_id: str) ->
...
-                       VALUES (%s, %s, %s, NULL, %s)
-                       ON CONFLICT (incident_id)
+                    "SELECT alert_title FROM incidents WHERE id = %s",
+                    (incident_id,),
+                )
+                incident_row = cur.fetchone()
+                alert_title = incident_row[0] if incident_row else None
+                short_id = str(incident_id)[:8]
+                title = f"Postmortem: {alert_title} [{short_id}]" if alert_title else f"Postmortem ({incident_id})"
+
```
beng360, line 277: maybe theres a better way o get the title. I notice that its not always accurate

beng360 (changes requested):

server/utils/db/db_utils.py
```diff
@@ -3031,6 +3031,77 @@ def initialize_tables():
...
+                cursor.execute("ALTER TABLE postmortems FORCE ROW LEVEL SECURITY")
+                cursor.execute("ALTER TABLE incidents ENABLE ROW LEVEL SECURITY")
+                cursor.execute("ALTER TABLE incidents FORCE ROW LEVEL SECURITY")
+                conn.commit()
+                if migrated_count > 0:
+                    logging.info(f"Migrated {migrated_count} postmortems into artifacts table.")
+            except Exception as e:
+                logging.warning(f"Error migrating postmortems to artifacts: {e}")
+                conn.rollback()
+                conn.rollback()
```
beng360: double rollback here

server/services/actions/postmortem_action.py
```diff
@@ -265,15 +265,29 @@ def _reserve_postmortem_row(user_id: str, incident_id: str, session_id: str) ->
...
+                            incident_id, generation_session_id, last_edited_by, updated_at)
+                       VALUES (%s, %s, %s, NULL, 'postmortem',
+                               %s, %s, 'agent', CURRENT_TIMESTAMP)
+                       ON CONFLICT (org_id, category, title)
                        DO UPDATE SET content = NULL,
                                      generation_session_id = EXCLUDED.generation_session_id,
+                                     last_edited_by = 'agent',
                                      updated_at = CURRENT_TIMESTAMP""",
-                    (incident_id, user_id, org_id, session_id),
+                    (org_id, user_id, title, incident_id, session_id),
```
beng360: if ever the incident's alert title is updated (during generation). . the upsert is fragile. why did we switch form ON CONFLICT(incident_id) which was more robust

beng360: same thing with save_postmorter

beng360 (changes requested):

server/routes/postmortem_routes.py
OlivierTrudeau: It's one query for all destinations, then a dict lookup per export type. Think it's clean enough — open to suggestions though

server/services/actions/postmortem_action.py
OlivierTrudeau: Good point — title is now just cosmetic. Switched the upsert to ON CONFLICT(incident_id) so it self-heals if alert_title changes

server/utils/db/db_utils.py
OlivierTrudeau: Good catch, removed the dupe

server/services/actions/postmortem_action.py
OlivierTrudeau: You're right — switched to a unique partial index on incident_id and ON CONFLICT(incident_id). Title still updates on save but isn't the conflict key anymore

OlivierTrudeau: Fixed there too — same ON CONFLICT(incident_id) pattern, plus title self-heals on save

PR #582, opened 2026-07-07
feat: optional org plan tier + seat-limit hooks

beng360:
## What

Adds an optional, off-by-default organization **plan tier** to the org model, plus the UI and hook points to support seat-limited plans. Everything here is **inert unless a deployment opts in** — a default OSS deployment sees no behavior change.

## Changes

- `plan_tier` column on `organizations` (defaults to `'free'`), surfaced as `planTier` on `/api/orgs/current`
- Reusable `UpgradeModal` component + `/api/orgs/track` route for lightweight UI events
- Org Members / Overview render an optional upgrade prompt, gated behind `NEXT_PUBLIC_ENABLE_PAYWALL` (unset by default → nothing shown)
- `NEXT_PUBLIC_ENABLE_PAYWALL` + `NEXT_PUBLIC_UPGRADE_BOOKING_URL` env plumbing (`env.ts`, `docker-entrypoint.sh`, `.env.example`, compose files)

## Enforcement

The existing `before_add_member` / `get_seat_limit` hooks remain **no-ops in OSS**. Any actual seat enforcement is left to a deployment that provides its own hook implementation via `AURORA_HOOKS_MODULE` — this PR adds none. Default self-hosted deployments are unaffected: no gating, no prompts, normal add-member flow.

## Testing

- Backend compiles; touched frontend files type-clean
- Flag unset (default): members/overview render exactly as before

## Summary by CodeRabbit

* **New Features**
  * Added optional organization paywall messaging in the app, including an upgrade dialog and booking link support.
  * Organization pages now show plan information and an upgrade prompt when applicable.
  * New environment settings let deployments turn the paywall on or off and configure an upgrade booking URL.

* **Bug Fixes**
  * Organization data now consistently includes plan tier details across the UI and backend.
  * Default plan handling is more reliable when no tier is set.


client/src/app/api/orgs/track/route.ts
```diff
@@ -0,0 +1,20 @@
...
+
+  const response = await fetch(`${API_BASE_URL}/api/orgs/track`, {
+    method: 'POST',
+    headers: { ...authResult.headers, 'Content-Type': 'application/json' },
+    body: JSON.stringify(body),
+  })
+
+  const data = await response.json()
+  return NextResponse.json(data, { status: response.status })
+}
```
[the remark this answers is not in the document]
beng360: Done in 28d99eba — rewrote the route to use the shared `forwardRequest` helper, so it now inherits auth, the 30s AbortSignal timeout, and error normalisation instead of the raw fetch.

client/src/app/api/orgs/track/route.ts
```diff
@@ -0,0 +1,20 @@
...
+  const authResult = await getAuthenticatedUser()
+  if (authResult instanceof NextResponse) return authResult
+
+  const body = await req.json()
+
+  const response = await fetch(`${API_BASE_URL}/api/orgs/track`, {
+    method: 'POST',
+    headers: { ...authResult.headers, 'Content-Type': 'application/json' },
+    body: JSON.stringify(body),
+  })
```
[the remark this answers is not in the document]
beng360: Done in 28d99eba — rewrote the route to use the shared `forwardRequest` helper, so it now inherits auth, the 30s AbortSignal timeout, and error normalisation instead of the raw fetch.

.env.example
```diff
@@ -149,6 +149,14 @@ VISUALIZATION_ENABLED=false
...
 NEXT_PUBLIC_ENABLE_VISUALIZATION=false
 
+# Opt-in org paywall (free/enterprise seat gating). Off by default — plain OSS
+# deployments never paywall. When "true", free-tier orgs see an upgrade prompt
+# instead of the Add-member flow. Actual seat enforcement is separate and only
+# runs when a deployment activates the before_add_member hook (AURORA_HOOKS_MODULE).
+NEXT_PUBLIC_ENABLE_PAYWALL=false
+# Booking/upgrade link shown in the upgrade modal (e.g. a Cal.com URL).
+NEXT_PUBLIC_UPGRADE_BOOKING_URL=
+
```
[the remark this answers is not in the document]
beng360: Done in 28d99eba — added `NEXT_PUBLIC_ENABLE_PAYWALL` and `NEXT_PUBLIC_UPGRADE_BOOKING_URL` to the frontend service env in both `docker-compose.yaml` and `docker-compose.prod-local.yml`.

client/src/lib/env.ts
```diff
@@ -11,6 +11,8 @@ type EnvKey =
   | 'NEXT_PUBLIC_ENABLE_CLOUDBEES'
   | 'NEXT_PUBLIC_ENABLE_BITBUCKET_OAUTH'
   | 'NEXT_PUBLIC_ENABLE_VISUALIZATION'
+  | 'NEXT_PUBLIC_ENABLE_PAYWALL'
+  | 'NEXT_PUBLIC_UPGRADE_BOOKING_URL'
   | 'AURORA_ENV';
```
OlivierTrudeau, line 15: We shouldn't add a booking URL as a param. We should hard code it IMO. It won't change depending on env and for open source it'll be deactivated anyways

server/utils/db/db_utils.py
```diff
@@ -1668,6 +1668,20 @@ def initialize_tables():
...
+            # before_add_member hook). Plain OSS never enforces on this column.
+            try:
+                cursor.execute(
+                    "ALTER TABLE organizations ADD COLUMN IF NOT EXISTS plan_tier VARCHAR(20) DEFAULT 'free';"
+                )
+                conn.commit()
+                logging.info("Ensured plan_tier column exists on organizations table.")
+            except Exception as e:
+                conn.rollback()
+                logging.warning(f"Migration for organizations.plan_tier: {e}")
```
OlivierTrudeau, line 1683: I think this duplicates what's already done in Harry's PR... Did you branch off of that Pr?

client/src/app/api/orgs/track/route.ts
```diff
@@ -0,0 +1,6 @@
+import { NextRequest } from 'next/server'
+import { forwardRequest } from '@/lib/backend-proxy'
+
+export async function POST(request: NextRequest) {
+  return forwardRequest(request, 'POST', '/api/orgs/track', 'Failed to record event')
+}
```
OlivierTrudeau: I believe this points to nothing. Only on Harry's branch it exists

client/src/app/org/components/UpgradeModal.tsx
```diff
@@ -0,0 +1,65 @@
...
+interface UpgradeModalProps {
+  open: boolean;
+  onOpenChange: (open: boolean) => void;
+}
+
+const features = [
+  "Multiple members per organization",
+  "Dedicated engineering support",
+  "Continuous feature development",
+  "10x usage limits",
```
OlivierTrudeau: The 10x usage limit isn't true tho.

beng360: yeah you're right fixed.

client/src/lib/env.ts
beng360: I mean its mostly like we shouldn't hardcode it in OSS

client/src/app/org/components/UpgradeModal.tsx
beng360: Now actually removed in 62c98ff4 (the modal only lists Multiple members / Dedicated support / Continuous development).

client/src/app/api/orgs/track/route.ts
beng360: Good catch — you're right, the backend /api/orgs/track route only exists in the deployment overlay, so this proxy 404'd in OSS. Removed both the proxy and the modal's tracking call in 62c98ff4; tracking is a deployment concern, not core OSS.

client/src/lib/env.ts
beng360: OSS empty, prod uses helm to push it

OlivierTrudeau: I mean there's just no one else that will set it and it adds another env var to manage in helm for no value

OlivierTrudeau (approved):

PR #583, opened 2026-07-07
feat: Memory Injector — async prefetch injects relevant memories into prompt

OlivierTrudeau:
## Summary
- Adds a single-pass LLM selector that picks up to 5 relevant memories from the org memory index (title + description) and injects their full content into the system prompt
- Runs as a non-blocking async prefetch in parallel with prompt building — fires early in `agentic_tool_flow`, result consumed in `composer.py`
- Session-level dedup (Redis set) and 60KB cumulative budget prevent re-surfacing and context bloat

## Key Design Decisions
- **Single LLM call** (not two-pass): speed > precision for an async prefetch that must settle before the turn ends
- **Quality pushed left**: selection quality depends on good `description` fields in memory metadata, not a second filter pass
- **4KB per-file cap**: truncated memories include a note to use `read_memory()` for full content
- **Staleness headers**: memories older than 2 days get a verification warning

## Files
- `server/services/memory/injector.py` (new)
- `server/chat/backend/agent/agent.py` (prefetch start)
- `server/chat/backend/agent/prompt/composer.py` (prefetch consumption)

## Test plan
- [ ] Verify prefetch fires and settles within timeout for interactive chats
- [ ] Verify background RCA sessions skip prefetch (`is_background` guard)
- [ ] Verify empty selection returns empty string (no injection)
- [ ] Verify session budget exhaustion stops further prefetches
- [ ] Verify dedup: same memory not re-surfaced in subsequent turns

### DIFF
--- a/server/chat/backend/agent/agent.py
+++ b/server/chat/backend/agent/agent.py
@@ -13,6 +13,7 @@ from .tools.cloud_tools import set_websocket_context
 from chat.backend.agent.utils.prefix_cache import PrefixCacheManager
 from chat.backend.agent.prompt.prompt_builder import build_prompt_segments, assemble_system_prompt, register_prompt_cache_breakpoints
 from chat.backend.agent.utils.llm_usage_tracker import LLMUsageTracker, LLMUsage
+from services.memory.injector import MemoryPrefetch
 import time
 import asyncio
 from datetime import datetime, timezone
@@ -348,17 +349,17 @@ class Agent:
             _memory_prefetch = None
             if state.user_id and state.session_id:
                 try:
-                    from services.memory.injector import start_memory_prefetch
                     _last_msg_content = ""
                     if state.messages:
                         _lm = state.messages[-1]
                         _last_msg_content = _lm.content if isinstance(getattr(_lm, 'content', ''), str) else str(getattr(_lm, 'content', ''))
                     if _last_msg_content:
-                        _memory_prefetch = start_memory_prefetch(
+                        _memory_prefetch = MemoryPrefetch(
                             user_id=state.user_id,
                             session_id=state.session_id,
                             user_message=_last_msg_content,
                         )
+                        _memory_prefetch.start()
                 except Exception as _mpe:
                     logging.debug("Failed to start memory prefetch: %s", _mpe)
 
--- a/server/chat/backend/agent/agent.py
+++ b/server/chat/backend/agent/agent.py
@@ -344,6 +344,24 @@ class Agent:
                 getattr(state, 'attachments', []),
             )
 
+            # Fire async memory prefetch — runs in parallel with prompt building and tool setup.
+            _memory_prefetch = None
+            if state.user_id and state.session_id:
+                try:
+                    from services.memory.injector import start_memory_prefetch
+                    _last_msg_content = ""
+                    if state.messages:
+                        _lm = state.messages[-1]
+                        _last_msg_content = _lm.content if isinstance(getattr(_lm, 'content', ''), str) else str(getattr(_lm, 'content', ''))
+                    if _last_msg_content:
+                        _memory_prefetch = start_memory_prefetch(
+                            user_id=state.user_id,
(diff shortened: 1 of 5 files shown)


server/chat/backend/agent/agent.py
```diff
@@ -344,6 +345,24 @@ async def agentic_tool_flow(
                 getattr(state, 'attachments', []),
             )
 
+            # Fire async memory prefetch — runs in parallel with prompt building and tool setup.
+            _memory_prefetch = None
+            if state.user_id and state.session_id:
```
beng360, line 350: the PR says background rca sessions skip this but i don't see an is_background guard anywhere? this fires for any session with a user_id + session_id, so background/action turns run the extra selector call too. intentional or should it be gated?

server/chat/backend/agent/agent.py
```diff
@@ -358,6 +377,16 @@ async def agentic_tool_flow(
                 tools = get_cloud_tools()
                 segments = _prompt_future.result()
 
+            # Collect prefetch result and append injected memories to the prompt
+            if _memory_prefetch:
+                try:
+                    injected = _memory_prefetch.get_result(timeout=6.0)
```
beng360, line 383: doesn't this block the turn up to 6s if the selector is slow? it runs parallel with prompt building but we still wait on get_result before assembling the prompt. for a quick interactive chat that's a real latency hit. could we just skip injection when it's not ready instead of blocking?

server/services/memory/injector.py
```diff
@@ -0,0 +1,271 @@
...
+        desc = entry["description"][:150] if entry["description"] else "(no description)"
+        manifest_lines.append(f"- [{entry['category']}] \"{entry['title']}\" | updated: {ts} | {desc}")
+
+    prompt = _SELECTOR_USER.format(
+        user_message=user_message[:2000],
+        manifest="\n".join(manifest_lines),
+    )
+
+    try:
+        llm = create_chat_model(SELECTOR_MODEL, temperature=0.0, streaming=False, max_tokens=256)
```
beng360: SELECTOR_TIMEOUT doesn't actually bound anything right? it's only the default arg for get_result (which agent.py overrides with 6.0) and never gets passed to invoke. so the selector call itself is unbounded — if the provider hangs, the mem-inject worker stays stuck and there's only 4 of them. should we pass a real timeout into create_chat_model?

server/services/memory/queries.py
```diff
@@ -0,0 +1,82 @@
...
+
+    try:
+        with db_pool.get_admin_connection() as conn:
+            with conn.cursor() as cursor:
+                org_id = set_rls_context(cursor, conn, user_id, log_prefix="[MemoryQueries:fetch]")
+                if not org_id:
+                    return []
+
+                results = []
+                for entry in entries:
```
beng360: one select per selected entry — fine at 5 but could be a single WHERE (category, title) IN (...) no?

server/chat/backend/agent/agent.py
OlivierTrudeau: Removed the comment. Every chat should have injected memories

server/chat/backend/agent/agent.py
OlivierTrudeau: The slower it is the higher the timeout needs to be. If I put it too short, it will never inject and the feature is basically dead...

server/services/memory/injector.py
OlivierTrudeau: Good catch — added `timeout=SELECTOR_TIMEOUT` to the `create_chat_model` call so the underlying HTTP request is bounded. If the provider hangs, the call now fails cleanly after 5s instead of blocking a pool worker indefinitely. Fixed in latest commit.

server/services/memory/queries.py
OlivierTrudeau: Yep, replaced the loop with a single `WHERE (category, title) IN (...)` query. One round-trip instead of N. Fixed in latest commit.

PR #584, opened 2026-07-07
feat: Memory Collector + Dream Consolidation

OlivierTrudeau:
## Summary
- **Collector**: Celery task that fires after every turn (interactive + RCA), reads the conversation, and saves durable learnings to org memory via LLM extraction
- **Dream Consolidation**: Nightly Celery Beat task that reviews all org memories — merges duplicates, removes stale entries, rewrites unclear ones

## Collector
- Hooked into `workflow.py` (interactive chats) and `task.py` (background RCA)
- Calls `extract_memories_from_session.delay(session_id, user_id)` directly
- Reads recent messages from `llm_context_history`, asks LLM what to remember
- Writes via existing `write_memory` / `append_to_memory` tool functions (versioned)
- Redis-based lock prevents parallel extractions for the same session

## Dream Consolidation
- `schedule_memory_consolidation` runs on Celery Beat (default: every 24h)
- Finds orgs with 3+ memories, enqueues `run_memory_consolidation` per org
- Loads all memories, asks LLM for a cleanup plan (merge/delete/rewrite)
- Executes plan using versioned memory writes — all changes are auditable
- Redis-based distributed lock + time gate prevents over-consolidation

## Files
- `server/services/memory/collector.py` (new)
- `server/services/memory/consolidation.py` (new)
- `server/chat/backend/agent/workflow.py` (collector hook)
- `server/chat/background/task.py` (collector hook)
- `server/celery_config.py` (task registration + beat schedule)

## Test plan
- [ ] Verify extraction fires after interactive turn completion
- [ ] Verify extraction fires after RCA completion
- [ ] Verify extraction lock prevents parallel runs for same session
- [ ] Verify consolidation skips orgs with < 3 memories
- [ ] Verify consolidation respects 24h time gate between runs
- [ ] Verify merge/delete/rewrite actions execute correctly


server/chat/backend/agent/workflow.py
```diff
@@ -1349,6 +1349,16 @@ async def stream(self, input_state: State):
                 else:
                     logger.warning(f"[WORKFLOW FINAL] Failed to save final context for session {session_id}")
 
+                # Save durable learnings from this conversation to org memory (non-blocking)
+                try:
+                    from services.memory.collector import extract_memories_from_session
+                    extract_memories_from_session.delay(
```
beng360, line 1355: this fires a full 30-turn extraction agent on MAIN_MODEL after every single interactive turn right? that's a heavy agent loop (+ cost) on every message, even for one-off Q&A. have you looked at what this does to per-turn cost/latency at volume? wondering if it should debounce or only run when the turn actually looks like it had org-specific info.

server/services/memory/collector.py
```diff
@@ -0,0 +1,172 @@
...
+    user_id: str,
+    session_id: str,
+    log_prefix: str,
+):
+    """Multi-turn agent loop: browses existing memories, writes new ones."""
+    tools = build_memory_tools(user_id, session_id=session_id)
+
+    llm = create_chat_model(ModelConfig.MAIN_MODEL, temperature=0.0, streaming=False)
+
+    system_prompt = _AGENT_SYSTEM.format(max_turns=MAX_AGENT_TURNS)
```
beng360: _AGENT_SYSTEM.format(max_turns=...) doesn't do anything here — there's no {max_turns} placeholder in the prompt, so MAX_AGENT_TURNS never actually reaches the model. either drop the .format or put the budget in the prompt.

server/services/memory/collector.py
```diff
@@ -0,0 +1,172 @@
...
+Memory Extraction Agent — Background Celery Task
+
+Triggered after each turn (interactive or RCA) to analyze the conversation
+and extract durable learnings into org memory. Modeled after Claude Code's
+extraction agent:
+
+- Fires post-turn (non-blocking, via Celery)
+- Full agentic loop: the LLM can browse, read, write, and append memories
+- Dedup is natural: the agent reads existing content before deciding to act
+- DB-level safety: FOR UPDATE on appends, ON CONFLICT on writes — no app-level lock needed
```
beng360: the PR description says "Redis-based lock prevents parallel extractions for the same session" but i don't see a lock anywhere, and this line even says no app-level lock needed. so if two turns finish close together (or celery retries) two extraction agents run on the same session at once. writes are safe but it's double work/cost. which is it — is the lock supposed to exist or not?

server/chat/backend/agent/workflow.py
OlivierTrudeau: It's ran in the background so doesn't really affect any sort of latency. It will only matter for cost and scaling of our own infra. It's not a 30 turn either, it only uses memory tools to make sure all info it sees is well retained in memory

server/services/memory/collector.py
OlivierTrudeau: No, no lockl. Multiple interactions with memories can happen at the same time so no need to lock the consolidation or the collector task. It's a leftover comment that was forgotten

server/services/memory/collector.py
OlivierTrudeau: Yeah ok, removed it.

PR #592, opened 2026-07-29
feat(actions): HPA/VPA right-sizing audit with Datadog percentiles and Slack review card

isiddharthsingh:
Adds a scheduled action that compares real Datadog usage against the CPU/memory requests, limits, and HPA `maxReplicas` declared in a team's IaC, opens one PR per materially mis-sized workload, and posts a Slack card with **View PR** and **Dismiss**. Aurora never applies anything — the PR is the change and a human merges it.

Seeded `enabled=False`, so nothing runs on a schedule until someone turns it on.

## What's here

- **`metric_stats` Datadog resource type** — Datadog has no time-percentile capability (every documented route is rejected or silently returns zero series, and both target metrics are gauges), so p95 is computed in Python from rolled-up points, behind a `_P95_BY_SOURCE` dispatch seam with an import-time guard. Returns **one compact row per series** rather than a single fat dict: `_truncate_results` keeps a prefix, so today a 973 KB metrics payload becomes `count: 0`, which is indistinguishable from "no data" and would lead the agent to cut an idle-looking workload. Budgeted at 30 KB against `PASS_THROUGH_CHARS` (40 K), not `MAX_OUTPUT_SIZE` (120 K) — above that an LLM summarizer paraphrases percentiles into plausible fiction.
- **`hpa_vpa_recommendations` table** + RLS, with a partial unique index enforcing at most one live proposal per workload.
- **Lifecycle module** owning all table access plus `close_pull_request` as a provider dispatch (GitHub implemented; GitLab/Bitbucket are a table entry each, not a migration — hence the `vcs_provider` column).
- **Card tools** — `send_hpa_vpa_recommendation` (first agent-callable Slack write tool in the repo) and `list_hpa_vpa_recommendations`, so the prompt checks cooldown state *before* opening a PR.
- **Dismiss handler** — closes the PR and starts a 30-day cooldown, then rewrites the card with the buttons removed.
- **Prompt + registration** — generic; it discovers the metrics and VCS providers rather than assuming them.

## Design decisions worth review

- **Dismiss is DB-first, then GitHub.** A GitHub failure leaves the rec dismissed with the PR open — deliberate. We never nag about a workload a human rejected, and a stale open PR is visible and closable by hand. GitHub-first would risk closing the PR then losing the cooldown.
- **A merged PR is acceptance, not rejection** — status `merged`, no cooldown, or the next genuine drift goes unreported.
- **First `requests.patch` in `server/`.** MCP `update_pull_request` is allowlisted in `mcp_tools.py` but absent from `tool_registry.py`, so `gate_action` denies it in background context. Direct REST is the only reliable path.
- **No `tool_registry.py` entry and no `gate_action` call** on the card tool. It writes only to Aurora's own RLS table and the channel Aurora created; a registry key would be a dead switch (`_is_org_tool_permitted` is only consulted inside `gate_action`, which this native tool never calls), and `gate_action` denies unconditionally in scheduled runs. The working kill switch is `enabled=False` on the action.
- **`resource_type='metrics'` behaviour is unchanged.** Its output contract predates this work and the RCA prompts depend on it; only `metric_stats` auto-picks an interval. A test pins that boundary.
- **Asymmetric sizing rules** live in the prompt because they're judgement, not mechanism: memory decreases check observed `max` (not p95) because the failure mode is an OOM-kill; CPU is symmetric because throttling is recoverable.

## Verification

- **341 tests pass** (84 new feature tests; `tests/architectural/` unchanged at 67).
- Triggered the action end-to-end twice against the live stack: clean `success` runs, and the agent independently built `sum:kubernetes.{cpu.usage.total,memory.usage}{*} by {kube_deployment}` — the intended `sum`-across-pods shape — then wrote the living document and correctly issued no recommendations rather than inventing any.
- Card path exercised live: posted, then updated **in place** on the same row (no duplicate); cooldown suppression and the materially-worse breakthrough both confirmed; double-click Dismiss is a no-op with `dismissed_at` unchanged; merge clears the cooldown; a duplicate live claim is blocked by the partial unique index.
- `/slack/interactions`: valid signature → 200, bad signature → 403, stale timestamp → 403, malformed UUID → 200 with a clean message.

Two bugs found in review and fixed with regression tests: a NaN metric point silently broke `list.sort()` and understated `max` (which guards the OOM rule), and Dismiss closed the PR as the clicker rather than the account that opened it, which would have failed for any org-mate without their own GitHub connection.

## Summary by CodeRabbit

* **New Features**
  * Added Datadog `metric_stats` support for percentile-style summaries (p50/p95/p99/max/mean) with interval auto-selection/clamping.
  * Added HPA/VPA right-sizing Slack recommendation cards, including “View PR” and “Dismiss”, plus proposal lifecycle with cooldowns.
  * Introduced a scheduled built-in rightsizing action (disabled by default).
* **Bug Fixes**
  * Strengthened read-only Ask mode by denying the right-sizing recommendation send tool.
  * Improved Datadog result truncation and “no data” messaging consistency.
* **Documentation**
  * Expanded Datadog `metric_stats`/interval guidance and truncation rules.
* **Tests**
  * Added unit tests for `metric_stats` and the HPA/VPA recommendation lifecycle.

### DIFF
--- a/server/chat/backend/agent/tools/cloud_tools.py
+++ b/server/chat/backend/agent/tools/cloud_tools.py
@@ -2213,10 +2213,15 @@ Once you identify which account has the issue, pass account_id (e.g. 'account') 
             func=final_dd_func,
             name="query_datadog",
             description=(
-                "Query Datadog for logs, metrics, monitors, events, traces, hosts, or incidents. "
-                "Set resource_type to 'logs', 'metrics', 'monitors', 'events', 'traces', 'hosts', or 'incidents'. "
+                "Query Datadog for logs, metrics, metric_stats, monitors, events, traces, hosts, or incidents. "
+                "Set resource_type to 'logs', 'metrics', 'metric_stats', 'monitors', 'events', 'traces', "
+                "'hosts', or 'incidents'. Use 'metric_stats' for p50/p95/p99/max per series over long "
+                "windows (capacity and right-sizing questions) -- Datadog cannot compute a time-percentile "
+                "in a query, so 'metrics' plus a percentile rollup does not work. "
                 "Examples: query_datadog(resource_type='logs', query='service:web status:error', time_from='-1h') "
-                "or query_datadog(resource_type='metrics', query='avg:system.cpu.user{*}', time_from='-2h')"
+                "or query_datadog(resource_type='metrics', query='avg:system.cpu.user{*}', time_from='-2h') "
+                "or query_datadog(resource_type='metric_stats', "
+                "query='sum:kubernetes.memory.usage{env:production} by {kube_deployment}', time_from='-30d')"
             ),
             args_schema=QueryDatadogArgs,
         ))
@@ -2470,6 +2475,34 @@ Once you identify which account has the issue, pass account_id (e.g. 'account') 
             logging.info(f"Added {len(specs)} Notion tools for user {user_id} (background={is_background})")
     except Exception as e:
         logging.warning(f"Failed to add Notion tools: {e}")
+
+    # Add HPA/VPA right-sizing card tools -- needs Slack (to post the card) and
+    # GitHub (the PR the card links to). Gated on connectors only: State
+    # .trigger_action_id is never set for background action runs, so there is no
+    # way today to scope these to the right-sizing action specifically.
+    # Excluded from PR review, which is read-only, like every other write tool here.
+    try:
+        # Imported here rather than relying on the Slack block ~1000 lines above:
+        # that name is only bound if this function reached that try block without
+        # raising, so depending on it would fail with UnboundLocalError.
+        from .slack_tool import is_slack_connected as _is_slack_connected
+        from .hpa_vpa_card_tool import HPA_VPA_TOOL_SPECS
+        if (not is_pr_review
+                and _safe_connected(_is_slack_connected, "Slack")
+                and _safe_connected(is_github_connected, "GitHub")):
+            for _func, _name, _schema, _desc in HPA_VPA_TOOL_SPECS:
+                _ctx = with_user_context(_func)
+                _notif = with_completion_notification(_ctx)
+                _final = wrap_func_with_capture(_notif, _name) if tool_capture else _notif
+                tools.append(StructuredTool.from_function(
+                    func=_final,
+                    name=_name,
+                    description=_desc,
+                    args_schema=_schema,
+                ))
+            logging.info(f"Added {len(HPA_VPA_TOOL_SPECS)} HPA/VPA right-sizing tools for user {user_id}")
+    except Exception as e:
+        logging.warning(f"Failed to add HPA/VPA right-sizing tools: {e}")
 
     # Add Jira tools if enabled
     try:
--- a/server/chat/backend/agent/tools/datadog_tool.py
+++ b/server/chat/backend/agent/tools/datadog_tool.py
@@ -374,12 +374,22 @@ def _p95_datadog(client, query: str, time_from: str, time_to: str, limit: int,
         # returned a partial series. A percentile over a silently-clipped window
         # is not the percentile the caller asked for.
         result["window_exceeds_point_cap"] = True
-        result["window_note"] = (
-            f"This window needs ~{expected_points} points at the coarsest supported "
-            f"interval, above Datadog's {_DATADOG_POINT_CAP}-point per-series limit, so the "
-            "series is partial and these percentiles cover only part of the window. "
-            "Use a window of 250 days or less."
+        _partial = (
+            f"This window needs ~{expected_points} points at a {chosen_interval} ms interval, "
+            f"above Datadog's {_DATADOG_POINT_CAP}-point per-series limit, so the series is "
+            "partial and these percentiles cover only part of the window. "
         )
+        if _clamp_interval(interval) is not None:
+            # Caller pinned the interval, so the fix is theirs to make: a coarser
+            # interval (or omitting it to auto-pick) fits the same window.
+            result["window_note"] = _partial + (
+                "Re-run with a coarser interval, or omit interval to auto-pick one."
+            )
+        else:
+            result["window_note"] = _partial + (
+                "This is already the coarsest supported interval, so use a window of "
+                "250 days or less."
+            )
 
     if truncated:
         result["series_truncated"] = True
@@ -405,11 +415,14 @@ _METRIC_STATS_SOURCES = ("datadog",)
 _P95_BY_SOURCE = {"datadog": _p95_datadog}
 
 # Fail fast at import if a source is advertised without a dispatch entry --
-# mirrors the _ALERTS_PATH_BY_SOURCE guard in aurora_mcp/tools_gated.py.
-assert set(_P95_BY_SOURCE) >= set(_METRIC_STATS_SOURCES), (
-    "metric_stats sources not fully covered by _P95_BY_SOURCE: "
-    f"missing {set(_METRIC_STATS_SOURCES) - set(_P95_BY_SOURCE)}"
-)
+# mirrors the _ALERTS_PATH_BY_SOURCE guard in aurora_mcp/tools_gated.py, but
+# raises instead of asserting: `python -O` strips asserts and would silently
+# drop the guard.
+_uncovered_sources = set(_METRIC_STATS_SOURCES) - set(_P95_BY_SOURCE)
+if _uncovered_sources:
+    raise RuntimeError(
+        f"metric_stats sources not fully covered by _P95_BY_SOURCE: missing {_uncovered_sources}"
+    )
 
 
 def _query_metric_stats(client, query: str, time_from: str, time_to: str, limit: int, **kwargs) -> dict:
@@ -427,6 +440,11 @@ def _query_metric_stats(client, query: str, time_from: str, time_to: str, limit:
             "query is required for resource_type='metric_stats' "
             "(e.g., 'sum:kubernetes.memory.usage{env:production} by {kube_deployment}')"
         )
+    # `source` is deliberately NOT part of QueryDatadogArgs: this tool is the
+    # Datadog surface, so the agent never selects a backend. It is an internal
+    # seam for a future caller that multiplexes providers (mirroring
+    # _resolve_source in aurora_mcp/tools_gated.py), reachable because
+    # query_datadog forwards **kwargs.
     source = (kwargs.get("source") or "datadog").lower().strip()
     backend = _P95_BY_SOURCE.get(source)
     if backend is None:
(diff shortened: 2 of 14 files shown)


isiddharthsingh:
Worked through the SonarQube and CodeRabbit findings. Fixed the valid ones; below are the ones I'm deliberately leaving, with reasons.

## Fixed in `2e2679cf`

- **S8572** — one `logger.error(..., exc_info=True)` in `_send_ephemeral` I missed last round; now `logger.exception`.
- **Lazy logging** — all four `logger.exception` calls converted to `%s` args and the redundant caught-exception object dropped (`exception()` already logs the traceback).
- **`window_note` at the interval ceiling** — real gap in my previous fix. It only checked *whether* the caller pinned an interval, not whether that interval was already `_MAX_INTERVAL_MS`, so pinning the ceiling on an oversized window advised "re-run with a coarser interval" when none exists. Now gated on `chosen_interval < _MAX_INTERVAL_MS`, with a regression test covering all four paths.
- **Unused lambda variable** in the sentinel-closer tests.
- **Two new S3776 (cognitive complexity)** — both were regressions my *own* previous fixes introduced, so they're in scope: `compute_severity_score` (16) and `_p95_datadog` (16). Extracted `_is_real_number`, `_dimension_score`, and `_point_cap_note`; the first also removes duplication between `compute_severity_score` and `is_materially_worse`. Verified behaviour is byte-identical on every input class before and after.

## Not fixing — 4 pre-existing S3776 + S107

`_handle_hpa_vpa_dismiss` (30), `send_hpa_vpa_recommendation` (27), `_validate` (22), `query_datadog` (20), and `claim_recommendation`'s 16 params.

These are **maintainability-only**, and the quality gate already passes maintainability at rating 1 — they are not what's failing it. Each is a linear sequence of guard clauses or an ordered transaction (authenticate → parse → authorize → transition → close → rewrite), where the step count *is* the cognitive complexity. Splitting them would mean threading a shared connection, cursor, and RLS context through new helper boundaries, which trades a lint metric for a genuinely higher chance of a correctness bug in the exact code paths that must not break: the atomic dismiss transition and the claim-row-before-post ordering.

`claim_recommendation`'s 16 keyword-only params map 1:1 to NOT NULL-ish columns; collapsing them into a dict would move column validation from the call site to runtime.

Happy to do these as a follow-up PR where they can be reviewed as a pure refactor with no behavioural changes mixed in.

## Not fixing — 2 in `SKILL.md`

- **MD022 blank lines after headings** — cosmetic, and not the convention in this file. `### Tool Usage`, `### Datadog Query Syntax`, and `### Examples` all predate this PR and put content on the next line. markdownlint isn't run on this repo, so matching the surrounding style is the better call.
- **"Auto-pick claim is wrong for `'metrics'`"** — already correct. Fixed in an earlier round; the section reads *"For `'metrics'`, omitting it leaves Datadog's own default resolution in place."* The comment was written against the pre-fix text.

## Acknowledged, mitigated, not fully fixed — Slack 3s ack budget

CodeRabbit is right that a 5s timeout is a mitigation rather than a fix, and that the proper answer is moving the PR close and card rewrite onto Celery.

I'm leaving it as a short timeout for this PR because the ordering is already designed so a timeout is safe: the dismissal transition **commits before** the GitHub call, so a slow provider costs only the automatic PR close — surfaced on the card as "could not be closed automatically; please close it on GitHub" — and never the 30-day cooldown. That's the failure direction we want; the alternative (closing the PR then losing the cooldown) is exactly the nagging this feature exists to prevent.

Deferring to Celery means a new task, its own RLS context, and a retry/idempotency story for the card rewrite. That's a meaningful chunk of new surface area, and I'd rather it not ride along in a PR this size. Tracked as the follow-up; the constraint and the reasoning are both in a comment at `_GITHUB_CLOSE_TIMEOUT`.

## Validation

1133 tests pass in the exact CI environment (`python:3.12-slim` + the 8 pinned packages from `linters.yml`); 94 feature tests; architectural 67. Live card path re-verified after the refactors: post → update-in-place on the same row → GitLab rejected up front.

server/chat/backend/agent/tools/cloud_tools.py
damianloch, line 1: Looks like chats can now post a slack card. The posting of the slack card should be exclusive to the seeded templated action being ran

server/chat/backend/agent/tools/datadog_tool.py
```diff
@@ -147,7 +165,306 @@ def _query_metrics(client, query: str, time_from: str, time_to: str, limit: int)
...
+    values = attrs.get("values") or []
+
+    max_series = min(max(limit, 1), _MAX_STATS_SERIES)
+    rows, used_bytes, truncated = [], 0, False
+    for idx, entry in enumerate(series[:max_series]):
+        meta = entry if isinstance(entry, dict) else {}
+        row = _summarize_series(
+            meta.get("group_tags"),
+            meta.get("unit"),
+            values[idx] if idx < len(values) else [],
```
damianloch, line 379: When values is shorter than series, the substituted [] produces points: 0, note: "no non-null points in window", indistinguishable from a genuinely idle workload, which is the exact failure mode this resource type exists to prevent. Fix: Emit a distinct marker (e.g. note: "series metadata present but no values row returned") so the agent can tell a malformed response from real absence.

server/utils/db/db_utils.py
```diff
@@ -1409,6 +1409,63 @@ def initialize_tables():
...
+                        recommendation JSONB NOT NULL DEFAULT '{}'::jsonb,
+                        severity_score NUMERIC,
+                        slack_channel_id VARCHAR(64),
+                        slack_message_ts VARCHAR(64),
+                        action_run_id UUID,
+                        dismissed_by VARCHAR(255),
+                        dismissed_at TIMESTAMP,
+                        cooldown_until TIMESTAMP,
+                        created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
+                        updated_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP
```
damianloch, line 1448: `cooldown_until` is written as aware UTC but compared with NOW(), the 30-day window drifts on a non-UTC session. TIMESTAMPTZ instead?

server/routes/slack/slack_events.py
```diff
@@ -556,4 +589,207 @@ def _handle_suggestion_details(payload: dict, action: dict, slack_user_id: str,
...
+                dismissed = dismiss_recommendation(cursor, rec_id, clicker_org_id, slack_user_id)
+                # Commit before the GitHub call so a slow API cannot hold the row lock.
+                conn.commit()
+
+        if not dismissed:
+            logger.info(f"Recommendation {rec_id} was already dismissed; no-op")
+            if client:
+                _send_ephemeral(client, channel_id, slack_user_id,
+                                "That recommendation was already dismissed.")
+            return jsonify({"text": ""}), 200
```
damianloch, line 703: The DB transition and PR close are unconditional, but every piece of feedback — ephemerals and _rewrite_dismissed_card — is gated on if client: / an early return when get_slack_client_for_user yields None, so the workload gets a 30-day cooldown and the PR gets closed while the card keeps its live Dismiss button and no message appears.

server/routes/slack/slack_events.py
```diff
@@ -556,4 +589,207 @@ def _handle_suggestion_details(payload: dict, action: dict, slack_user_id: str,
...
+                           "No cooldown applied.")
+        elif close_result.get("success"):
+            status_line = (f"*Status:* Dismissed by <@{slack_user_id}> -- PR #{pr_number} closed. "
+                           f"Aurora will not raise this workload again for {HPA_VPA_COOLDOWN_DAYS} days.")
+        else:
+            logger.error(
+                f"Failed to close PR #{pr_number} for recommendation {rec_id}: {close_result.get('error')}"
+            )
+            status_line = (f"*Status:* Dismissed by <@{slack_user_id}> -- PR #{pr_number} could not be "
+                           "closed automatically; please close it on GitHub.")
```
damianloch, line 742: status_line says "could not be closed automatically; please close it on GitHub", but the rewrite passes with_actions=False, which drops the entire actions block including the View PR button — so the one message asking for manual work deletes the only link to the PR.

server/routes/slack/slack_events.py
damianloch, line 1: Invalid UUID, unresolvable org, and unauthenticated-without-workspace-token all return an empty 200 with no ephemeral, so the click is indistinguishable from a broken button

server/routes/slack/slack_events.py
```diff
@@ -556,4 +589,207 @@ def _handle_suggestion_details(payload: dict, action: dict, slack_user_id: str,
...
+                if not row or row[0] != clicker_org_id:
+                    logger.warning(
+                        f"Slack user {sanitize(slack_user_id)} tried to dismiss recommendation outside their org"
+                    )
+                    if client:
+                        _send_ephemeral(client, channel_id, slack_user_id,
+                                        "Unauthorized: that recommendation does not belong to your organization.")
+                    return jsonify({"text": ""}), 200
+
+                dismissed = dismiss_recommendation(cursor, rec_id, clicker_org_id, slack_user_id)
```
damianloch, line 694: maybe there should be a ui confirmation for users if they want to mute it for 30 days or not

isiddharthsingh:
Superseded by #593 — the branch was renamed and GitHub auto-closed this PR. Same commits, same head SHA (`693be8c6`). Review history here is preserved for reference; please continue on #593.

server/services/actions/hpa_vpa_recommendations.py
```diff
@@ -0,0 +1,488 @@
...
+
+
+def mark_superseded(cursor, rec_id: str) -> None:
+    """Retire a row whose mis-size worsened, or that a newer rec took over."""
+    cursor.execute(
+        """UPDATE hpa_vpa_recommendations
+              SET status = %s, cooldown_until = NULL, updated_at = NOW()
+            WHERE id = %s::uuid""",
+        (STATUS_SUPERSEDED, rec_id),
+    )
```
damianloch, line 315: mark_superseded is committed before the Slack post, but the post's compensation path only deletes the new claim, the superseded dismissal is never restored, so the remaining anti-nag window is gone for good.

server/services/actions/hpa_vpa_recommendations.py
```diff
@@ -0,0 +1,488 @@
...
+            body = resp.json()
+        except ValueError:
+            body = {}
+        # A merged PR means the human already accepted. GitHub reports the
+        # merge rather than erroring, and treating it as a rejection would
+        # start an anti-nag window on an *approved* change.
+        if body.get("merged"):
+            return {"success": True, "already_merged": True, "state": body.get("state")}
+        return {"success": True, "state": body.get("state", "closed")}
+    if resp.status_code == 404:
```
damianloch, line 423: i think this should be treated as a error or reqiures more code to tell if its realyl a success. Maybe a use toast saying not reachable

server/chat/backend/agent/tools/hpa_vpa_card_tool.py
damianloch, line 1: severity_score is an LLM-supplied argument. Someone clicks Dismiss and gets 30 days of quiet but the AI is allowed to break that silence early if the problem got 25% worse, but the AI is the one reporting how bad it is so this sort of doesnt make sense

PR #593, opened 2026-07-30
feat(actions): HPA/VPA right-sizing audit with Datadog percentiles and Slack review card

isiddharthsingh:
Adds a scheduled action that compares real Datadog usage against the CPU/memory requests, limits, and HPA `maxReplicas` declared in a team's IaC, opens one PR per materially mis-sized workload, and posts a Slack card with **View PR** and **Dismiss**. Aurora never applies anything — the PR is the change and a human merges it.

Seeded `enabled=False`, so nothing runs on a schedule until someone turns it on.

## What's here

- **`metric_stats` Datadog resource type** — Datadog has no time-percentile capability (every documented route is rejected or silently returns zero series, and both target metrics are gauges), so p95 is computed in Python from rolled-up points, behind a `_P95_BY_SOURCE` dispatch seam with an import-time guard. Returns **one compact row per series** rather than a single fat dict: `_truncate_results` keeps a prefix, so today a 973 KB metrics payload becomes `count: 0`, which is indistinguishable from "no data" and would lead the agent to cut an idle-looking workload. Budgeted at 30 KB against `PASS_THROUGH_CHARS` (40 K), not `MAX_OUTPUT_SIZE` (120 K) — above that an LLM summarizer paraphrases percentiles into plausible fiction.
- **`hpa_vpa_recommendations` table** + RLS, with a partial unique index enforcing at most one live proposal per workload.
- **Lifecycle module** owning all table access plus `close_pull_request` as a provider dispatch (GitHub implemented; GitLab/Bitbucket are a table entry each, not a migration — hence the `vcs_provider` column).
- **Card tools** — `send_hpa_vpa_recommendation` (first agent-callable Slack write tool in the repo) and `list_hpa_vpa_recommendations`, so the prompt checks cooldown state *before* opening a PR.
- **Dismiss handler** — closes the PR and starts a 30-day cooldown, then rewrites the card with the buttons removed.
- **Prompt + registration** — generic; it discovers the metrics and VCS providers rather than assuming them.

## Design decisions worth review

- **Dismiss is DB-first, then GitHub.** A GitHub failure leaves the rec dismissed with the PR open — deliberate. We never nag about a workload a human rejected, and a stale open PR is visible and closable by hand. GitHub-first would risk closing the PR then losing the cooldown.
- **A merged PR is acceptance, not rejection** — status `merged`, no cooldown, or the next genuine drift goes unreported.
- **First `requests.patch` in `server/`.** MCP `update_pull_request` is allowlisted in `mcp_tools.py` but absent from `tool_registry.py`, so `gate_action` denies it in background context. Direct REST is the only reliable path.
- **No `tool_registry.py` entry and no `gate_action` call** on the card tool. It writes only to Aurora's own RLS table and the channel Aurora created; a registry key would be a dead switch (`_is_org_tool_permitted` is only consulted inside `gate_action`, which this native tool never calls), and `gate_action` denies unconditionally in scheduled runs. The working kill switch is `enabled=False` on the action.
- **`resource_type='metrics'` behaviour is unchanged.** Its output contract predates this work and the RCA prompts depend on it; only `metric_stats` auto-picks an interval. A test pins that boundary.
- **Asymmetric sizing rules** live in the prompt because they're judgement, not mechanism: memory decreases check observed `max` (not p95) because the failure mode is an OOM-kill; CPU is symmetric because throttling is recoverable.

## Verification

- **341 tests pass** (84 new feature tests; `tests/architectural/` unchanged at 67).
- Triggered the action end-to-end twice against the live stack: clean `success` runs, and the agent independently built `sum:kubernetes.{cpu.usage.total,memory.usage}{*} by {kube_deployment}` — the intended `sum`-across-pods shape — then wrote the living document and correctly issued no recommendations rather than inventing any.
- Card path exercised live: posted, then updated **in place** on the same row (no duplicate); cooldown suppression and the materially-worse breakthrough both confirmed; double-click Dismiss is a no-op with `dismissed_at` unchanged; merge clears the cooldown; a duplicate live claim is blocked by the partial unique index.
- `/slack/interactions`: valid signature → 200, bad signature → 403, stale timestamp → 403, malformed UUID → 200 with a clean message.

Two bugs found in review and fixed with regression tests: a NaN metric point silently broke `list.sort()` and understated `max` (which guards the OOM rule), and Dismiss closed the PR as the clicker rather than the account that opened it, which would have failed for any org-mate without their own GitHub connection.

## Summary by CodeRabbit

- **New Features**
  - Added HPA/VPA right-sizing audits with pull-request recommendations.
  - Added Slack recommendation cards with View PR and Dismiss actions, cooldown handling, and status updates.
  - Added Datadog metric statistics, including percentile summaries and configurable query intervals.
  - Added recommendation tracking to prevent duplicate or excessive notifications.

- **Bug Fixes**
  - Read-only Ask mode now blocks recommendation-sending actions.
  - Improved handling and messaging for truncated or unusable Datadog results.

- **Documentation**
  - Expanded Datadog guidance, supported resource types, interval behavior, and metric interpretation.

---

Replaces #592, which GitHub auto-closed when the branch was renamed. Same commits, same head SHA (`693be8c6`); no code changed.

### DIFF
--- a/server/chat/backend/agent/tools/cloud_tools.py
+++ b/server/chat/backend/agent/tools/cloud_tools.py
@@ -2213,10 +2213,15 @@ Once you identify which account has the issue, pass account_id (e.g. 'account') 
             func=final_dd_func,
             name="query_datadog",
             description=(
-                "Query Datadog for logs, metrics, monitors, events, traces, hosts, or incidents. "
-                "Set resource_type to 'logs', 'metrics', 'monitors', 'events', 'traces', 'hosts', or 'incidents'. "
+                "Query Datadog for logs, metrics, metric_stats, monitors, events, traces, hosts, or incidents. "
+                "Set resource_type to 'logs', 'metrics', 'metric_stats', 'monitors', 'events', 'traces', "
+                "'hosts', or 'incidents'. Use 'metric_stats' for p50/p95/p99/max per series over long "
+                "windows (capacity and right-sizing questions) -- Datadog cannot compute a time-percentile "
+                "in a query, so 'metrics' plus a percentile rollup does not work. "
                 "Examples: query_datadog(resource_type='logs', query='service:web status:error', time_from='-1h') "
-                "or query_datadog(resource_type='metrics', query='avg:system.cpu.user{*}', time_from='-2h')"
+                "or query_datadog(resource_type='metrics', query='avg:system.cpu.user{*}', time_from='-2h') "
+                "or query_datadog(resource_type='metric_stats', "
+                "query='sum:kubernetes.memory.usage{env:production} by {kube_deployment}', time_from='-30d')"
             ),
             args_schema=QueryDatadogArgs,
         ))
@@ -2470,6 +2475,34 @@ Once you identify which account has the issue, pass account_id (e.g. 'account') 
             logging.info(f"Added {len(specs)} Notion tools for user {user_id} (background={is_background})")
     except Exception as e:
         logging.warning(f"Failed to add Notion tools: {e}")
+
+    # Add HPA/VPA right-sizing card tools -- needs Slack (to post the card) and
+    # GitHub (the PR the card links to). Gated on connectors only: State
+    # .trigger_action_id is never set for background action runs, so there is no
+    # way today to scope these to the right-sizing action specifically.
+    # Excluded from PR review, which is read-only, like every other write tool here.
+    try:
+        # Imported here rather than relying on the Slack block ~1000 lines above:
+        # that name is only bound if this function reached that try block without
+        # raising, so depending on it would fail with UnboundLocalError.
+        from .slack_tool import is_slack_connected as _is_slack_connected
+        from .hpa_vpa_card_tool import HPA_VPA_TOOL_SPECS
+        if (not is_pr_review
+                and _safe_connected(_is_slack_connected, "Slack")
+                and _safe_connected(is_github_connected, "GitHub")):
+            for _func, _name, _schema, _desc in HPA_VPA_TOOL_SPECS:
+                _ctx = with_user_context(_func)
+                _notif = with_completion_notification(_ctx)
+                _final = wrap_func_with_capture(_notif, _name) if tool_capture else _notif
+                tools.append(StructuredTool.from_function(
+                    func=_final,
+                    name=_name,
+                    description=_desc,
+                    args_schema=_schema,
+                ))
+            logging.info(f"Added {len(HPA_VPA_TOOL_SPECS)} HPA/VPA right-sizing tools for user {user_id}")
+    except Exception as e:
+        logging.warning(f"Failed to add HPA/VPA right-sizing tools: {e}")
 
     # Add Jira tools if enabled
     try:
--- a/server/chat/backend/agent/tools/cloud_tools.py
+++ b/server/chat/backend/agent/tools/cloud_tools.py
@@ -1035,12 +1035,16 @@ def get_cloud_tools():
     rca_flag = getattr(state_context, 'trigger_rca_requested', False) if state_context else False
     is_background = getattr(state_context, 'is_background', False) if state_context else False
     is_postmortem_action = getattr(state_context, 'is_postmortem_action', False) if state_context else False
+    is_hpa_vpa_action = getattr(state_context, 'is_hpa_vpa_action', False) if state_context else False
     is_pr_review = getattr(state_context, 'is_pr_review', False) if state_context else False
     is_rca_context = _is_background_rca(state_context, is_background)
     _action_id = getattr(state_context, 'trigger_action_id', None) if state_context else None
     _incident_id = getattr(state_context, 'incident_id', None) if state_context else None
     capture_tag = "capture" if tool_capture else "nocapture"
-    cache_key = f"{user_id}:{capture_tag}:{mode_suffix}:background={is_background}:rca={rca_flag}:postmortem={is_postmortem_action}:is_rca_ctx={is_rca_context}:pr_review={is_pr_review}:action_id={_action_id}:incident={_incident_id}"
+    # hpa_vpa is part of the key for the same reason postmortem is: it changes
+    # which tools are returned, so sharing a cache entry across contexts would
+    # leak the card tool into an ordinary chat (or withhold it from the action).
+    cache_key = f"{user_id}:{capture_tag}:{mode_suffix}:background={is_background}:rca={rca_flag}:postmortem={is_postmortem_action}:hpa_vpa={is_hpa_vpa_action}:is_rca_ctx={is_rca_context}:pr_review={is_pr_review}:action_id={_action_id}:incident={_incident_id}"
     
     current_time = time.time()
     if (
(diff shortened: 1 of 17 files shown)


server/chat/backend/agent/tools/cloud_tools.py
damianloch, line 1: Looks like chats can now post a slack card. The posting of the slack card should be exclusive to the seeded templated action being ran

server/chat/backend/agent/tools/datadog_tool.py
```diff
@@ -147,7 +165,306 @@ def _query_metrics(client, query: str, time_from: str, time_to: str, limit: int)
...
+    values = attrs.get("values") or []
+
+    max_series = min(max(limit, 1), _MAX_STATS_SERIES)
+    rows, used_bytes, truncated = [], 0, False
+    for idx, entry in enumerate(series[:max_series]):
+        meta = entry if isinstance(entry, dict) else {}
+        row = _summarize_series(
+            meta.get("group_tags"),
+            meta.get("unit"),
+            values[idx] if idx < len(values) else [],
```
damianloch: When values is shorter than series, the substituted [] produces points: 0, note: "no non-null points in window", indistinguishable from a genuinely idle workload, which is the exact failure mode this resource type exists to prevent. Fix: Emit a distinct marker (e.g. note: "series metadata present but no values row returned") so the agent can tell a malformed response from real absence.

server/utils/db/db_utils.py
```diff
@@ -1409,6 +1409,63 @@ def initialize_tables():
...
+                        recommendation JSONB NOT NULL DEFAULT '{}'::jsonb,
+                        severity_score NUMERIC,
+                        slack_channel_id VARCHAR(64),
+                        slack_message_ts VARCHAR(64),
+                        action_run_id UUID,
+                        dismissed_by VARCHAR(255),
+                        dismissed_at TIMESTAMP,
+                        cooldown_until TIMESTAMP,
+                        created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
+                        updated_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP
```
damianloch: `cooldown_until` is written as aware UTC but compared with NOW(), the 30-day window drifts on a non-UTC session. TIMESTAMPTZ instead?

server/routes/slack/slack_events.py
```diff
@@ -556,4 +589,207 @@ def _handle_suggestion_details(payload: dict, action: dict, slack_user_id: str,
...
+                dismissed = dismiss_recommendation(cursor, rec_id, clicker_org_id, slack_user_id)
+                # Commit before the GitHub call so a slow API cannot hold the row lock.
+                conn.commit()
+
+        if not dismissed:
+            logger.info(f"Recommendation {rec_id} was already dismissed; no-op")
+            if client:
+                _send_ephemeral(client, channel_id, slack_user_id,
+                                "That recommendation was already dismissed.")
+            return jsonify({"text": ""}), 200
```
damianloch: The DB transition and PR close are unconditional, but every piece of feedback — ephemerals and _rewrite_dismissed_card — is gated on if client: / an early return when get_slack_client_for_user yields None, so the workload gets a 30-day cooldown and the PR gets closed while the card keeps its live Dismiss button and no message appears.

server/routes/slack/slack_events.py
```diff
@@ -556,4 +589,207 @@ def _handle_suggestion_details(payload: dict, action: dict, slack_user_id: str,
...
+                           "No cooldown applied.")
+        elif close_result.get("success"):
+            status_line = (f"*Status:* Dismissed by <@{slack_user_id}> -- PR #{pr_number} closed. "
+                           f"Aurora will not raise this workload again for {HPA_VPA_COOLDOWN_DAYS} days.")
+        else:
+            logger.error(
+                f"Failed to close PR #{pr_number} for recommendation {rec_id}: {close_result.get('error')}"
+            )
+            status_line = (f"*Status:* Dismissed by <@{slack_user_id}> -- PR #{pr_number} could not be "
+                           "closed automatically; please close it on GitHub.")
```
damianloch: status_line says "could not be closed automatically; please close it on GitHub", but the rewrite passes with_actions=False, which drops the entire actions block including the View PR button — so the one message asking for manual work deletes the only link to the PR.

server/routes/slack/slack_events.py
damianloch, line 1: Invalid UUID, unresolvable org, and unauthenticated-without-workspace-token all return an empty 200 with no ephemeral, so the click is indistinguishable from a broken button

server/routes/slack/slack_events.py
```diff
@@ -556,4 +589,207 @@ def _handle_suggestion_details(payload: dict, action: dict, slack_user_id: str,
...
+                if not row or row[0] != clicker_org_id:
+                    logger.warning(
+                        f"Slack user {sanitize(slack_user_id)} tried to dismiss recommendation outside their org"
+                    )
+                    if client:
+                        _send_ephemeral(client, channel_id, slack_user_id,
+                                        "Unauthorized: that recommendation does not belong to your organization.")
+                    return jsonify({"text": ""}), 200
+
+                dismissed = dismiss_recommendation(cursor, rec_id, clicker_org_id, slack_user_id)
```
damianloch, line 745: maybe there should be a ui confirmation for users if they want to mute it for 30 days or not

server/services/actions/hpa_vpa_recommendations.py
```diff
@@ -0,0 +1,488 @@
...
+
+
+def mark_superseded(cursor, rec_id: str) -> None:
+    """Retire a row whose mis-size worsened, or that a newer rec took over."""
+    cursor.execute(
+        """UPDATE hpa_vpa_recommendations
+              SET status = %s, cooldown_until = NULL, updated_at = NOW()
+            WHERE id = %s::uuid""",
+        (STATUS_SUPERSEDED, rec_id),
+    )
```
damianloch, line 527: mark_superseded is committed before the Slack post, but the post's compensation path only deletes the new claim, the superseded dismissal is never restored, so the remaining anti-nag window is gone for good.

server/services/actions/hpa_vpa_recommendations.py
```diff
@@ -0,0 +1,488 @@
...
+            body = resp.json()
+        except ValueError:
+            body = {}
+        # A merged PR means the human already accepted. GitHub reports the
+        # merge rather than erroring, and treating it as a rejection would
+        # start an anti-nag window on an *approved* change.
+        if body.get("merged"):
+            return {"success": True, "already_merged": True, "state": body.get("state")}
+        return {"success": True, "state": body.get("state", "closed")}
+    if resp.status_code == 404:
```
damianloch, line 666: i think this should be treated as a error or reqiures more code to tell if its realyl a success. Maybe a use toast saying not reachable

server/chat/backend/agent/tools/hpa_vpa_card_tool.py
damianloch, line 1: severity_score is an LLM-supplied argument. Someone clicks Dismiss and gets 30 days of quiet but the AI is allowed to break that silence early if the problem got 25% worse, but the AI is the one reporting how bad it is so this sort of doesnt make sense

server/services/actions/hpa_vpa_action.py
```diff
@@ -0,0 +1,201 @@
...
+  per-dimension current, recommended, and evidence values, and the autoscaler so the card can
+  explain a partial recommendation. Omit any dimension you are not recommending a change to.
+- If it returns `suppressed`, that is the anti-nag rule working correctly. Record it in the
+  living document and move on. **Do not work around it** -- do not re-post, rename the
+  workload, or open the PR anyway.
+- If you found no material mis-sizing anywhere: no PR, no card. Update the living document
+  with which workloads you checked, over what window, and what you found. That is a good
+  outcome, not a failed run.
+
+Hard anti-slop rules:
```
damianloch: this is user facing and doesn't sound professional

server/services/actions/hpa_vpa_action.py
damianloch, line 1: This sort of assumes the user already has slack connected. Its not really generalizable to the sense where anyone with any infra gets a similar expeirence

damianloch (changes requested):

server/routes/slack/slack_events.py
```diff
@@ -556,4 +589,207 @@ def _handle_suggestion_details(payload: dict, action: dict, slack_user_id: str,
...
+        #    except the opener's own.
+        pr_number = dismissed.get("pr_number")
+        close_as_user_id = dismissed.get("user_id") or clicker_user_id
+        close_result = close_pull_request(
+            close_as_user_id,
+            dismissed.get("vcs_provider") or "github",
+            dismissed.get("repo_full_name") or "",
+            pr_number,
+            timeout=_GITHUB_CLOSE_TIMEOUT,
+        ) if pr_number else {"error": "No PR recorded for this recommendation"}
```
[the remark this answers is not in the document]
isiddharthsingh: not fixed, deliberately. the clicker is already authenticated and org-verified, and default casbin policy gives actions:write to editor and above, so only viewers slip through. adding an enforcer call inside slack's ~3s ack budget costs real reliability for that narrow gap. worth doing when the pr close and card rewrite move to celery.

server/chat/backend/agent/tools/cloud_tools.py
isiddharthsingh: not fixed here, and you are right. i cannot gate it today: state.trigger_action_id is never set for background action runs, only the foreground /action path sets it. the precedent is is_postmortem_action, plumbed through chat/background/task.py via a system_key lookup, so it is about 10 lines there plus state.py. today it is gated on slack + github + not is_pr_review, so interactive users in a connected org do see the tool. your call whether that lands here or as a follow up.

server/chat/backend/agent/tools/datadog_tool.py
isiddharthsingh: fixed. values_missing is now tracked per series and passed into _summarize_series, which emits malformed_response: true with a note saying it is a broken response and not an idle workload. also surfaced malformed_series at the top level since a per row flag is easy to miss across 70 rows. went further than length: a null or non list entry inside a full length values array is caught too.

server/utils/db/db_utils.py
isiddharthsingh: fixed, and it was a live bug not a style point. all four columns are timestamptz now, plus a guarded migration for existing rows. proved it on a live table: 702 hours to expiry on both a utc and an america/los_angeles session, where before it drifted by the offset. the first migration i wrote was itself not idempotent and re-shifted 5 hours per boot, so it is now wrapped in a do block that only converts columns still naive.

server/routes/slack/slack_events.py
isiddharthsingh: fixed. _send_ephemeral now guards internally instead of every caller doing if client, and a new _slack_client_for_feedback falls back to the workspace token when the clicker has no personal slack credential. also added ephemerals to the invalid uuid and unresolvable org branches, so the db transition and the feedback can no longer diverge.

server/routes/slack/slack_events.py
isiddharthsingh: fixed. added a link_only mode to the card builder, used on the two branches that ask a human to close the pr themselves. dismiss is still dropped so the card cannot be re-clicked, but view pr survives since it is a url link out and carries no action. verified all three variants render: dismissed card has no buttons, manual work card keeps view pr only.

server/routes/slack/slack_events.py
isiddharthsingh: fixed. all three branches now send an ephemeral before returning the empty 200. the body stays empty on purpose because a non empty text in a block_actions response replaces the original message and would wipe the card, so the ephemeral is what makes the click distinguishable from a dead button.

server/routes/slack/slack_events.py
isiddharthsingh: not fixed, agreed it is a good idea but out of scope here. a confirm step needs a slack modal, views.open plus a view_submission handler, which is new interaction plumbing rather than a change to this flow. worth filing separately. the current mitigation is that dismiss is reversible in effect: the cooldown can be superseded if the mis-size materially worsens.

server/services/actions/hpa_vpa_recommendations.py
isiddharthsingh: fixed, good catch. added restore_superseded and it is called on every failed post path. also stopped nulling cooldown_until in mark_superseded: it is invisible to every read path since they all filter on status, but it was destroying the only record of how much window was left. verified live that claim, dismiss, supersede, restore round trips with cooldown_until preserved, and a double restore is a safe no op.

server/services/actions/hpa_vpa_recommendations.py
isiddharthsingh: fixed. github returns 404 for both deleted and no access deliberately, so it stays idempotent, but it now returns unverified: true with a detail string and the card says aurora could not confirm the pr is closed and asks the reviewer to check, rather than claiming it was closed. the view pr link is kept on that card so there is something to click.

server/chat/backend/agent/tools/hpa_vpa_card_tool.py
isiddharthsingh: fixed, this was the strongest catch. removed severity_score from the tool args entirely. severity is now derived server side from the same current and recommended display strings the model has to state anyway, via a kubernetes quantity parser. verified a model claiming a large mis-size while stating small values can no longer break a cooldown. also guarded cross unit division, since a 512m vs 256Mi typo used to produce severity 5.2e8 and break any cooldown.

server/services/actions/hpa_vpa_action.py
isiddharthsingh: fixed. renamed the block to hard rules. the rules themselves are unchanged, only the heading, since these instructions are user visible and editable per org in the actions ui.

server/services/actions/hpa_vpa_action.py
isiddharthsingh: fixed in the prompt, partly. step 9 now treats the card as one surface among several: it calls the tool if available, and if not it records the same per dimension evidence in the living document and says no card surface was available, so a run without slack is a complete run rather than a failed one. the tool itself is still slack only. google chat is the realistic second renderer since it has real callback buttons and a live CARD_CLICKED router, and the card builder already takes a neutral dict for that reason, so it is an addition not a rewrite. worth confirming whether that is needed before ship.

server/chat/backend/agent/tools/cloud_tools.py
damianloch: You can let this tool only be registered in list for agent to see if that specifc action is being ran, shouldnt be complicated

server/routes/slack/slack_events.py
damianloch: I don't think thats much work should only be a few lines of code. Even a small disclaimer is better than nothing here cuz the user doenst know that pressing the close PR button mutes it for 30 days. Maybe it was teh correct PR but its a duplicate for example. We are assuming the reason they close the PR which shouldnt be done blindly

server/chat/backend/agent/tools/cloud_tools.py
isiddharthsingh: you were right. action_id is in trigger_metadata, so i resolved is_hpa_vpa_action in task.py via a system_key lookup, generalising the existing postmortem helper. fails closed, and added to the tool cache key.

server/routes/slack/slack_events.py
isiddharthsingh: you were right, no modal needed. slack buttons take a native confirm field — plain block payload, existing handler unchanged, cancel never reaches the server.

 dialog: "Closes PR #x and mutes this workload for 30 days. Only want the PR gone (duplicate, or you will handle it)? Close it on the repo instead, so Aurora keeps watching this workload."

note: my first version failed with invalid_blocks — 356 chars against slack's 300 limit kills the whole message. now 183, with _clip guards and tests pinning the limits.

damianloch (approved):

PR #599, opened 2026-08-14
Add Gemini 3.6 Flash and 3.5 Flash-Lite as supported Vertex models.

damianloch:
Replace 3.5 Flash in the UI picker with 3.6 Flash, add Flash-Lite for guardrails and high-throughput use, and update docs, pricing, and routing accordingly. Also limited what models are shown in the model picker based on api keys and llm provider mode that is configured

## Summary by CodeRabbit

- **New Features**
  - Added dynamic model availability based on configured providers and credentials.
  - Added support for Gemini 3.6 Flash, Gemini 3.5 Flash, and Gemini 3.5 Flash Lite across Google AI and Vertex AI.
  - Improved model selection with validation, fallback handling, and provider-specific visibility.

- **Bug Fixes**
  - Prevented unsupported thinking configuration for Flash Lite models.
  - Added accurate context limits and pricing for newly supported models.

- **Documentation**
  - Updated deployment and Vertex AI configuration examples for the latest Gemini models.


isiddharthsingh:
tested this locally against vertex and the models themselves are fine, 3.6 flash and 3.5 flash-lite both respond and the flash-lite thinking skip does what it should.

few things on pricing though. google/gemini-3.5-flash and vertex/gemini-3.5-flash are set to 0.0005/0.003 but google lists 3.5 flash at $1.50 in / $9.00 out per 1M, so it's 3x under. looks like those got copied from the existing gemini-3-flash row since the numbers are identical. worth fixing because 3.5 flash ends up as catalog[0] in vertex mode, so it's the default model people land on. the 3.6 flash numbers are correct today but that's introductory pricing that goes to $1.50/$7.50 on jan 1 2027, so maybe worth a comment next to it so it doesn't quietly go stale. also flash-lite has a cached_input value but google lists context caching as "not available" for that model. (coderabbit's pricing comment was right on both counts, it was quoting 3.6's post-2027 rate and 3.5's current rate.)

on the frontend, the localstorage migration in ModelSelector.tsx clears the saved model instead of persisting it. hit this live: switching to vertex correctly migrated my selection to vertex/gemini-3.1-pro-preview, but localStorage.selectedModel came back null and after a reload it fell through to catalog[0] (gemini 3.5 flash) instead of what it had just migrated to. the else if on line 204 can never run when savedModel is truthy, so the migrated value never gets written. picking from the dropdown manually persists fine, it's only the auto migration path.

last one, changing the effect dep from [] to [visibleModels] means it now reruns every time the picker query revalidates on tab focus, since useQuery hands back a fresh object each fetch. if you open a saved session that used a different model, ChatClient.tsx:202 sets selectedModel from the session uiState without touching localstorage, so the next revalidation quietly resets you back to the localstorage model mid conversation, and the next message writes that wrong model back into the session. previously the effect was mount-only so it couldn't stomp on a restored session.

isiddharthsingh (approved):

PR #604, opened 2026-08-20
Add Bitbucket repo select-all and bulk Incident Prevention enable

damianloch:
## Summary
- Add GitHub-style select-all (and a filter when there are more than 10 repos) to the Bitbucket repo picker so a workspace can be connected without checking repos one by one.
- Add `PUT /bitbucket/repo-selections/change-gating` plus an **Enable Incident Prevention on all** control on Connected Repositories. Connecting a repo still does not turn gating on.
- If webhook auto-create fails, show one shared URL/secret dialog instead of a modal per repo.

Came out of the Bombora POC kickoff (Aug 20): Juan has 49 Bitbucket repos connected and one-by-one Incident Prevention toggles are not usable.

## Test plan
- [ ] Bitbucket connector → pick a workspace with many repos: Select all checks every listed repo; Save connects them. Filter + Select all only affects the filtered set.
- [ ] Saving a selection does **not** enable Incident Prevention.
- [ ] Connected Repositories → Enable Incident Prevention on all: all remaining repos turn on; per-repo switches stay usable.
- [ ] If `write:webhook:bitbucket` works: toast only, no 49 dialogs. If it does not: one dialog with the shared URL/secret.
- [ ] Repos with no default branch are skipped with an error toast; already-enabled repos are left alone in Bitbucket.

## Summary by CodeRabbit

* **New Features**
  * Added repository search, filtering, and filtered “select all” support for larger workspaces.
  * Added bulk Incident Prevention enablement and disablement for selected repositories.
  * Added multi-repository webhook setup with progress tracking and notifications.
  * Added selected-repository disconnect handling.

* **Improvements**
  * Automatically configured webhooks are recognized without unnecessary setup.
  * Repositories without default branches are clearly skipped.
  * Changing workspaces clears the repository search filter.
  * Improved handling of empty repositories, temporary Bitbucket rate limits, and connection errors.

### DIFF
--- a/client/src/components/bitbucket-workspace-browser.tsx
+++ b/client/src/components/bitbucket-workspace-browser.tsx
@@ -80,16 +80,12 @@ export default function BitbucketWorkspaceBrowser() {
     setGatingUpdating(prev => new Set(prev).add(repoFullName));
     try {
       const result = await BitbucketIntegrationService.updateChangeGating(repoFullName, enabled);
-      // webhook_configured is never set optimistically: creating a hook is not
-      // evidence that deliveries reach Aurora. It stays false until Verify or a
-      // real delivery confirms it, so the badge can't claim a working setup.
       setSavedRepos(prev => prev.map(r =>
         r.full_name === repoFullName
-          ? { ...r, change_gating_enabled: enabled, webhook_configured: false, webhook_stale: false }
+          ? { ...r, change_gating_enabled: enabled, webhook_configured: !!(enabled && result.webhook_auto_created), webhook_stale: false }
           : r
       ));
-      if (enabled && result.webhook_url) {
-        // Always walk the user through webhook setup/verification on enable.
+      if (enabled && result.webhook_url && result.webhook_auto_created === false) {
         setWebhookSetup(result);
       } else if (!enabled && webhookSetup?.repo_full_name === repoFullName) {
         setWebhookSetup(null);
@@ -119,19 +115,30 @@ export default function BitbucketWorkspaceBrowser() {
   };
 
   const handleBulkEnable = async () => {
-    const names = savedRepos.filter(r => !r.change_gating_enabled).map(r => r.full_name);
-    if (names.length === 0) return;
+    const names = savedRepos
+      .filter(r => r.workspace === selectedWorkspace && checkedRepos.has(r.slug) && !r.change_gating_enabled)
+      .map(r => r.full_name);
+    if (names.length === 0) {
+      const checkedConnected = [...checkedRepos].some(s => savedReposByWorkspace.get(selectedWorkspace)?.has(s));
+      toast({
+        title: checkedConnected ? "Incident Prevention is already on for the selected connected repos" : "Save selected repos first",
+      });
+      return;
+    }
     setBulkEnabling(true);
     try {
-      const result = await BitbucketIntegrationService.updateChangeGatingBulk(names);
-      const byName = new Map(result.results.map(r => [r.repo_full_name, r]));
-      setSavedRepos(prev => prev.map(r => {
-        const row = byName.get(r.full_name);
-        if (!row || row.error) return r;
-        return { ...r, change_gating_enabled: true, webhook_configured: false, webhook_stale: false };
-      }));
-      const manual = result.results.filter(r => r.webhook_auto_created === false);
-      if (manual.length && result.webhook_url) {
+      const job = await BitbucketIntegrationService.updateChangeGatingBulk(names);
+      let status = await BitbucketIntegrationService.getChangeGatingBulkJob(job.task_id);
+      while (!status.complete) {
+        await new Promise(r => setTimeout(r, 2000));
+        status = await BitbucketIntegrationService.getChangeGatingBulkJob(job.task_id);
+      }
+      if (status.error) throw new Error(status.status || "Failed to enable Incident Prevention");
+      const result = status.result;
+      const data = await BitbucketIntegrationService.loadWorkspaceSelection();
+      if (data?.repositories) setSavedRepos(parseConnectedRepos(data.repositories));
+      const manual = result?.results.filter(r => r.webhook_auto_created === false) ?? [];
+      if (manual.length && result?.webhook_url) {
         setWebhookSetup({
           repo_full_name: manual[0].repo_full_name,
           change_gating_enabled: true,
@@ -142,24 +149,59 @@ export default function BitbucketWorkspaceBrowser() {
           manual_count: manual.length,
         });
       } else {
-        const enabled = result.results.filter(r => !r.error).length;
+        const enabled = result?.results.filter(r => !r.error).length ?? 0;
         if (enabled) {
           toast({ title: `Incident Prevention enabled on ${enabled} repo${enabled === 1 ? '' : 's'}` });
         }
       }
-      const skipped = result.results.filter(r => r.error === 'no_default_branch').length;
+      const skipped = result?.results.filter(r => r.error === 'no_default_branch').length ?? 0;
       if (skipped) {
         toast({
           title: `${skipped} repo${skipped === 1 ? '' : 's'} skipped`,
-          description: 'No default branch recorded. Re-save the repository selection, then try again.',
-          variant: 'destructive',
+          description: "No default branch recorded. Re-save the repository selection, then try again.",
+          variant: "destructive",
         });
       }
     } catch (error: unknown) {
       const err = error as Error;
-      toast({ title: 'Error', description: err.message || 'Failed to enable Incident Prevention', variant: 'destructive' });
+      toast({ title: "Error", description: err.message || "Failed to enable Incident Prevention", variant: "destructive" });
     } finally {
       setBulkEnabling(false);
+    }
+  };
+
+  const handleDisconnectSelected = async () => {
+    if (!selectedWorkspace) return;
+    const saved = savedReposByWorkspace.get(selectedWorkspace);
+    if (!saved?.size) return;
+    const remainingSlugs = [...saved].filter(s => !checkedRepos.has(s));
+    const removed = saved.size - remainingSlugs.length;
+    setIsSaving(true);
+    try {
+      await BitbucketIntegrationService.saveWorkspaceSelection({
+        workspace: selectedWorkspace,
+        repositories: remainingSlugs.map(slug => {
+          const repo = repos.find(r => r.slug === slug);
+          return repo ?? { slug, name: slug, full_name: `${selectedWorkspace}/${slug}`, is_private: false };
+        }),
+      });
+      setCheckedRepos(new Set(remainingSlugs));
+      setSavedReposByWorkspace(prev => {
+        const next = new Map(prev);
+        if (remainingSlugs.length) next.set(selectedWorkspace, new Set(remainingSlugs));
+        else next.delete(selectedWorkspace);
+        savedReposByWorkspaceRef.current = next;
+        return next;
+      });
+      const data = await BitbucketIntegrationService.loadWorkspaceSelection();
+      setSavedRepos(data?.repositories ? parseConnectedRepos(data.repositories) : []);
+      window.dispatchEvent(new CustomEvent('providerStateChanged'));
+      toast({ title: "Disconnected", description: `${removed} repo${removed === 1 ? '' : 's'} disconnected` });
+    } catch (error: unknown) {
+      const err = error as Error;
+      toast({ title: "Error", description: err.message || "Failed to disconnect", variant: "destructive" });
+    } finally {
+      setIsSaving(false);
     }
   };
 
@@ -554,17 +596,15 @@ export default function BitbucketWorkspaceBrowser() {
             </div>
           ) : repos.length > 0 ? (
             <div className="space-y-1">
-              {repos.length > 10 && (
-                <div className="relative">
-                  <Search className="absolute left-2 top-1/2 -translate-y-1/2 h-3 w-3 text-muted-foreground" />
-                  <Input
-                    placeholder="Filter repositories..."
-                    value={searchFilter}
-                    onChange={(e) => setSearchFilter(e.target.value)}
-                    className="h-8 text-xs pl-7"
-                  />
-                </div>
-              )}
+              <div className="relative">
+                <Search className="absolute left-2 top-1/2 -translate-y-1/2 h-3 w-3 text-muted-foreground" />
+                <Input
+                  placeholder="Filter repositories..."
+                  value={searchFilter}
+                  onChange={(e) => setSearchFilter(e.target.value)}
+                  className="h-8 text-xs pl-7"
+                />
+              </div>
               {visibleRepos.length > 0 && (
                 <label className="flex items-center gap-2 px-2 py-1 text-xs text-muted-foreground cursor-pointer hover:text-foreground">
                   <Checkbox
@@ -626,17 +666,34 @@ export default function BitbucketWorkspaceBrowser() {
       )}
 
       {selectedWorkspace && repos.length > 0 && (
-        <div className="flex items-center gap-2">
+        <div className="flex items-center gap-2 flex-wrap">
           <Button
             size="sm"
             onClick={handleSave}
-            disabled={isSaving || checkedRepos.size === 0 || !selectionChanged}
+            disabled={isSaving || bulkEnabling || checkedRepos.size === 0 || !selectionChanged}
           >
             {isSaving ? <Loader2 className="h-3.5 w-3.5 animate-spin mr-1.5" /> : null}
             Save
           </Button>
+          {!!currentWorkspaceSaved?.size && [...checkedRepos].some(s => currentWorkspaceSaved.has(s)) && (
+            <Button size="sm" variant="outline" onClick={handleDisconnectSelected} disabled={isSaving || bulkEnabling}>
+              Disconnect selected
+            </Button>
+          )}
+          {incidentPreventionEnabled && checkedRepos.size > 0 && (
+            <Button
+              size="sm"
+              variant="outline"
+              disabled={isSaving || bulkEnabling}
+              onClick={handleBulkEnable}
+              data-testid="bb-enable-gating-selected"
+            >
+              {bulkEnabling ? <Loader2 className="h-3.5 w-3.5 animate-spin mr-1.5" /> : null}
+              Enable Incident Prevention
+            </Button>
+          )}
           {totalSavedRepos > 0 && (
-            <Button size="sm" variant="outline" onClick={handleClear}>
+            <Button size="sm" variant="outline" onClick={handleClear} disabled={isSaving || bulkEnabling}>
               Clear All
             </Button>
           )}
@@ -645,22 +702,7 @@ export default function BitbucketWorkspaceBrowser() {
 
       {savedRepos.length > 0 && (
         <div className="space-y-2 pt-2 border-t border-border">
-          <div className="flex items-center justify-between gap-2">
-            <p className="text-sm font-medium text-muted-foreground">Connected Repositories</p>
-            {incidentPreventionEnabled && savedRepos.some(r => !r.change_gating_enabled) && (
-              <Button
-                size="sm"
-                variant="outline"
-                className="h-7 text-xs"
-                disabled={bulkEnabling}
-                onClick={handleBulkEnable}
-                data-testid="bb-enable-all-gating"
-              >
-                {bulkEnabling ? <Loader2 className="h-3 w-3 animate-spin mr-1.5" /> : null}
-                Enable Incident Prevention on all
-              </Button>
-            )}
-          </div>
+          <p className="text-sm font-medium text-muted-foreground">Connected Repositories</p>
           {savedRepos.map(repo => {
             const isEditing = editingMetadata[repo.full_name] !== undefined;
             const isReady = repo.metadata_status === 'ready';
--- a/client/src/components/bitbucket-workspace-browser.tsx
+++ b/client/src/components/bitbucket-workspace-browser.tsx
@@ -134,33 +134,43 @@ export default function BitbucketWorkspaceBrowser() {
       const job = await BitbucketIntegrationService.updateChangeGatingBulk(names, enabled);
       let status = await BitbucketIntegrationService.getChangeGatingBulkJob(job.task_id);
       const deadline = Date.now() + 10 * 60 * 1000;
+      let consecutiveErrors = 0;
       while (!status.complete) {
         if (Date.now() > deadline) {
           throw new Error('Incident Prevention is taking longer than expected. Reload this page to see the current state.');
         }
-        const data = await BitbucketIntegrationService.loadWorkspaceSelection();
-        if (data?.repositories) {
-          const updated = parseConnectedRepos(data.repositories);
-          setSavedRepos(updated);
-          setGatingUpdating(prev => {
-            const next = new Set(prev);
-            for (const r of updated) {
-              if (!next.has(r.full_name)) continue;
-              const settled = enabled
-                ? r.webhook_configured || !r.change_gating_enabled
-                : !r.change_gating_enabled;
-              if (settled) next.delete(r.full_name);
+        // A failed poll is not evidence the job failed — the worker runs
+        // regardless. Only give up once several in a row fail; a real task
+        // failure arrives as status.error and is handled after the loop.
+        try {
+          const data = await BitbucketIntegrationService.loadWorkspaceSelection();
+          if (data?.repositories) {
+            const updated = parseConnectedRepos(data.repositories);
+            setSavedRepos(updated);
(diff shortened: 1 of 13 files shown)


client/src/components/bitbucket-workspace-browser.tsx
```diff
@@ -115,6 +118,51 @@ export default function BitbucketWorkspaceBrowser() {
...
+      if (manual.length && result.webhook_url) {
+        setWebhookSetup({
+          repo_full_name: manual[0].repo_full_name,
+          change_gating_enabled: true,
+          webhook_url: result.webhook_url,
+          webhook_secret: result.webhook_secret,
+          webhook_events: result.webhook_events,
+          webhook_auto_created: false,
+          manual_count: manual.length,
+        });
```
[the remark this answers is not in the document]
damianloch: Skipping: URL and secret are org-shared, so the admin pastes once. The count is enough; listing every repo would not change the setup step.

client/src/components/bitbucket-workspace-browser.tsx
```diff
@@ -500,34 +553,71 @@ export default function BitbucketWorkspaceBrowser() {
...
-                        {repo.is_private ? 'Private' : 'Public'}
-                      </Badge>
+                  <span>
+                    Select all
+                    {searchFilter ? ' (filtered)' : ''}
+                    {' · '}
+                    {visibleRepos.length} repos
+                  </span>
+                </label>
+              )}
```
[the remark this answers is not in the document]
damianloch: Skipping: our Checkbox indicator is only a check mark, so Radix indeterminate would look like checked. Not worth a custom dash indicator on this picker.

client/src/services/bitbucket-integration-service.ts
```diff
@@ -37,6 +37,21 @@ export interface ChangeGatingResponse {
   webhook_note?: string;
   webhook_auto_created?: boolean;
   webhook_cleanup_failed?: boolean;
+  manual_count?: number;
+}
```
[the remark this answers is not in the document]
damianloch: Skipping: manual_count is only written by the picker before setWebhookSetup. A second view type would not change runtime.

server/routes/bitbucket/bitbucket_selection.py
```diff
@@ -420,6 +420,113 @@ def _store_hook_uuid(hook_uuid: str) -> None:
...
+                })
+
+        logger.info(
+            "Bitbucket Incident Prevention bulk-enabled %s repos (org %s, by user %s)",
+            len(to_enable), _sanitize_log(org_id), _sanitize_log(user_id),
+        )
+        return jsonify(response)
+    except Exception:
+        logger.exception("Error bulk-updating Bitbucket change gating")
+        return jsonify({"error": "Failed to update Incident Prevention"}), 500
```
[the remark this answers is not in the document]
damianloch: Skipping: extracting helpers here is a style-only refactor and would force a full retest of the bulk route. Leaving the validation inline.

server/routes/bitbucket/bitbucket_selection.py
```diff
@@ -420,6 +420,113 @@ def _store_hook_uuid(hook_uuid: str) -> None:
...
+            for name in to_enable:
+                if name in already_on:
+                    results.append({"repo_full_name": name})
+                    continue
+                results.append({
+                    "repo_full_name": name,
+                    "webhook_auto_created": _try_auto_create_hook(
+                        user_id, org_id, name, webhook_url, webhook_events, secret
+                    ),
+                })
```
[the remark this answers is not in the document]
damianloch: Already done: bulk enable/disable is a Celery task (bitbucket.enable_change_gating_bulk). The Flask route only enqueues and returns 202.

client/src/components/bitbucket-workspace-browser.tsx
```diff
@@ -115,6 +114,97 @@ export default function BitbucketWorkspaceBrowser() {
...
+      return;
+    }
+    setBulkEnabling(true);
+    try {
+      const job = await BitbucketIntegrationService.updateChangeGatingBulk(names);
+      let status = await BitbucketIntegrationService.getChangeGatingBulkJob(job.task_id);
+      while (!status.complete) {
+        await new Promise(r => setTimeout(r, 2000));
+        status = await BitbucketIntegrationService.getChangeGatingBulkJob(job.task_id);
+      }
```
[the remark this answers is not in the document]
damianloch: Fixed: the poll loop now has a 10-minute deadline and throws so bulkEnabling clears. Disable still returns as soon as the DB flag is off.

client/src/components/bitbucket-workspace-browser.tsx
```diff
@@ -115,6 +114,97 @@ export default function BitbucketWorkspaceBrowser() {
...
+    const remainingSlugs = [...saved].filter(s => !checkedRepos.has(s));
+    const removed = saved.size - remainingSlugs.length;
+    setIsSaving(true);
+    try {
+      await BitbucketIntegrationService.saveWorkspaceSelection({
+        workspace: selectedWorkspace,
+        repositories: remainingSlugs.map(slug => {
+          const repo = repos.find(r => r.slug === slug);
+          return repo ?? { slug, name: slug, full_name: `${selectedWorkspace}/${slug}`, is_private: false };
+        }),
```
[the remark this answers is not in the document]
damianloch: Fixed: Disconnect no longer sends is_private: false for catalog-miss fallbacks. Save uses COALESCE(EXCLUDED.is_private, connected_repos.is_private) when the field is omitted.

server/routes/bitbucket/bitbucket_selection.py
```diff
@@ -345,6 +348,18 @@ def cleanup_org_hooks(user_id: str, org_id: str, repo_full_names=None) -> list[s
...
+def _until_not_429(fn):
+    delay = 1
+    result = fn()
+    for _ in range(7):
+        if not (isinstance(result, dict) and result.get("status") == 429):
+            return result
+        time.sleep(delay)
+        delay = min(delay * 2, 32)
+        result = fn()
+    return result
```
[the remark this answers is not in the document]
damianloch: Skipping: shrinking _until_not_429 on the per-row Flask path reintroduces the 429 failures we already hit. Bulk work is on Celery; leaving the retry budget as-is.

server/routes/bitbucket/bitbucket_selection.py
```diff
@@ -420,6 +437,120 @@ def _store_hook_uuid(hook_uuid: str) -> None:
...
+        def _one(name):
+            return {
+                "repo_full_name": name,
+                "webhook_auto_created": _try_auto_create_hook(
+                    user_id, org_id, name, webhook_url, webhook_events, secret
+                ),
+            }
+        with ThreadPoolExecutor(max_workers=3) as pool:
+            results.extend(pool.map(_one, need_hooks))
+    return out
```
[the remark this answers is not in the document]
damianloch: Fixed: the task result never includes webhook_secret (poll attaches it from Vault after SUCCESS). It now stores org_id; the poll endpoint 404s on mismatch and strips org_id before JSON.

server/routes/bitbucket/bitbucket_selection.py
```diff
@@ -420,6 +437,120 @@ def _store_hook_uuid(hook_uuid: str) -> None:
...
+@bitbucket_selection_bp.route("/repo-selections/change-gating/jobs/<task_id>", methods=["GET"])
+@require_permission("connectors", "write")
+def get_change_gating_bulk_job(user_id, task_id):
+    """Poll the Celery bulk-enable job."""
+    task = enable_change_gating_bulk.AsyncResult(task_id)
+    if task.state in ("PENDING", "STARTED"):
+        return jsonify({"state": task.state, "complete": False})
+    if task.state == "SUCCESS":
+        return jsonify({"state": task.state, "complete": True, "result": task.result or {}})
+    return jsonify({"state": task.state, "complete": True, "error": True, "status": str(task.info)}), 200
```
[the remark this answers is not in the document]
damianloch: Fixed: FAILURE now logs task.info server-side and returns a generic 'Failed to update Incident Prevention' to the client.

server/celery_config.py
```diff
@@ -75,6 +75,11 @@
     worker_prefetch_multiplier=1,  # Process one task at a time
     broker_connection_retry_on_startup=True,  # Explicitly enable for Celery 6.0+
     result_expires=3600,  # Expire task results after 1 hour (backend= is set above)
+    # ponytail: workers listen `-Q high,celery` (high first). Interactive
+    # Bitbucket Incident Prevention must not sit behind a Save of N metadata jobs.
+    task_routes={
+        "bitbucket.enable_change_gating_bulk": {"queue": "high"},
+    },
```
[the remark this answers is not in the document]
damianloch: Skipping: workers already listen -Q high,celery / --queues=high,celery. A dedicated high-only worker is an infra split we are not taking in this PR.

server/connectors/bitbucket_connector/api_client.py
```diff
@@ -112,9 +113,19 @@ def _handle_error(self, response):
...
+        delay = 1
+        response = None
+        for _ in range(6):
+            response = requests.get(
+                url, headers=self._get_headers(), params=params, timeout=self.REQUEST_TIMEOUT
+            )
+            if response.status_code != 429:
+                break
+            time.sleep(delay)
+            delay = min(delay * 2, 16)
```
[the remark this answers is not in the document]
damianloch: Fixed: the three GET retry loops no longer sleep after the sixth 429. Same six attempts, faster give-up.

server/routes/bitbucket/bitbucket_selection.py
```diff
@@ -309,10 +312,15 @@ def cleanup_org_hooks(user_id: str, org_id: str, repo_full_names=None) -> list[s
...
+        repo_full_name, hook_uuid = item
+        return repo_full_name, _delete_repo_hook(user_id, repo_full_name, hook_uuid)
+
     deleted, failed = [], []
-    for repo_full_name, hook_uuid in hooks:
-        target = deleted if _delete_repo_hook(user_id, repo_full_name, hook_uuid) else failed
-        target.append(repo_full_name)
+    with ThreadPoolExecutor(max_workers=3) as pool:
+        for repo_full_name, ok in pool.map(_one, hooks):
+            (deleted if ok else failed).append(repo_full_name)
```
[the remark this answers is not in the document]
damianloch: Skipping: this is sized for tens of repos (Bombora-scale), not thousands of 30s timeouts. Splitting into batched tasks is a larger change than this PR.

server/tests/tasks/test_repo_metadata_bitbucket.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+@patch("utils.repo_metadata._generate_summary")
+@patch("utils.repo_metadata._update_metadata")
+@patch("utils.repo_metadata._fetch_repo_context", return_value=("", ""))
+@patch("utils.repo_metadata._get_credentials", return_value={"access_token": "x"})
+def test_empty_repo_marked_ready(get_creds, fetch, update, summarize):
+    generate_repo_metadata.run("user-1", "bitbucket", "ws/empty")
+    summarize.assert_not_called()
+    update.assert_called_with("user-1", "bitbucket", "ws/empty", _EMPTY_SUMMARY, "ready")
```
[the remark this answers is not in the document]
damianloch: Already done: those generate_repo_metadata.run tests were removed in c03af38a. CI stubs Celery so .run() is a no-op; the _fetch_bitbucket_context tests still pin empty vs error.

server/utils/repo_metadata.py
```diff
@@ -91,37 +91,71 @@ def _fetch_gitlab_listing(base_url: str, token: str, project_path: str) -> str:
...
-    for filename in ("README.md", "README.rst", "README.txt", "README"):
-        result = client.get_file_contents(workspace, repo_slug, filename)
+    repo = client.get_repository(workspace, repo_slug)
+    if not isinstance(repo, dict) or repo.get("error"):
+        return "", "(could not list files)"
+    if repo.get("size") == 0:
+        return "", ""
+    ref = (repo.get("mainbranch") or {}).get("name") or "HEAD"
+    commit = client._resolve_commit(workspace, repo_slug, ref)
+    listing_raw = client.get_directory_tree(workspace, repo_slug, "", commit=commit, list_files=True)
```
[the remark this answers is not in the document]
damianloch: Fixed: get_directory_tree(list_files=True) now requests pagelen=100 so README candidates past Bitbucket's default page of 10 are visible.

server/utils/repo_metadata.py
```diff
@@ -140,13 +174,14 @@ def _update_metadata(user_id: str, provider: str, repo_full_name: str, summary:
...
+            org_id = set_rls_context(cur, conn, user_id, log_prefix=f"[{provider.title()}Metadata]")
+            if not org_id:
                 return
             cur.execute(
                 """UPDATE connected_repos
                    SET metadata_summary = %s, metadata_status = %s, updated_at = NOW()
-                   WHERE user_id = %s AND provider = %s AND repo_full_name = %s""",
-                (summary, status, user_id, provider, repo_full_name),
+                   WHERE provider = %s AND repo_full_name = %s""",
+                (summary, status, provider, repo_full_name),
```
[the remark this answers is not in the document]
damianloch: Disagree: connected_repos rows for Bitbucket are org-shared (duplicates per user, writes to all org rows). RLS already scopes the UPDATE. Putting user_id back would leave sibling users with stale 'Failed to generate description'.

client/src/components/bitbucket-workspace-browser.tsx
```diff
@@ -500,34 +618,69 @@ export default function BitbucketWorkspaceBrowser() {
...
+            <div className="space-y-1">
+              <div className="relative">
+                <Search className="absolute left-2 top-1/2 -translate-y-1/2 h-3 w-3 text-muted-foreground" />
+                <Input
+                  placeholder="Filter repositories..."
+                  value={searchFilter}
+                  onChange={(e) => setSearchFilter(e.target.value)}
+                  className="h-8 text-xs pl-7"
+                />
+              </div>
```
isiddharthsingh, line 664: Give the repository filter Input an accessible name by adding an appropriate label or aria-label while keeping the existing searchFilter value and onChange behavior unchanged; treat the adjacent Search icon as decorative.

client/src/components/bitbucket-workspace-browser.tsx
```diff
@@ -115,6 +115,129 @@ export default function BitbucketWorkspaceBrowser() {
...
+            title: `${skipped} repo${skipped === 1 ? '' : 's'} skipped`,
+            description: 'No default branch recorded. Re-save the repository selection, then try again.',
+            variant: 'destructive',
+          });
+        }
+      } else {
+        toast({
+          title: `Incident Prevention disabled on ${names.length} repo${names.length === 1 ? '' : 's'}`,
+        });
+      }
```
isiddharthsingh, line 229: Update the enabled-state success toast in the result handling around setWebhookSetup to count result entries with no error instead of using names.length. Preserve the existing skipped-repository toast and manual webhook setup behavior, and ensure the enabled count reflects only repositories successfully processed.

server/routes/bitbucket/bitbucket_selection.py
```diff
@@ -345,6 +353,18 @@ def cleanup_org_hooks(user_id: str, org_id: str, repo_full_names=None) -> list[s
...
+    result = fn()
+    for _ in range(7):
+        if not (isinstance(result, dict) and result.get("status") == 429):
+            return result
+        time.sleep(delay)
+        delay = min(delay * 2, 32)
+        result = fn()
+    return result
+
+
```
isiddharthsingh: Make _until_not_429 accept an explicit retry-attempt budget while preserving the existing full budget by default; thread this parameter through _try_auto_create_hook, passing a capped budget from update_change_gating and the default full budget from _apply_change_gating_bulk.

server/routes/bitbucket/bitbucket_selection.py
```diff
@@ -420,6 +442,166 @@ def _store_hook_uuid(hook_uuid: str) -> None:
...
+    logger.error(
+        "Bitbucket bulk-enable task %s ended in state %s: %s",
+        _sanitize_log(task_id), task.state, task.info,
+    )
+    return jsonify({
+        "state": task.state,
+        "complete": True,
+        "error": True,
+        "status": "Failed to update Incident Prevention",
+    }), 200
```
isiddharthsingh: other handlers are in try/except block also wrap the poll handler in try/except with logging.

client/src/components/bitbucket-workspace-browser.tsx
```diff
@@ -115,6 +115,129 @@ export default function BitbucketWorkspaceBrowser() {
...
+          toast({
+            title: `${skipped} repo${skipped === 1 ? '' : 's'} skipped`,
+            description: 'No default branch recorded. Re-save the repository selection, then try again.',
+            variant: 'destructive',
+          });
+        }
+      } else {
+        toast({
+          title: `Incident Prevention disabled on ${names.length} repo${names.length === 1 ? '' : 's'}`,
+        });
```
isiddharthsingh, line 228: bulk disable always shows the plain "disabled on N repos" toast and never surfaces webhook_cleanup_failed. the worker returns it from _apply_change_gating_bulk, but the disable path breaks out of the poll loop as soon as the db flags flip (line 156-160), before the job finishes — so status.result is usually still empty and the cleanup info never even reaches the client. the single-repo toggle handles this correctly (line 94-101) and the backend comment literally calls undeletable web-ui hooks "the COMMON case, not an edge one". so after a bulk disable, orphaned hooks keep posting pr events to aurora and the admin gets zero indication manual cleanup is needed. either keep polling to completion on disable, or keep the early break for the spinners but continue a detached poll and show the warning toast when the result lands.

server/routes/bitbucket/bitbucket_selection.py
```diff
@@ -399,18 +423,16 @@ def _store_hook_uuid(hook_uuid: str) -> None:
...
             ws, slug, webhook_url, webhook_events,
             description="Aurora Incident Prevention",
             secret=secret,
-        )
+        ))
         if not isinstance(created, dict) or created.get("error"):
             return False
-        if created.get("uuid"):
-            _store_hook_uuid(created["uuid"])
+        _mark_verified(created.get("uuid"))
```
isiddharthsingh, line 450: the newly-created-hook branch calls _mark_verified unconditionally after create_webhook succeeds, but the existing-hook branch right above it (422-426) requires active + both pullrequest triggers before marking verified. the docstring says creating the hook is "the same bar as Verify" — but it isn't, the new-hook path skips the checks the existing-hook path enforces. probably fine in practice since we just created it with those exact events, but if bitbucket ever returns a hook that's created-but-inactive we'd stamp webhook_verified_at/url on something that never fires. worth applying the same active/events check to the create response since it's right there in created.

server/routes/bitbucket/bitbucket_selection.py
isiddharthsingh, line 397: this comment is stale now — it still says "the cost of a lost uuid is only that disable/disconnect can't auto-delete this hook later", but _mark_verified also writes webhook_verified_at and webhook_verified_url now. a swallowed db blip here silently drops the verification too, so the hook exists and webhook_auto_created comes back true while webhook_configured stays false. the warning log has the same problem, only mentions uuid persistence. update both so whoever hits this log line knows verification state was also lost.

client/src/components/bitbucket-workspace-browser.tsx
```diff
@@ -115,6 +115,129 @@ export default function BitbucketWorkspaceBrowser() {
...
+          });
+          // Disable is done once the DB flag is off. Hook deletes keep
+          // running in the worker; the UI does not wait on them.
+          if (!enabled && names.every(n => !updated.find(r => r.full_name === n)?.change_gating_enabled)) {
+            break;
+          }
+        }
+        await new Promise(r => setTimeout(r, 2000));
+        status = await BitbucketIntegrationService.getChangeGatingBulkJob(job.task_id);
+      }
```
isiddharthsingh, line 196: the poll loop runs for up to 10 minutes with no per-iteration try/catch — one transient 5xx on loadWorkspaceSelection or getChangeGatingBulkJob throws out to the generic error toast and clears all the spinners, while the celery task keeps running server-side. so the ui reports failure on a bulk op that probably succeeded, and the natural reaction is to hit the button again. wrap the per-iteration calls and tolerate a few consecutive failures before giving up.

server/connectors/bitbucket_connector/api_client.py
```diff
@@ -185,14 +196,26 @@ def _paginated_get(self, url, params=None, page_limit=100):
                 _validate_bitbucket_url(url)
             except ValueError:
                 logger.warning("Pagination rejected untrusted next URL: %s", _sanitize_url(url))
+                if all_values:
+                    return all_values
```
isiddharthsingh: the new if all_values: return all_values inside the except ValueError makes a rejected pagination url indistinguishable from a legitimately short list. that validator halt was made an explicit error dict on purpose in the #302 security sweep so callers could tell the data might be incomplete. if partial-on-bad-next-url is intentional for the bulk flow, at least log it at warning like the http-error partial path below does, so it's visible.

client/src/components/bitbucket-workspace-browser.tsx
damianloch: Fixed. Added `aria-label`, plus `aria-hidden` on the icon.

client/src/components/bitbucket-workspace-browser.tsx
damianloch: Fixed. Counts non error results instead of `names.length`.

server/routes/bitbucket/bitbucket_selection.py
damianloch: Fixed. Wrapped, `logger.exception`, returns 500 with a generic message.

server/routes/bitbucket/bitbucket_selection.py
damianloch: Fixed, and good call on scoping it. `_until_not_429` takes `attempts`; the single repo route passes 2, bulk keeps 7. Bulk behavior unchanged.

client/src/components/bitbucket-workspace-browser.tsx
damianloch: Fixed. Loop tolerates 3 consecutive poll failures, resets on success. Deadline check stays outside the try, and the task failure check stays after the loop, so neither gets swallowed.

server/routes/bitbucket/bitbucket_selection.py
damianloch: Fixed. Comment and log now mention verification state, not just the uuid.

server/connectors/bitbucket_connector/api_client.py
damianloch: Agreed, reverted. Error dict restored.

client/src/components/bitbucket-workspace-browser.tsx
damianloch: Gating is authoritative: the dispatcher skips unenrolled repos, and `_mark_webhook_verified` has `AND change_gating_enabled = TRUE` for this exact case. No review runs, badge stays off.

Also `webhook_cleanup_failed` only covers hooks with a stored uuid, so a hand made hook reports success. Toasting on it would reassure falsely in the case you mean. Filing a follow-up to detect orphans from a live `list_webhooks` check.

server/routes/bitbucket/bitbucket_selection.py
damianloch: Docstring fixed, disagreeing on the logic. We build that create request ourselves with `active=True` and both triggers, and reject error responses, so Bitbucket would have to ignore our own payload. The checks on the existing hook branch are there because someone else configured that hook. Guarding the create path would also orphan a hook we just made, since returning False skips recording the uuid.

isiddharthsingh:
the blip-tolerance, 429 budget, and pagination revert all look good 

one thing from the earlier review is still open though: bulk disable never tells the admin when webhook cleanup fails. the worker returns webhook_cleanup_failed, but the ui breaks out of the poll loop as soon as the db flags flip (bitbucket-workspace-browser.tsx ~line 158), before the job result exists — and the disable toast never reads it anyway. the single-repo toggle handles this (~line 97) and the backend comment says undeletable web-ui hooks are the common case, not an edge case.

to be clear, nothing breaks, the dispatcher drops deliveries for disabled repos as not_enrolled, so no reviews get posted. but the leftover hooks keep sending pr payloads (titles, branches, authors) to aurora forever, and the admin has no idea. someone disabling 49 repos to off-board will find stray aurora webhooks in a security review months later and wonder why we never mentioned them.

simplest fix: keep the early break for the spinners, but let a detached poll run to completion and pop the "couldn't remove some webhooks — clean them up in repository settings" warning when the result lands. same message the single-repo path already shows.

isiddharthsingh (approved):

PR #606, opened 2026-08-25
fix(bitbucket): paged file reads and edit_file for Actions

damianloch:
## Summary
- Page `get_file_contents` into verbatim ~30k-char slices so tool-output summarizers never replace large Terraform/config with `[Summarized from longer output]` mid-edit.
- Add `bitbucket_repos` `edit_file` (anchored search-and-replace, same guards as `bitbucket_fix`) and steer Actions toward it for existing files; keep `create_or_update_file` for new paths only.
- Update Bitbucket skill, generic Action guidelines (`executor.py`), stock alert-gap / HPA-VPA prompts, and remove the comment-only `.tf` example that modeled bad commits.

## Test plan
- [ ] `cd server && python -m pytest tests/chat/test_bitbucket_file_ops.py tests/chat/test_tool_output_cap.py -q`
- [ ] Action path: read a large Bitbucket file via `get_file_contents` — confirm verbatim page header / `start_line` continue, no summarization marker
- [ ] Action path: `edit_file` on an existing file → branch → PR; confirm real diff (not comments-only)
- [ ] RCA path: writes still described as forbidden; `bitbucket_fix` still available for suggestions
- [ ] PR-review path: `edit_file` rejected by the existing read-only write gate

## Summary by CodeRabbit

* **New Features**
  * Added anchored `edit_file` support for modifying existing Bitbucket files.
  * Added paginated file-content retrieval for large files.
  * Added validation to reject ineffective or empty file edits.
* **Bug Fixes**
  * Improved safeguards for read-only review workflows and prevented invalid Terraform deletion changes.
* **Documentation**
  * Clarified when to use file editing versus creating new files across supported actions.
* **Tests**
  * Added coverage for file pagination, continuation, successful edits, and rejected no-op changes.

PR #607, opened 2026-08-26
feat(bitbucket): paged reads + surgical edit_file for large files

isiddharthsingh:
Fixes two Bitbucket change-application bugs (fresh implementation of the design from #606, with review feedback incorporated):

1. **Big files were summarized before editing** — any file read over 40K chars was replaced by an LLM prose summary, so the agent couldn't anchor a real edit and fell back to trivial/comment-only changes. `get_file_contents` now returns verbatim pages (~30K escaped chars, always under the summarizer threshold) with `start_line`/`start_char` continue hints pinned to the read commit, plus a `find_in_file` action to locate lines in huge files.
2. **Whole-file rewrites instead of surgical edits** — the only write path during Actions was a full-body upload. New `edit_file` action applies anchored old/new edits server-side against the true file (shared replacer chain), commits with `parents` set to the read SHA so a moved branch tip rejects the write instead of being clobbered.

Also fixed along the way:

- **UTF-8 corruption on all Bitbucket file reads**: `/src` serves `text/plain` with no charset, so reads were decoded as ISO-8859-1 and mojibaked non-ASCII content; a subsequent write would have re-encoded the damage permanently. Undeclared charsets now default to UTF-8.
- `.json` files read as "nonexistent" (parsed-dict Content-Type path) — directory-listing detection is now shape-checked.
- `bitbucket_fix` silently replaced files when a read failed for any reason — only a confirmed 404 (with the ref verified to resolve) reaches the new-file path now; everything else is a hard error.
- The Bitbucket read-only gate now hard-enforces during background RCA, not just PR review, and `find_in_file` is regex-timeout-guarded.

Verification: 50-test suite (`tests/chat/test_bitbucket_file_ops.py`), full server suite green, plus a live smoke against Bitbucket Cloud — one `edit_file` call with 4 edits scattered across a 278K-char/10K-line file committed a byte-identical result with exactly 4 hunks, and a stale-`parents` write was rejected by Bitbucket with the file unchanged.

## Summary by CodeRabbit

* **New Features**
  * Added Bitbucket file search and anchored editing for safer, targeted repository updates.
  * Added support for navigating large files with paginated content and continuation guidance.
* **Bug Fixes**
  * Improved handling of missing files, invalid references, authentication failures, and file types.
  * Prevented stale suggestions and conflicting edits from overwriting newer repository changes.
  * Added safeguards against empty, no-op, or invalid file updates.
* **Documentation**
  * Updated repository editing guidance and Terraform deletion instructions.
  * Clarified read-only behavior and file-editing recommendations across automated actions.

### DIFF
--- a/server/chat/backend/agent/tools/bitbucket/repos_tool.py
+++ b/server/chat/backend/agent/tools/bitbucket/repos_tool.py
@@ -4,7 +4,16 @@ import json
 import logging
 import re
 import threading
+import time
 from typing import Literal, Optional
+
+try:
+    # Supports a true per-call timeout for the find_in_file regex fallback.
+    # Transitively present via tiktoken; the daemon-thread deadline below
+    # covers its absence.
+    import regex as _regex_mod
+except ImportError:  # pragma: no cover — regex ships with tiktoken
+    _regex_mod = None
 
 from pydantic import BaseModel, Field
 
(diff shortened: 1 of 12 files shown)


server/chat/backend/agent/tools/bitbucket/repos_tool.py
```diff
@@ -38,10 +55,144 @@ class BitbucketReposArgs(BaseModel):
...
+def _regex_scan_with_module_timeout(lines: list, query: str) -> tuple[list[str], int]:
+    """Regex fallback scan using the ``regex`` module's native per-call
+    timeout — a genuinely cancellable boundary: the engine checks the
+    deadline DURING matching, so a catastrophic pattern stops burning CPU
+    at the deadline instead of running to completion in an abandoned
+    thread."""
+    try:
+        pattern = _regex_mod.compile(query)
+    except _regex_mod.error:
+        return [], 0  # not a regex either — genuinely no matches
```
damianloch: not super big deal but this makews compilation failure and look like no matches. Its more like "model provided invalid arguments" rather than "model provided valid arguments but no matches". If there is an easy way to distinguish, we should

server/chat/backend/agent/tools/bitbucket/utils.py
```diff
@@ -11,11 +12,33 @@
...
 logger = logging.getLogger(__name__)
 
 DIFF_TRUNCATE_LIMIT = 50_000
 
+# Budget for one page of file content, measured on the JSON-ESCAPED form of
+# the text (what actually lands in the serialized tool output). Derived from
+# tool_output_cap.PASS_THROUGH_CHARS with a 10K margin for the JSON envelope,
+# so a file read is returned verbatim and never replaced by an LLM summary —
+# even if the pass-through threshold is later lowered.
+PAGE_CONTENT_BUDGET = min(30_000, _PASS_THROUGH_CHARS - 10_000)
```
damianloch: clamp with max(1, min(30_000, PASS_THROUGH_CHARS - 10_000)) so a future cap change cannot produce a non-positive page budget

server/chat/backend/agent/tools/cloud_tools.py
```diff
@@ -2332,19 +2366,22 @@ def _pinned_trigger(action_id: str = "", _pid=pinned_id, _fn=final_func, **kw):
             )
 
             _is_rca_background = _is_background_rca(state_context, is_background)
+            _bb_read_only = _is_rca_background or is_pr_review
+            _ro_reason = "PR review" if is_pr_review else "RCA"
 
             # PR change-gating reviews are strictly read-only: the review
```
damianloch: update the comment to say PR review and background RCA are both hard read-only on Bitbucket writes.

damianloch (changes requested):

server/chat/backend/agent/tools/bitbucket/repos_tool.py
isiddharthsingh: fixed. if the query has no literal hits and also doesn't compile as a regex, it now raises back a proper error with the actual compile message instead of a silent zero-match, so the model knows to fix the pattern rather than assume the code isn't there. added a test for both cases (broken pattern that exists literally still matches fine)

server/chat/backend/agent/tools/bitbucket/utils.py
isiddharthsingh: done, clamped exactly as you suggested — max(1, min(30_000, PASS_THROUGH_CHARS - 10_000)) so a future cap change can't push the budget non-positive

server/chat/backend/agent/tools/cloud_tools.py
isiddharthsingh: updated. comment now says PR review and background RCA are both hard read-only on bitbucket writes, and points at the function-level gate that enforces it in both modes

damianloch (approved):

PR #611, opened 2026-09-03
Root-cause dedup layer 2: recurrence grouping and UI

isiddharthsingh:
## Summary

Layer 2 of root-cause dedup. Layer 1 (#610) writes `incidents.recurrence_of_incident_id` when the correlation agent folds a recurrence into its anchor; nothing read it yet, so `/incidents` still showed one card per incident. This PR turns the pointer into one expandable card per group and surfaces the relationship on the detail page. Layer 3 (Slack threading) is out of scope.

## API

- `_format_incident_response` always emits `recurrenceOf` (plus `recurrenceOfTitle` when set). List and detail share one `_INCIDENT_SELECT`.
- `GET /api/incidents?groups=1` pages over recurrence groups: `limit`/`offset`/`status` select distinct roots (newest fire first) and `total` counts groups, then every non-merged anchor and member of those groups is returned as a flat list, capped at 1000 rows with anchors first and the largest group trimmed first. Each row carries `occurrenceTotal` (full group size) so the client can show how many members were cut. `aurora_summary` and `alert_metadata` are NULL in this mode. Without `groups` the existing query path is unchanged, so MCP paging is unaffected.
- Anchor detail gets `occurrences[{id, alertTitle, status, startedAt, alertFiredAt}]` (newest 200, emitted oldest-first) plus `occurrencesTotal`; members get `recurrenceOf` + `recurrenceOfTitle` and stay fully GET-able.

## UI

- `/incidents` groups via a pure helper (`incident-grouping.ts`): anchor + members become one card showing `fired N times` and time since the last fire; the chevron expands a chronological occurrence list without navigating. Standalones keep the existing row. Orphans (anchor not in the fetch) are never hidden.
- Expand state lives at page level keyed by anchor id, so SSE refetches (`incident_update`, `recurrence_folded`) never collapse an open card. Refetches are coalesced.
- Member detail shows an "Occurrence of …" banner linking to the anchor; its own investigation stays visible. Anchor detail gets an expandable "Fired N times" section listing the members.
- `ExpandablePanel` extracts the expand animation; `CorrelatedAlertsSection` now uses it and no longer lists the fold's `recurrence` rows (the occurrences section covers them).
- Detail page subscribes to the incident stream so a fold that lands after the RCA finishes updates an open page.

## Verification

- Server suite: 1372 passed; the 3 `test_rls_cross_tenant` failures are the known dev-container Casbin denials.
- API smoke against the dev group: `groups=1&limit=1` returns the whole 4-row group, `limit=1` alone returns 1 row, anchor detail has 3 chronological occurrences, member detail has pointer + title and no `occurrences` key.
- 14-case check of `groupIncidents` (orphans, depth-2 chains, `alertFiredAt` fallback, sections, sort, stalled).
- Chrome: group card expands/collapses without navigating, refetch keeps it open, member banner links to the anchor, "Fired 4 times" expands to 3, Correlated Alerts still renders. No console errors.

## Follow-ups (not this PR)

- `recurrence_unfolded` is recorded as a lifecycle event but never broadcast over SSE.
- Depth-1 pointer invariant is assumed here; `recurrence_fold._check_locked_pair` does not yet reject a child that already has dependents.

## Summary by CodeRabbit

- **New Features**
  - Incidents are grouped by recurrence, with occurrence counts, timing, status, and investigation state.
  - Expandable groups and incident cards show individual occurrences, links, and notices for older occurrences not loaded.
  - Incident details identify recurring alerts and link to the original incident.

- **Improvements**
  - Incident and recurrence updates refresh automatically.
  - Expandable sections provide smoother interactions.
  - Time and duration displays no longer show negative values.
  - Newer refresh results take precedence over older responses.

### DIFF
--- a/client/src/app/incidents/components/IncidentGroupRow.tsx
+++ b/client/src/app/incidents/components/IncidentGroupRow.tsx
@@ -122,7 +122,7 @@ export default function IncidentGroupRow({ group, expanded, onToggle }: Readonly
         </div>
 
         <ExpandablePanel open={expanded} className="mx-4" contentClassName="py-3 border-t border-border space-y-1">
-          {group.occurrences.slice(0, anchorIndex + 1).map(renderOccurrence)}
+          {group.occurrences.slice(0, anchorIndex + 1).map((occurrence, index) => renderOccurrence(occurrence, index))}
           {group.notLoaded > 0 && (
             <Link
               href={`/incidents/${anchor.id}`}
--- a/client/src/app/incidents/components/IncidentGroupRow.tsx
+++ b/client/src/app/incidents/components/IncidentGroupRow.tsx
@@ -45,6 +45,25 @@ export default function IncidentGroupRow({ group, expanded, onToggle }: Readonly
   // Severity is typed as a closed union but the API can still send 'unknown'.
   const severity: string = anchor.alert.severity;
   const showSeverity = (severity && severity !== 'unknown') || anchor.status === 'analyzed';
+  // Members the server cut are the oldest, so they sit between the anchor and
+  // the first loaded member: ordinals after the anchor skip them, and the
+  // "not shown" row is rendered in that gap.
+  const anchorIndex = group.occurrences.findIndex(o => o.id === anchor.id);
+  const renderOccurrence = (occurrence: Incident, index: number) => (
+    <Link
+      key={occurrence.id}
+      href={`/incidents/${occurrence.id}`}
+      className="flex items-center gap-3 px-2 py-1.5 rounded-md text-sm hover:bg-muted/60 transition-colors"
+    >
+      <span className="w-32 shrink-0 tabular-nums text-muted-foreground whitespace-nowrap">{formatFireTime(occurrence)}</span>
+      <span className="flex-1 min-w-0 truncate">
+        occurrence {index + 1 + (index > anchorIndex ? group.notLoaded : 0)}
+        {index === 0 && <span className="text-muted-foreground"> · first</span>}
+      </span>
+      <OccurrenceStatusIcon incident={occurrence} now={now} />
+      <ChevronRight className="h-4 w-4 text-muted-foreground" />
+    </Link>
+  );
 
   return (
     <Card className={`hover:border-primary/50 transition-colors ${isActive ? 'border-l-4 border-l-muted-foreground' : ''}`}>
@@ -70,6 +89,9 @@ export default function IncidentGroupRow({ group, expanded, onToggle }: Readonly
                   <Repeat className="h-3 w-3" />
                   fired {group.occurrenceCount} times
                 </span>
+                {/* Time since the newest fire. No "N related" here: correlatedAlertCount is the
+                    old correlator's counter and folds bump it, so it would double as a second
+                    occurrence count (design doc: do not show it on the group header). */}
                 <span className="flex items-center gap-1">
                   <Clock className="h-3 w-3" />
                   {incidentsService.formatDuration(new Date(group.lastFiredAt).toISOString())}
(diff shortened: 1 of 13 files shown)


server/routes/incidents_routes.py
```diff
@@ -267,57 +323,96 @@ def get_incidents(user_id):
...
+                    # The writer (recurrence_fold) keeps pointers depth-1 —
+                    # only that depth is completed here. Anchors sort first so
+                    # the _GROUPS_MAX_ROWS cut drops the oldest members, never
+                    # the group root. group_size is a window count, evaluated
+                    # before the LIMIT, so each row carries the full non-merged
+                    # size of its group and the client can tell when members
+                    # were cut.
+                    query = (
+                        "WITH roots AS ("
+                        "SELECT COALESCE(i.recurrence_of_incident_id, i.id) AS root_id FROM incidents i"
```
beng360: the roots CTE limits incidents, not groups, so one flapping group eats the whole page. tested it live on this branch: seeded a 121-member group plus 4 unrelated incidents, `?groups=1` with the default limit came back with 126 rows but only 3 distinct roots — the two older standalone incidents just vanish from /incidents, and there's no pagination in the UI to reach them. the page should pick distinct roots instead, something like `SELECT COALESCE(i.recurrence_of_incident_id, i.id) AS root_id, MAX(i.started_at) AS last_started ... GROUP BY 1 ORDER BY 2 DESC LIMIT %s`.

server/routes/incidents_routes.py
```diff
@@ -636,6 +720,27 @@ def get_incident(user_id, incident_id: str):
                     )
                 incident["correlatedAlerts"] = correlated_alerts
 
+                # Later occurrences folded into this incident (root-cause dedup).
+                # Only anchors carry the list; members link back via recurrenceOf.
+                if incident["recurrenceOf"] is None:
+                    cursor.execute(
+                        """SELECT id, alert_title, status, started_at, alert_fired_at
+                           FROM incidents
+                           WHERE recurrence_of_incident_id = %s AND org_id = %s
```
beng360, line 755: no LIMIT here. the list caps completion at _GROUPS_MAX_ROWS because a flapping alert can fold thousands of members, but the "N older occurrences not shown · open the incident" row in IncidentGroupRow sends people to this endpoint, where the same group comes back unbounded and OccurrencesSection renders every row as a Link. cap it (newest first) and return the total so the detail page can say how many were cut.

client/src/app/incidents/components/IncidentGroupRow.tsx
```diff
@@ -0,0 +1,133 @@
...
+              </Badge>
+            )}
+
+            <div className="flex-1 min-w-0">
+              <p className="font-medium truncate">{anchor.alert.title}</p>
+              <div className="flex items-center gap-3 text-sm text-muted-foreground mt-0.5">
+                {anchor.alert.service !== 'unknown' && <span>{anchor.alert.service}</span>}
+                <span className="flex items-center gap-1">
+                  <Repeat className="h-3 w-3" />
+                  fired {group.occurrenceCount} times
```
beng360, line 117: the group card drops the `N related` chip that IncidentRow shows for correlatedAlertCount, so once an incident becomes an anchor its correlated alerts stop being visible from the list. intentional?

beng360 (changes requested):

server/routes/incidents_routes.py
isiddharthsingh: fixed. the page now selects distinct roots (group by coalesce(recurrence_of_incident_id, id), newest fire first) and total counts groups. reproduced your 121-member case, all three roots come back and the standalones stay on page one.

server/routes/incidents_routes.py
isiddharthsingh: yep. capped at 200 newest, returned oldest-first, plus occurrencesTotal so the section shows the real count and "k older occurrences not shown" when cut.

client/src/app/incidents/components/IncidentGroupRow.tsx
isiddharthsingh: intentional, the design doc says not to show it on the group header. that chip is the old correlator's counter and folds bump it, so it'd just be a second occurrence count.

beng360 (approved):

PR #612, opened 2026-09-04
fix(repo-metadata): Gemini thinking-mode content breaks Bitbucket/GitLab repo descriptions

Noahcasarotto:
## What

`server/utils/repo_metadata.py` called `response.content.strip()` on the LLM reply. On Gemini 2.5/3.x (google and vertex providers) thinking is on by default and `content` is a list of thinking/text blocks, so every Bitbucket and GitLab repo description crashed with `AttributeError: 'list' object has no attribute 'strip'`, retried twice, and ended in `metadata_status='error'`. This is what Bombora hit on chart 1.2.19.

The fix routes the reply through the existing `_extract_text_from_response` helper, which the GitHub metadata route has used from day one. It drops thinking blocks and passes string content through unchanged. One line of behaviour change plus a comment.

## Verified

- Reproduced on the prebuilt `aurora-server:1.2.19` image with `MAIN_MODEL=gemini-3.5-flash` (thinking on): three Bitbucket repos stuck in `error`.
- Hot-patched the worker with this change, requeued the same three repos on the same model: all three `ready` with real summaries in 26 s, thinking mode still enabled.
- `tests/tasks/test_repo_metadata_summary.py` covers list content, plain string content, and empty or thinking-only content. Passes in the CI minimal-deps environment.

Refs DEV-1489

https://claude.ai/code/session_01JvJJuCfb8DortPF5Xiqzmj

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of AI responses that include internal reasoning, ensuring readable text is used across repository summaries, incident summaries, recommendations, task analysis, notifications, and streamed messages.
  - Summaries continue to remove surrounding whitespace and fall back to “No summary generated” when no usable text is available.

- **Tests**
  - Added coverage for thinking-mode responses, plain text, structured content, empty content, and responses containing only internal reasoning.

### DIFF
--- a/server/utils/repo_metadata.py
+++ b/server/utils/repo_metadata.py
@@ -220,6 +220,7 @@ def _generate_summary(user_id: str, context: str) -> str:
     from chat.backend.agent.utils.llm_usage_tracker import tracked_invoke
     from langchain_core.messages import HumanMessage
     from utils.hooks import get_hook
+    from chat.background.summarization import _extract_text_from_response
 
     # Hook: check if LLM call is allowed
     from utils.auth.stateless_auth import get_org_id_for_user
@@ -241,7 +242,10 @@ def _generate_summary(user_id: str, context: str) -> str:
         model_name=ModelConfig.INCIDENT_REPORT_SUMMARIZATION_MODEL,
         request_type="repo_metadata",
     )
-    return response.content.strip() if response.content else "No summary generated"
+    # Gemini thinking models return content as a list of thinking/text blocks;
+    # the helper drops the thinking blocks and returns the text (str content passes through).
+    summary = _extract_text_from_response(response.content) if response.content else ""
+    return summary or "No summary generated"
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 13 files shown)


OlivierTrudeau:
Fix looks good for the repo-metadata path

There are 4 more spots in `server/chat/background/recommender.py` with basically the same bug. They use `str(response.content)` instead of `.content.strip()`:

- L291 `_infer_investigation_state` — `str(response.content).strip()` then `json.loads`
- L555 `_self_execute_safe_diagnostics` — `_parse_json_list_response(str(response.content))`
- L933 `_enrich_validated_fixes` — `_apply_enrichment_response(str(response.content).strip(), ...)`
- L996 `_generate_summaries` — `str(response.content).strip()`

It is beyond the scope of this exact PR but will potentially break in the future and if we can catch it before it affects someone that would be great

server/utils/repo_metadata.py
```diff
@@ -220,6 +220,7 @@ def _generate_summary(user_id: str, context: str) -> str:
     from chat.backend.agent.utils.llm_usage_tracker import tracked_invoke
     from langchain_core.messages import HumanMessage
     from utils.hooks import get_hook
+    from chat.background.summarization import _extract_text_from_response
 
     # Hook: check if LLM call is allowed
```
OlivierTrudeau, line 225: There's like 4 different versions of this method. We should consolidate them. Doesn't mean this PR won't work but worth doing to avoid trouble in the future

OlivierTrudeau (approved):

isiddharthsingh (approved):

PR #615, opened 2026-09-08
Root-cause dedup layer 3: Slack thread consolidation

isiddharthsingh:
## Summary

Root-cause dedup layer 3: one Slack channel message per root cause. Layers 1 (#610) and 2 (#611) write and display `incidents.recurrence_of_incident_id`; this PR makes Slack notifications follow it.

- "Analysis Complete" / "Investigation Failed" now post as thread replies under the incident's own "Investigation Started" message instead of a second top-level card.
- A folded recurrence posts a compact ":repeat: Still firing — occurrence N of M" reply under the anchor's message (button to the occurrence, context link to the anchor) and retires its own Started message; its `slack_message_ts` is NULLed so later posts go top-level again.
- Every failure path degrades to today's top-level full card. Slack silently posts top-level when the parent ts is missing or deleted (no error), so placement is verified from the response's `message.thread_ts`; a stray compact post is removed and a stale anchor ts is cleared.
- Standalone incidents with no Started message (start notifications off) get the Complete ts backfilled with a guarded compare-and-set so later recurrences can still thread.
- Dispatcher incident data carries `recurrence_of`, `anchor_slack_message_ts`, `anchor_alert_title`, `occurrence_number`, `group_size` (group membership matches `_group_is_stale` and the layer-2 groups query; ordering matches the UI).
- Follow-up chats on an existing incident no longer re-announce completion (`send_notifications` now also gates summarization).
- Inbound thread @mentions resolve org-wide (RLS-scoped) since a recurrence reply lives in the anchor's thread; the RCA session is reused only for its owner.
- Slack client: `SlackAPIError` (ok=false) distinct from transport errors so only thread-specific rejections retry top-level; `get_message` via `conversations.replies`.

No schema, env, frontend, or Google Chat changes (Google Chat parity deferred).

## Verification

- `tests/utils/notifications`: 77 tests (pure, no DB/network). Full server suite: 1450 passed; the 3 failures in `tests/auth/test_rls_cross_tenant.py` are pre-existing and identical on `main`.
- Live on the dev stack against a real workspace: standalone Complete threaded under its Started; a folded child posted the compact reply under the anchor, its Started message deleted and ts NULLed; anchor ts NULL → full card top-level; bogus/deleted anchor ts → stray removed, full card top-level; inbound `get_message` on live, deleted, and bogus ts.

## Follow-ups (not in this PR)

- Detection agent keeps claiming the first-ever occurrence even after that group is stale (`stale_group` reject), so nothing re-anchors on the fresh incident; needs a layer-1 prompt/fold change.
- `reply_broadcast` on the full cards (thread replies notify nobody; the Started card stays "In Progress").
- Persist `{ts, channel}` for the Started message so a channel switch is not read as "message gone".
- Google Chat parity; docs page for notification behaviour.

## Summary by CodeRabbit

- **New Features**
  - Slack incident notifications now consolidate into threads, including recurrence updates and completed or failed investigation messages.
  - Notifications fall back to top-level posts when threads or parent messages are unavailable.
  - Incident searches from Slack can identify incidents across the organization while protecting session access.
  - Recurrence analysis now considers recent related incidents.
  - Background summaries can optionally complete without sending duplicate notifications.
  - Existing Google Chat cards can be refreshed without creating new messages.

- **Bug Fixes**
  - User-provided incident details are escaped before appearing in Slack messages.

### DIFF
--- a/server/chat/background/summarization.py
+++ b/server/chat/background/summarization.py
@@ -718,15 +718,25 @@ def generate_incident_summary_from_chat(
     incident_id: str,
     user_id: str,
     session_id: str,
-    send_notifications: bool = True,
+    announce_completion: bool = True,
 ) -> Dict[str, Any]:
     """Regenerate incident summary after RCA using the RCA chat transcript with citations.
 
-    send_notifications: False for follow-up chats (Slack/Google Chat replies on
-    an existing incident) so re-summarizing does not re-announce completion —
-    for a folded recurrence that would post another "Still firing" reply.
+    announce_completion: False when the summary is being regenerated after a
+    follow-up chat (Slack/Google Chat reply on an existing incident). The
+    summary is still written; the "investigation completed" announcement
+    (email, Slack card) is skipped since the investigation already completed
+    once — for a folded recurrence it would otherwise post another "Still
+    firing" reply — and only the Google Chat card, which is edited in place,
+    is refreshed to match the new summary.
     """
     from celery.exceptions import SoftTimeLimitExceeded
+
+    def _notify_completed() -> None:
+        from utils.notifications.dispatcher import notify_investigation_completed
+        if not announce_completion:
+            logger.info(f"{_LOG_PREFIX} Follow-up summary for {incident_id}; completion already announced, refreshing cards only")
+        notify_investigation_completed(user_id, incident_id, session_id=session_id, refresh_only=not announce_completion)
 
     logger.info(
         f"{_LOG_PREFIX} Regenerating summary from chat for incident {incident_id} (user={user_id}, session={session_id})"
@@ -899,11 +909,7 @@ def generate_incident_summary_from_chat(
             )
 
         # Send completion notifications via centralized dispatcher
-        if send_notifications:
-            from utils.notifications.dispatcher import notify_investigation_completed
-            notify_investigation_completed(user_id, incident_id, session_id=session_id)
-        else:
-            logger.info(f"{_LOG_PREFIX} Follow-up summary for {incident_id}; completion notifications skipped")
+        _notify_completed()
 
         return {
             "incident_id": incident_id,
@@ -922,9 +928,7 @@ def generate_incident_summary_from_chat(
             # overwrite the good summary with an error — send the completion
             # notification the task guarantees and finish.
             try:
-                if send_notifications:
-                    from utils.notifications.dispatcher import notify_investigation_completed
-                    notify_investigation_completed(user_id, incident_id, session_id=session_id)
+                _notify_completed()
             except Exception:
                 logger.exception(
                     f"{_LOG_PREFIX} Failed to notify after soft-limit for {incident_id}"
--- a/server/chat/background/summarization.py
+++ b/server/chat/background/summarization.py
@@ -718,8 +718,14 @@ def generate_incident_summary_from_chat(
     incident_id: str,
     user_id: str,
     session_id: str,
+    send_notifications: bool = True,
 ) -> Dict[str, Any]:
-    """Regenerate incident summary after RCA using the RCA chat transcript with citations."""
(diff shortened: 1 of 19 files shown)


server/chat/background/summarization.py
```diff
@@ -893,8 +899,11 @@ def generate_incident_summary_from_chat(
             )
 
         # Send completion notifications via centralized dispatcher
-        from utils.notifications.dispatcher import notify_investigation_completed
-        notify_investigation_completed(user_id, incident_id, session_id=session_id)
+        if send_notifications:
+            from utils.notifications.dispatcher import notify_investigation_completed
```
OlivierTrudeau: I find this is a bit misleading. send_notificatio false makes it seem like we aren't sending a slack notification, but from my understanding, it's more like don't create a new card in Slack right? Like if it's a chat reply don't create a investigation_completed card

server/routes/slack/slack_events_helpers.py
```diff
@@ -544,18 +549,19 @@ def get_incident_by_slack_message(user_id: str, slack_message_ts: str):
                 )
                 cursor.execute(
                     """
-                    SELECT id, aurora_chat_session_id
+                    SELECT id, aurora_chat_session_id, user_id
                     FROM incidents
-                    WHERE user_id = %s AND slack_message_ts = %s
+                    WHERE slack_message_ts = %s
                     LIMIT 1
```
OlivierTrudeau, line 558: Can't we use the org_id? because ts even if there's only a slight chance that 2 messages have the same ts, if they do, you could have cross org contamination

server/connectors/slack_connector/client.py
```diff
@@ -98,8 +113,26 @@ def update_message(self, channel: str, ts: str, text: str, blocks: Optional[List
         return self._make_request("POST", "chat.update", data)
 
     def delete_message(self, channel: str, ts: str) -> None:
-        """Delete a message from a Slack channel. Raises ValueError on failure."""
-        self._make_request("POST", "chat.delete", {"channel": channel, "ts": ts})
+        """Delete a message from a Slack channel. Raises ValueError on failure.
```
OlivierTrudeau, line 123: Should use this or edit message or something to replace the investigation started message in slack with the investigation completed. This would make it cleaner instead of replying in the thread which is good start but shouldn't be the final state

OlivierTrudeau (changes requested):
Two more issues. 

1- the Correlator which provides the hint for relating old RCAs and incidents works with the alert correlator logic. This logic only checks the for incidents that are still beign investigated, not analyzed or resolved. Even if we do make the window larger for the correlator, the scores are heavily time dependent and so an incident done 15 min will not get a high correlation score. 

2- The list incident tool which is used by the LLM to determine the previous incidents and determine if it's a reoccuring incident can input the parameters of it's choice for the status. In my testing, the LLM only put resolved incidents. however, my incidents were analyzed status so it didn't get back a single incident and was unable to determine whether the current notification was related to a previous incident or not

server/connectors/slack_connector/client.py
isiddharthsingh: Done. The Started card is now updated in place into the Complete or Failed card via chat.update, so the top-level message always shows the outcome and the thread stays for recurrences. Recurrences remain thread replies, and the anchor card gets a "Still firing — N occurrences, last <time>" footer on every fold. A child card kept because people replied in it becomes a "Folded into <anchor>" stub. A Failed card never overwrites a Complete card.

server/routes/slack/slack_events_helpers.py
isiddharthsingh: Done. The lookup resolves the caller's org first and adds AND org_id = %s on top of RLS. Also fixed the sibling get_session_from_thread lookup, whose RLS SET was a Postgres syntax error and always fell through.

OlivierTrudeau (approved):

PR #621, opened 2026-09-09
Recurrence detection via a memory-backed Incident Index

OlivierTrudeau:
## What & why

Replaces the interim embedding-loop RCA search (`_search_similar_rcas_impl`) with a **self-maintaining Incident Index** — a compact, id-keyed memory artifact the recurrence agent scans to find candidate anchors, then drills into via `get_incident`.

The interim port re-embedded the query ~200× and every candidate on each check (~400 embedding calls) and grew with table size. Full `aurora_summary`s (avg ~365 tokens, p95 ~4.5k) don't scale in-prompt. The index is one bounded `read_memory` — no embeddings, no growth with history.

## How it works
- **Storage:** a normal `artifact` memory entry titled "Incident Index" (one per org). Visible read-only in the memory UI, discoverable by agents via `list_memories`.
- **Write (deterministic + fold-aware):** in `summarization.py`, after the recurrence check:
  - *new incident* → generate a **cause-focused one-line synopsis** (one cheap dedicated LLM call, falls back to alert title on failure) and append a fresh root line.
  - *folded (recurrence)* → roll the incident up under the ROOT's existing line (`↳ recurrences: … (N total, last <date>)`), so no sibling line the next check could wrongly fold into. Idempotent; best-effort; never blocks notify.
- **Retrieve:** the index is injected into the recurrence agent's input block via `read_index()` (hard-capped at ~24k chars). The existing recent-incidents section (last `GROUP_IDLE_HOURS`, cap 15) stays as the bounded cold-start fallback + joinability guard.
- **Groom:** the nightly `memory_consolidation` action clusters recurrences, marks solved, trims to a ~150-line cap, and preserves the `INC <id>` join key. Only the index is reshaped aggressively; other artifacts get formatting fixes only.
- **Fold + Slack threading: unchanged** — still DB-keyed on `recurrence_of_incident_id` → anchor `slack_message_ts`. The index is a discovery map only.

## Review asks (for the original Weaviate-feature author)
1. Does dropping semantic similarity for an LLM-scanned compact index hold up on your recall expectations vs. the old Weaviate store?
2. Is the cause-focused synopsis (one LLM call at RCA completion) + nightly LLM-grooming split the right model, or would you push more onto grooming?
3. Grooming cadence/cap (~150 lines, nightly) — reasonable for your incident volume?
4. Two accepted behaviors to sanity-check: synopsis is generated only on the new-root path (folded incidents just bump the roll-up count); and in **shadow** mode nothing folds, so every incident becomes a new root line until grooming clusters them.

## Tests
- `tests/services/memory/test_incident_index.py`: synopsis (fallback + LLM success), formatting, idempotent append, bounded read, recurrence roll-up (create/extend/idempotent/fallback).
- `test_recurrence_candidates.py` / `test_recurrence_agent.py`: index section + context wiring + fold-outcome return.
- Deleted the obsolete `test_recurrence_rca_search.py`. Memory + correlation + architectural suites green.

## Summary by CodeRabbit

- **New Features**
  - Added a system-managed “Incident Index” to organize incidents, track recurring issues, and preserve root-cause summaries.
  - Recurring incidents are now grouped under existing root incidents, while new incidents are added to the index automatically.
  - Incident Index maintenance can consolidate solved entries and limit index size.

- **Bug Fixes**
  - System-managed memory entries are now protected from accidental editing, recategorization, or deletion.
  - Invalid attempts to modify protected or missing entries now return clear errors.


isiddharthsingh:
@coderabbitai review

server/services/memory/incident_index.py
```diff
@@ -0,0 +1,369 @@
...
+
+# One synopsis line must stay short so 100s of incidents fit the recurrence
+# agent's input block. Titles/summaries are trimmed to this before formatting.
+_MAX_SYNOPSIS_CHARS = 160
+
+# Hard ceiling on how much of the index we ever inject into a prompt. The
+# groomer keeps it well under this; the cap is belt-and-braces so an un-groomed
+# index (e.g. collector ran but consolidation hasn't) can't blow the context
+# window. ~24k chars ≈ 6k tokens.
+INDEX_INJECTION_CHAR_BUDGET = 24000
```
isiddharthsingh: The "groomer keeps it well under this" comment doesn't hold against the groomer's own target. system_actions.py:57 caps at ~150 lines, but a line is `- [INC <36-char uuid> | date | svc | status] ` (~85 chars) plus a synopsis the prompt aims at ~20 words and _clip allows up to 160, so ~200-250 chars each. 150 lines is ~30k+ chars before any `↳ recurrences:` continuations, which is over the 24k budget. So read_index at :355-360 will drop the oldest ~25% on every check even on a freshly groomed index, and _incident_index_lines (recurrence_agent.py:228-252) gives the agent no "N older lines not shown" marker, unlike _recent_incidents_lines right below it. Either lower the groom cap to ~100 lines or raise the budget to ~40k, and add a truncation line the way the recent-incidents section does.

server/services/memory/incident_index.py
```diff
@@ -0,0 +1,369 @@
...
+        return ""
+    # Keep the most recent lines: appends go to the end, so trim from the
+    # front when over budget (grooming normally keeps this well under).
+    if len(content) > INDEX_INJECTION_CHAR_BUDGET:
+        content = content[-INDEX_INJECTION_CHAR_BUDGET:]
+        # Drop a partial leading line after trimming.
+        nl = content.find("\n")
+        if nl != -1:
+            content = content[nl + 1:]
+    return content
```
isiddharthsingh: "Trim from the front = drop oldest" is only true while the index is append-only. The groomer prompt tells the LLM to cluster and reorder by frequency and never asks it to preserve chronological order, so after the first nightly groom this can drop arbitrary lines, and can also cut between a root line and its `↳ recurrences:` continuation, leaving an orphan continuation at the top of the prompt. Worth either telling the groomer to keep newest-last, or trimming by whole root+continuation blocks.

server/services/memory/__init__.py
```diff
@@ -1,7 +1,18 @@
...
-AGENT_CATEGORIES = ("artifact",)
+# The "artifact" category is system-maintained: Aurora writes and grooms these
+# entries (e.g. the Incident Index) itself. Users can view them but must not
+# create/edit/recategorize/delete them, or the id-keyed lookups in
+# services/memory break. Every other category is user-writable.
+SYSTEM_CATEGORY = "artifact"
+USER_WRITABLE_CATEGORIES = tuple(c for c in MEMORY_CATEGORIES if c != SYSTEM_CATEGORY)
+
+# Agent-only categories — not exposed via user-facing memory routes.
+AGENT_CATEGORIES = ()
```
isiddharthsingh, line 12: Two things here. First, this comment now describes an empty tuple and should go. Second, moving "artifact" into MEMORY_CATEGORIES makes _CATEGORY_HELP (memory_tool.py:27) advertise it to the chat agent, and write_memory / append_to_memory / edit_memory (memory_tool.py:238, :332, :432) only run _validate_category, with no SYSTEM_CATEGORY guard. The REST routes now 403 on artifact, but any RCA or interactive chat turn can still write_memory(category="artifact", title="Incident Index", overwrite=true) and wipe the org's recurrence map, which incident_index.py parses by substring. Consolidation needs write access, but a general chat does not. Suggest gating writes to SYSTEM_CATEGORY behind an explicit flag that only the memory_consolidation action sets.

server/services/memory/incident_index.py
```diff
@@ -0,0 +1,369 @@
...
+            streaming=False,
+        )
+        response = tracked_invoke(
+            llm,
+            [HumanMessage(content=prompt)],
+            user_id=user_id,
+            session_id=session_id,
+            model_name=ModelConfig.INCIDENT_REPORT_SUMMARIZATION_MODEL,
+            request_type="incident_index_synopsis",
+        )
```
isiddharthsingh, line 133: create_chat_model here has no timeout, and this call sits in the pre-notify path after the recurrence check has possibly already consumed its 120s. The block's except Exception in summarization.py:914 also swallows SoftTimeLimitExceeded, so if the soft limit fires inside this call we continue toward notify with only the hard 420s limit left. Low likelihood since the call is tiny, but a timeout= on create_chat_model closes it for free.

server/chat/background/summarization.py
```diff
@@ -870,6 +871,51 @@ def _notify_completed() -> None:
...
+                    user_id=user_id,
+                    incident_id=incident_id,
+                    date_iso=date_iso,
+                    service=basics.get("service") or "",
+                    status="resolved",
+                    alert_title=basics.get("alert_title") or "",
+                    summary=summary or "",
+                    synopsis=synopsis,
+                    session_id=session_id,
+                )
```
isiddharthsingh: record_recurrence's bool return is dropped here, and append_incident_line's below it. Both functions deliberately return False on no_match / error so the caller can react; at minimum log at warning when either returns False so a lost roll-up (see the atomicity thread) shows up in logs instead of vanishing.

server/services/correlation/recurrence_agent.py
```diff
@@ -652,30 +652,34 @@ def run_recurrence_check(
...
-            return
+            return None
 
         if get_existing_verdict(incident_id, user_id, decision_point):
             logger.info(
                 "%s Verdict already exists for incident %s (%s); skipping re-run",
                 _LOG_PREFIX, incident_id, decision_point,
             )
-            return
+            return None
```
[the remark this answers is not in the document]
OlivierTrudeau: Good catch. Fixed so an existing verdict now returns the persisted fold outcome instead of None, so a retry rolls up under the root instead of adding a duplicate line.

server/services/memory/incident_index.py
```diff
@@ -0,0 +1,369 @@
...
+    service: str,
+    status: str,
+    synopsis: str,
+) -> str:
+    """Canonical one-line index entry. The incident_id is the join key back to
+    the DB (get_incident / recurrence_fold), so it MUST be present and first."""
+    svc = (service or "unknown").strip() or "unknown"
+    st = (status or "resolved").strip() or "resolved"
+    date = (date_iso or "").strip() or "?"
+    return f"- [INC {incident_id} | {date} | {svc} | {st}] {synopsis}"
```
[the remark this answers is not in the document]
OlivierTrudeau: Done. format_index_line now collapses interior whitespace so a newline in the service name can't split the entry.

server/services/memory/incident_index.py
```diff
@@ -0,0 +1,369 @@
...
+            date_iso=date_iso,
+            service=service,
+            status=status,
+            synopsis=line_synopsis,
+        )
+
+        # Idempotency: don't re-append if this incident_id is already indexed.
+        # Read is cheap (single artifact) and avoids duplicate lines on retry.
+        existing = get_memory_content(user_id, INCIDENT_INDEX_CATEGORY, INCIDENT_INDEX_TITLE)
+        if existing and f"INC {incident_id}" in existing:
```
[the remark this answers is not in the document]
OlivierTrudeau: The tool writes already hold FOR UPDATE, so the torn write is covered. Added a retry on no_match in record_recurrence to handle a stale read from a concurrent groom.

server/services/memory/incident_index.py
OlivierTrudeau: You're right, the numbers didn't line up. Bumped the budget to 40k so a groomed index fits, and added a truncation marker when we do trim.

server/services/memory/incident_index.py
OlivierTrudeau: Fixed. read_index now trims by whole root+continuation blocks so it can't orphan a recurrences line, plus the marker so the agent knows it's partial.

server/services/memory/__init__.py
OlivierTrudeau: Cleaned up the stale comment. We're ok with a non groomer agent touching the index for now, so leaving the write path open.

server/services/memory/incident_index.py
OlivierTrudeau: Added a 30s timeout on the call. Cheap insurance, thanks.

server/chat/background/summarization.py
OlivierTrudeau: Good point. Both returns are checked now and log a warning on failure so a lost roll-up shows up in the logs.

server/services/memory/incident_index.py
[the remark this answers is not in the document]
OlivierTrudeau: Fair, the append path still had it. append_to_memory now takes an optional dedup marker and rechecks for the incident id after it grabs the FOR UPDATE lock, so a second task that raced the pre-check just skips instead of duplicating the line. append_incident_line passes INC <id> as the marker.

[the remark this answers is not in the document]
OlivierTrudeau: Kept it opt in rather than always skipping because append_to_memory is a general tool. Duplicate content is often legit there (repeated findings, notes, headers), so an always on skip would silently drop valid appends. It also can't guess the right dedup key. The index wants to match on the incident id, not the whole line, so the caller passes that marker explicitly. Fix is committed.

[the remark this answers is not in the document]
OlivierTrudeau: I had forgotten to push. Check again

isiddharthsingh (approved):

PR #630, opened 2026-09-11
feat(memory): let users edit memory entries

OlivierTrudeau:
## Description

Builds on top of [#556](https://github.com/Arvo-AI/aurora/pull/556) (the unified Memory system). Previously users could only **view** the list of memory entries and delete them. This adds a **pen (edit) action** next to the trash icon so a user can open an entry, review its full content, and modify it.

Targets the `feat/memory-system` branch (not `main`) so it stacks on #556.

### Frontend (`client/src/components/MemorySettings.tsx`)
- Added a **Pencil** edit button next to the existing delete button in each memory row (write access only).
- Clicking it opens an **Edit dialog** that fetches the entry's full content via `GET /api/memory/entries/<id>` (the list endpoint omits content) and lets the user edit **title, category, description, and content**.
- Save button is disabled while loading/saving or when title/content are empty; changes are versioned server-side. Added `title` tooltips to both edit and delete buttons.

### Backend (`server/routes/memory/routes.py`)
- Extended `PUT /entries/<id>` from **category-only** to accept **any subset** of `title`, `category`, `description`, `content`.
  - Content edits are recorded as a new version via `create_version` (`source="manual"`); category/title/description-only edits do **not** create a version.
  - Validates category, non-empty title, non-empty content, and the 500KB content cap.
  - Returns **409** on `(org_id, category, title)` unique-index collisions when renaming/recategorizing.
  - Rejects requests with no updatable fields (400). RLS-safe (`set_rls_context`).
- No proxy change needed — the Next.js memory proxy already forwards `PUT`.

## Testing

- Rebuilt `aurora-server` + `frontend` on this branch; `artifacts` schema (category/description/versioning) confirmed present.
- Ran an in-container test against the real `update_entry` view (RLS + versioning) covering: creating an entry, editing title/category/description/content, verifying persistence via GET, confirming content edits bump version to 2, category-only edits create no version, empty-content and no-field requests are rejected (400), version count in DB, and cleanup. **All assertions passed.**
- `cd client && npm run lint` passes.

### DIFF
--- a/client/src/components/MemorySettings.tsx
+++ b/client/src/components/MemorySettings.tsx
@@ -19,6 +19,14 @@ import {
   SelectTrigger,
   SelectValue,
 } from "@/components/ui/select";
+import {
+  Dialog,
+  DialogContent,
+  DialogDescription,
+  DialogFooter,
+  DialogHeader,
+  DialogTitle,
+} from "@/components/ui/dialog";
 import { useToast } from "@/hooks/use-toast";
 import { useUserId } from "@/hooks/use-user-id";
 import {
@@ -33,6 +41,7 @@ import {
   Lightbulb,
   ScrollText,
   Package,
+  Pencil,
 } from "lucide-react";
 import { useUser } from "@/hooks/useAuthHooks";
 import { DiscoverySettings } from "@/components/DiscoverySettings";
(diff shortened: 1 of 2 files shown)


Noahcasarotto (commented):
Collision on create has no warning and just overwrites.

Creating a memory entry whose category + title already exist doesn't warn, it silently replaces the existing entry (create_entry's ON CONFLICT DO UPDATE) and the UI toasts "Memory entry created". I hit this twice testing today: a 12 KB agent-written entry got replaced by an 18-character one with no warning. Same thing via the API: two POSTs to /api/memory/entries with the same category + title both return 201 with the same id, one row remains, content replaced, version bumped.

This PR makes rename-onto-duplicate return 409, so create should refuse the same way, or the UI should show a warning and confirm before overwriting. The old content is still in artifact_versions but there's no restore in Memory settings.

PR #635, opened 2026-09-15
fix(helm): default image tag to chart appVersion instead of "latest"

OlivierTrudeau:
## Problem

Client deployments pin `image.tag: "latest"` in `values.generated.yaml`. Running `helm upgrade --version X.Y.Z` bumps the **chart** but never rolls the **pods**:

- The rendered image string stays `.../aurora-server:latest` — an unchanged pod template, so Kubernetes sees no diff and triggers **no rollout**.
- Every deployment hardcodes `imagePullPolicy: IfNotPresent`, so even a manual pod restart reuses the node's **stale cached** `latest` layer instead of re-pulling.

Net effect: after an "upgrade", clients keep running the old images. `latest` is mutable and non-deterministic — the wrong thing to pin for a shipped product.

## Fix (Level 1: tie image tag to the chart)

Make the deployed image version follow the chart version automatically, so the client's upgrade is a single deterministic command with no `--set image.tag` juggling.

- **All image templates** now resolve the tag as `{{ .Values.image.tag | default .Chart.AppVersion }}`, matching the pattern `mcp-deployment.yaml` already used. Covers `server`, `chatbot`, `celery-worker`, `celery-beat`, and `frontend`.
- **`values.yaml`** default changed from `tag: "latest"` to `tag: ""`, so the `appVersion` fallback actually applies. Pinning an explicit tag (e.g. a `sha-<short>` build) still overrides as before.

With this, the client upgrade path becomes:

```bash
helm repo update
helm upgrade --install aurora aurora/aurora-oss --version 1.3.0 -n aurora
```

The chart is 1.3.0 → images resolve to `:1.3.0` → the image string changes vs. the prior release → pods roll automatically. Because version tags are immutable, `IfNotPresent` remains correct and cache-friendly.

## Verification

`helm lint` passes. `helm template` (with `image.registry=ghcr.io/arvo-ai`):

- **Default (no `image.tag`)** → `ghcr.io/arvo-ai/aurora-server:1.3.0`, `ghcr.io/arvo-ai/aurora-frontend:1.3.0`
- **`--set image.tag=sha-abc1234`** → `...aurora-server:sha-abc1234`, `...aurora-frontend:sha-abc1234` (override still works)

## Client migration note

Existing clients should drop `image.tag: "latest"` from their `values.generated.yaml` (or set it to `""`) so the appVersion fallback takes over. Alternatively they can pin `image.tag: "1.3.0"` explicitly for one release to force the initial rollout off `latest`.

## Follow-ups (not in this PR)

- Level 2: stamp immutable image **digests** (`@sha256:...`) into the chart at release time for fully reproducible, supply-chain-auditable deploys (CI already builds by digest).

## Summary by CodeRabbit

* **Deployment**
  * Helm deployments now use the chart’s application version as the container image tag when no tag is specified.
  * Explicit image tags remain supported for pinning deployments to a chosen build.
  * The default image tag no longer uses the mutable `latest` tag, improving version consistency and enabling matching upgrades to roll out automatically.
* **Documentation**
  * Updated chart guidance to explain automatic version tracking and explicit image pinning.


OlivierTrudeau:
Superseded by #636, which combines this appVersion-tag default with the digest-pinning work (steps 1 + 2) in a single PR.

PR #636, opened 2026-09-15
feat(helm): deterministic image deploys — appVersion tag default + digest pinning

OlivierTrudeau:
Upgrading a client was bumping the chart but leaving the old images running. The cause: they pin `image.tag: "latest"`, so `helm upgrade` re-rendered the same `...:latest` string — no pod change, no rollout — and `IfNotPresent` kept the stale cached image.

This makes deploys deterministic:

- **Tag follows the chart.** All image templates now use `image.tag | default .Chart.AppVersion`, and the `values.yaml` default is `""` instead of `"latest"`. So `helm upgrade --version 1.3.0` just deploys `:1.3.0` and rolls the pods. Explicit tags still override.
- **Optional digest pinning.** Set `image.digests.{server,frontend}` to run an exact binary by `@sha256:...`. Release CI now stamps these into a `values.images-<version>.yaml` attached to each GitHub Release, so clients can pin reproducibly with `-f values.images-1.3.0.yaml`.

Existing clients just need to drop `image.tag: "latest"` from their values (or pin `1.3.0` once) to get off the stale tag.

Verified with `helm lint` + `helm template`: default → `:appVersion`, digest → `@sha256` (incl. mcp/chatbot/celery, which share the server image), digest beats tag, and the `frontendImage` custom override still wins.

## Summary by CodeRabbit

- **New Features**
  - Helm deployments can pin server and frontend images to immutable digests for reproducible releases.
  - Published releases include a ready-to-use values overlay containing exact image digests.
  - Image configuration supports chart-version-based image tracking and optional custom frontend repositories.

- **Documentation**
  - Updated Kubernetes upgrade guidance to use chart versions and optional digest-pinned release overlays.
  - Added rollout verification steps and guidance for avoiding stale deployments caused by mutable image tags.


.github/workflows/publish-images.yml
```diff
@@ -219,10 +219,106 @@ jobs:
...
+          # Image digest pins for Aurora ${VERSION}.
+          # Generated by CI on tag v${VERSION} — do not edit by hand.
+          # Usage: helm upgrade --install aurora aurora/aurora-oss \\
+          #          --version ${VERSION} -f ${OUT}
+          image:
+            registry: ghcr.io/arvo-ai
+            digests:
+              server: "${SERVER_DIGEST}"
+              frontend: "${FRONTEND_DIGEST}"
+          YAML
```
damianloch, line 304: Check validity of this:

The closing delimiter YAML is indented with 10 spaces. In Bash, unless the <<- operator is used (which only works with tabs, not spaces), the closing delimiter of a heredoc must be completely unindented at the very beginning of the line. Because of this, Bash will fail to parse the script, causing the entire tag release workflow to crash.

deploy/helm/aurora/templates/frontend-deployment.yaml
```diff
@@ -37,7 +37,7 @@ spec:
       {{- include "aurora.scheduling" (dict "service" "frontend" "global" $) | nindent 6 }}
       containers:
         - name: aurora-frontend
-          image: "{{ if and .Values.frontendImage .Values.frontendImage.tag }}{{ .Values.frontendImage.registry }}/{{ .Values.frontendImage.repository }}:{{ .Values.frontendImage.tag }}{{ else }}{{ .Values.image.registry }}/aurora-frontend:{{ .Values.image.tag }}{{ end }}"
+          image: "{{ if and .Values.frontendImage .Values.frontendImage.tag }}{{ .Values.frontendImage.registry }}/{{ .Values.frontendImage.repository }}:{{ .Values.frontendImage.tag }}{{ else }}{{ include "aurora.image" (dict "image" "frontend" "global" $) }}{{ end }}"
```
damianloch: If someone configures a custom registry for the frontend image but leaves the tag empty, the helper will silently ignore their custom registry and fall back to the global registry instead.

damianloch (changes requested):

deploy/helm/aurora/templates/frontend-deployment.yaml
OlivierTrudeau: Good catch — fixed in fa2c220d. The override is now keyed off `frontendImage.repository` (the field that actually identifies a custom image) instead of `tag`. `registry` falls back to the global `image.registry` and `tag` to the chart `appVersion` when omitted, so a custom registry is never silently dropped.

Verified with `helm template`:
- registry + repository, no tag → `ghcr.io/myco/aurora-frontend-custom:1.3.0` (custom registry honored ✅)
- repository only → `<global-registry>/aurora-frontend-custom:1.3.0`
- registry + repository + tag → exact override
- no `frontendImage` → default `aurora.image` path (tag/digest) unchanged

.github/workflows/publish-images.yml
OlivierTrudeau: Looked into this closely — it actually parses fine, and I added a clarifying comment in fa2c220d so it's not a trap for the next reader.

The subtlety: this is a YAML `run: |` block scalar. YAML strips the block's common leading indentation (10 spaces here) *before* bash ever sees the script. So although the raw file shows the closing `YAML` indented 10 spaces, bash receives it at column 0, which `<<YAML` matches. The body is likewise de-indented to column 0, so the generated file is correctly unindented.

I verified by parsing the workflow with PyYAML and executing the extracted `run` script with fake digests — exit 0, and the produced `values.images-1.3.0.yaml` is valid, column-0 YAML:

```yaml
image:
  registry: ghcr.io/arvo-ai
  digests:
    server: "sha256:aaa"
    frontend: "sha256:bbb"
```

You're right that it's fragile-looking though — an *extra* indent on just the terminator would break it. The new comment calls that out.

damianloch (dismissed):

damianloch (approved):

PR #637, opened 2026-09-15
Make Azure multi-subscription fan-out efficient

OlivierTrudeau:
## Summary

Fixes the two efficiency issues raised in today's meeting about the new Azure multi-subscription update.

### 1. Inefficient parallel checking of subscriptions (code fix)

**Before:** the fan-out (`_cloud_exec_azure_multi_subscription` in `server/chat/backend/agent/tools/cloud_exec_tool.py`) ran a full `az login --service-principal` **and** allocated a private `AZURE_CONFIG_DIR` **per subscription** — **2N subprocesses and N temp dirs** for N subscriptions.

**Why that was wasteful:** every connected subscription belongs to the *same* service principal, and an SP `az login` is **tenant-scoped** — one login already authenticates every subscription. (Discovery's `azure_asset_discovery` already logs in once and queries many subscriptions.)

**After:** authenticate **once** into a **single shared** `AZURE_CONFIG_DIR`, then fan the per-subscription commands (each pinned with `--subscription`) across it. Cost drops from **2N → N+1 subprocesses** and **N → 1 temp dir**. Safe because login completes up front and the pinned commands never mutate the CLI's active-subscription state.

### 2. Don't check all subscriptions when a specific one was requested (guidance fix)

Rather than adding a name→id resolver in code (overfitting), this updates the **Azure RCA skill** (`server/chat/backend/agent/skills/rca/provider_azure.md`) to:
- warn that omitting `account_id` fans out across **every** subscription and is **expensive**,
- instruct the agent to pass `account_id='SUBSCRIPTION_ID'` from the **first** call whenever the target subscription is already known or named by the user,
- omit `account_id` only when the owning subscription is genuinely unknown.

This keeps the mechanism simple and steers the agent away from the "fan out to all, then find the requested one" pattern.

## Tests

Updated `server/tests/connectors/test_azure_multi_subscription.py` for the single-login model:
- `test_fanout_authenticates_once_and_pins_every_subscription` — exactly **one** `az login` / one shared config dir, every command still pinned to its subscription, secret still reaches `az` byte-for-byte and never via a shell.
- `test_fanout_cleans_up_its_shared_temp_dir`, `test_fanout_fails_closed_when_shared_login_fails`, `test_fanout_isolates_per_subscription_failures` — updated for the single-login model.

All 75 tests in the file pass locally.

## Summary by CodeRabbit

* **Performance**
  * Azure operations spanning multiple subscriptions now authenticate once and reuse that session, reducing execution overhead while preserving subscription-specific targeting.
  * Operations stop safely if shared Azure authentication fails.

* **Bug Fixes**
  * Azure account-level commands no longer receive an incompatible subscription option.

* **Documentation**
  * Azure investigation guidance now recommends specifying a subscription when the target is known.
  * Multi-subscription searches are clearly identified as an expensive option for cases where the target subscription is unknown.


damianloch (dismissed):

server/chat/backend/agent/tools/cloud_exec_tool.py
isiddharthsingh, line 147: This comment is now out of date with the fan-out you changed below. It says a shared dir would race and that the fan-out removes the dir per subscription. After this PR the fan-out calls this once with subscription_id=None, shares that one dir across every worker, and removes it once in its own finally (lines 1499-1502). Could you reword it to match? Something like: "The single-subscription path gets its own dir; the multi-subscription fan-out authenticates once, shares the dir across its workers (commands are pinned with --subscription so nothing mutates the active-subscription state), and removes it in its own finally."

isiddharthsingh, line 2706: Same staleness here: "the fan-out path removes its own per subscription" is no longer true, it removes one shared dir once. Just drop "per subscription".

PR #645, opened 2026-09-18
Reuse one az login per Azure credential set

isiddharthsingh:
Stacked on #637.

## Summary

Every Azure `cloud_exec` command ran a full `az login` before the real command. AWS and GCP pass credentials through the environment and spawn one process per command; `az` has no equivalent, so Azure paid two process starts per command, and the multi-subscription fan-out paid a login per call on top. This keeps one logged-in `AZURE_CONFIG_DIR` per set of credentials and reuses it.

## How it works

- New `utils/cloud/azure_login_cache.py`. The directory name is an HMAC (keyed by `FLASK_SECRET_KEY`) of tenant id, client id and client secret. A caller can only reach a directory by already holding the credentials stored in it, and those are still resolved per request through Vault/RBAC. Org members sharing a connection share one login; different orgs, the read-only vs agent service principal, and a rotated secret each get their own directory.
- `cloud_exec` (single path) and `_cloud_exec_azure_multi_subscription` call `attach`/`ensure`. A warm fan-out spawns no login at all. The login runs under a file lock, so concurrent cold requests log in once.
- If a command fails because the login is gone or predates a newly connected subscription, it logs in again and retries once. The fan-out does this once per batch and re-runs only the failed subscriptions.
- Commands that write CLI-local state (`az login`, `logout`, `account set|clear`, `config`, `configure`, `extension`, `cloud`, ...) keep a private throwaway directory.
- Directories are 0700, expire after `AZURE_LOGIN_CACHE_IDLE_SECONDS` idle (default 1800; `0` disables and restores the previous behaviour), and are refreshed after 8 hours. The cache is off under pod isolation, where `az` runs in a pod the server filesystem cannot see.
- `setup_azure_environment_isolated` is unchanged; discovery deletes the directory it returns.
- Two guardrail signatures block commands that name the cache directory or the `service_principal_entries` / `msal_token_cache` files, including through globs.

## Credential isolation

`az` writes the service principal secret in plaintext inside the config dir and refuses to run without it (`load_entry` in `azure/cli/core/auth/identity.py`), so the secret now rests on disk for the idle window instead of for a single command. The directory is keyed by the credentials, never by user, and nothing here decides who may use credentials.

## Measured (local, 2 subscriptions)

- `az login` cost 2.3 s per command before. Warm commands after this change log `Azure login reused from cache` and run zero logins.
- The remaining time per command is the safety judge (2 to 6 s, on the base command and again on each subscription copy) plus about 1 s of actual `az` work. Not changed here.

## Tests

- `tests/utils/test_azure_login_cache.py` (new): warm path, concurrent cold requests, isolation per credential set, root and symlink checks, local-state commands, idle and max-age expiry, sweep, relogin dedupe, pod-isolation off.
- `tests/connectors/test_azure_multi_subscription.py`: warm fan-out spawns no login and keeps the shared directory; a dead cached login heals with one relogin; `cloud_exec` never deletes a cached directory.
- `tests/security/test_credential_creation.py`: the new signatures.
- End to end in the chatbot container with a fake `az`: cold 2 processes, warm 1, `account set` in a private directory, login-lost self-heal, kill switch.

## Summary by CodeRabbit

* **New Features**
  * Azure CLI logins are now cached and reused across commands for the same credentials, reducing repeated authentication.
  * Stale or failed cached sessions automatically trigger a fresh login.
  * Added configurable cache expiration via `AZURE_LOGIN_CACHE_IDLE_SECONDS`, defaulting to 1,800 seconds; set to `0` to disable reuse.

* **Documentation**
  * Added configuration guidance for Azure CLI login reuse and its limitations.


isiddharthsingh:
@coderabbitai review

server/utils/cloud/azure_login_cache.py
```diff
@@ -0,0 +1,308 @@
...
+    words = [t.lower() for t in tokens if not t.startswith("-")]
+    # Deliberately loose: a global flag's value can sit before the group
+    # (`az --output json login`). A false positive only costs one private login;
+    # a false negative would let a command replace the identity in a shared directory.
+    if any(w in _LOCAL_STATE_GROUPS for w in words[:3]):
+        return True
+    return any(
+        a == "account" and b in _LOCAL_STATE_ACCOUNT_SUBCOMMANDS
+        for a, b in zip(words, words[1:])
+    )
```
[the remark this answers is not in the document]
isiddharthsingh: Fixed in 6208ea12. `uses_local_cli_state` now checks every non-flag word, not the first three, so any number of global flag values ahead of `login` or `account set` is caught. Added the four-flag case and the `account set` variant to the tests.

server/utils/cloud/azure_login_cache.py
```diff
@@ -0,0 +1,308 @@
...
+        started = time.perf_counter()
+        result = run_login()
+        logger.info("TIME: az login took %.2fs", time.perf_counter() - started)
+        if result.returncode != 0:
+            shutil.rmtree(self.config_dir, ignore_errors=True)
+            return False, (result.stderr or "")
+        fd = os.open(self._marker, os.O_CREAT | os.O_WRONLY | os.O_TRUNC, 0o600)
+        with os.fdopen(fd, "w", encoding="utf-8") as fh:
+            json.dump({"created": time.time()}, fh)
+        return True, ""
```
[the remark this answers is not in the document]
isiddharthsingh: Intentional, and described in the PR body. `az` cannot run without the service principal entry in its config dir (`load_entry` in `azure/cli/core/auth/identity.py` raises without it), so a private per-invocation dir is exactly the login-per-command cost this PR removes. What bounds the exposure: the dir is keyed by the credentials themselves (HMAC of tenant, client id and secret, so only a caller already holding them can land in it), it is 0700 under a root that is checked for ownership and mode, it expires after `AZURE_LOGIN_CACHE_IDLE_SECONDS` idle with an 8h hard cap, commands that write CLI state never use it, the guardrail now blocks commands naming the dir or the credential files, and the whole thing is off under pod isolation or with the idle window set to 0. Same-UID processes in the container could already read the per-command dir and the secret in the process environment during the seconds a command ran; this widens that window to the idle period, nothing more. If we want this recorded as an exemption to the credential-isolation invariant in CLAUDE.md, happy to add it in a follow-up.

docker-compose.prod-local.yml
```diff
@@ -153,6 +153,9 @@ x-common-env: &common-env
   # Pod Isolation (disabled for local)
   ENABLE_POD_ISOLATION: ${ENABLE_POD_ISOLATION}
 
+  # Azure CLI login reuse: idle seconds before a cached `az login` is dropped (0 disables)
+  AZURE_LOGIN_CACHE_IDLE_SECONDS: ${AZURE_LOGIN_CACHE_IDLE_SECONDS:-1800}
```
OlivierTrudeau, line 157: this is polluting the env in my opinion. It should be set in a file or something rather than at the env level

server/chat/backend/agent/tools/cloud_exec_tool.py
```diff
@@ -1383,6 +1385,64 @@ def _apply_azure_subscription(command: str, subscription_id: str) -> str:
...
+def _run_azure_on_subscription(conn: dict, command: str, isolated_env: dict, timeout: Optional[int]) -> dict:
+    """Run one subscription-pinned command against the shared, already-authenticated config dir.
+
+    The command is pinned to the subscription, so concurrent workers never contend
+    on the CLI's active-subscription state.
+    """
+    sub_id = conn.get("account_id", "unknown")
+    try:
+        cmd = _apply_azure_subscription(command.strip(), sub_id)
+        try:
```
OlivierTrudeau, line 1397: NOt sure if worth modifying, but we could technically get rid of the list of commands that don't support --subscription and instead, catch the error message from azure and if we get "Unrecognized argument --subscription", we try running the command without subscription instead. Food for thought decide whether or not you want to implement it

server/utils/cloud/azure_login_cache.py
```diff
@@ -0,0 +1,309 @@
...
+  * Commands that mutate the CLI's own state (`az login`, `az account set`, ...)
+    never run against a cached directory; the caller keeps its private one.
+  * Directories are 0700, expire after an idle window, and are force-refreshed
+    after a maximum age. `az` stores the client secret in plaintext inside the
+    directory and refuses to run without it, so the idle window bounds how long
+    the secret rests on disk.
+
+Each container has its own cache; nothing is shared across containers. With pod
+isolation on, `az` runs in a per-session terminal pod whose filesystem this
+module cannot see, so the cache is disabled there and every command logs in.
```
OlivierTrudeau, line 25: This bloats the description for not much added value

server/utils/cloud/azure_login_cache.py
```diff
@@ -0,0 +1,309 @@
...
+
+
+def _pod_isolation_enabled() -> bool:
+    """True when terminal_run executes commands in a per-session terminal pod.
+
+    Same default as terminal_run: unset means on. The directory `az` logs into
+    then lives in the pod, not on this filesystem, so a marker here would vouch
+    for a login the pod never performed.
+    """
+    return os.getenv("ENABLE_POD_ISOLATION", "true") == "true"
```
OlivierTrudeau, line 139: Kind of a useless function

OlivierTrudeau (changes requested):

docker-compose.prod-local.yml
isiddharthsingh: Agreed. Made it a module constant next to the other timing knobs (max age, sweep interval, orphan grace) and dropped the variable from the compose files, .env.example, Helm values and the env docs. Trade-off: no runtime switch anymore, disabling means editing the constant.

server/chat/backend/agent/tools/cloud_exec_tool.py
isiddharthsingh: Leaving the list for now. In the fan-out the retry would cost one failed az run per subscription for exactly the tenant-wide commands: with 47 subscriptions `az account list` becomes 47 failures plus 47 reruns instead of 47 runs. The follow-up to run tenant-wide commands once before fanning out also needs to know those commands up front, which an error-driven retry only learns mid fan-out. The over-broad `az account` prefix gets narrowed in that same follow-up.

server/utils/cloud/azure_login_cache.py
isiddharthsingh: Removed the paragraph.

server/utils/cloud/azure_login_cache.py
isiddharthsingh: Removed the wrapper and inlined the check in `attach`. The check itself stays: with pod isolation on the login happens inside the pod's filesystem while the marker would be written here, so the next command would find an empty directory and pay a failed run plus a relogin every time.

